PR #28, opened 2015-02-26
cli program is ready

areeda:
Hi Duncan,
Testing is going well, I expect anything we find to be minor.

Joe


duncanmmacleod:
The `importlib` module is only available in python 2.7 and above, meaning the gwpy-ldvw.py executable doesn't work out-of-the-box on py26. Can you add an optional dependency to the `setup.py` script to require `importlib` if on py26? Just copy the system I used for the `ordereddict` package.

duncanmmacleod:
Can you also rename the `gwpy-ldvw.py` executable to `gwpy-plot` or something like that. The interface is technically independent of the LDVW service, and python executables typically don't carry a file extension.

duncanmmacleod:
Another minor thing. Your codes are not fully PEP-8 compliant. I'm not going to ask you to change it just for this PR, but will ask that you run the `pep8` command line tool over the new codes at some point, and push a commit with some whitespace changes.

areeda:
The importlib module has been backported to 2.6, I'm using on SL6 with that version of python.  It is needed by the program so we wouldn't be able to run it on clusters without it.  It is available via pip:
joe@george:~/ligo/ldvwTest$ pip search importlib
pyitect                   - A Simple system for structuring a modeler project architecture via plugin like modules, uses the new importlib abilities first avalible in
                            python 3.4, includes an exec load mode for support of pyhton 3.0+
importlib                 - Backport of importlib.import_module() from Python 2.7
  INSTALLED: 1.0.3 (latest)
importlib2                - A backport of the Python 3 importlib package.
backport_importlib        - A script to backport the full importlib to 2.x
importlib-no-failure      - Backport of importlib.import_module() from Python 2.7
importlib_full            - Backport of Brett Cannon's full importlib module.

If there is a 2.6 alternative, and you don't want to make it a depencency, we can change it.

areeda:
If you insist we can remove ldvw from the program name (no problem removing the .py) but the reason it is there is more than self promotion (there is some of that).  I think a common use case will be people using ldvw to make a plot or two and then copy paste the gwpy-ldvw (or gwpy-plot) command line and scripting it for use locally.  So I was thinking the ldvw tag makes it clear it's the same one used on the web.

We can change the name if you think it's confusing.

areeda:
I've installed pep8 and see the huge list of picky shit.

I think they are all easy to clean up.  I'll start on them and see if I can get them done today or tomorrow.

duncanmmacleod:
re: importlib:

It is backported, so you can just add a version check in `setup.py` to add `importlib` as an installation dependency similar to what is done for ordereddict [here](https://github.com/gwpy/gwpy/blob/master/setup.py#L50). That way it will get installed by default for users on py26

duncanmmacleod:
I've just noticed that you committed the `docs/lalsuite_dox.tag` file, we're going to need to remove that file before we continue. It's big, and ugly. I'll have a think

areeda:
I though that lalsuite_dox.tag came when I merged upstream.  I'll remove it from my git repository (well from further commits).

duncanmmacleod:
The file does seem to have come in from a merge operation, but I have never committed that file to the master repository. Strange...

areeda:
Do you want to put it into .gitignore and remove it.  I think it will then go away when I merge upstream.

duncanmmacleod:
It's already in the upstream [`.gitignore`](https://github.com/gwpy/gwpy/commit/aee15edbf2fad74751b843604e644807948e72f4). I want to get this removed permanently from the history, so not sure what you mean when you say it'll 'go away' when you merge.

areeda:
I didn't think there was a way to remove stuff from the history.  rm stops it from going forward.

If you find a way let me know.

duncanmmacleod:
The help page here

https://help.github.com/articles/remove-sensitive-data/

describes how to permanently remove files from the history, basically forcefully deleting the commits that introduced it.

areeda:
Should I do that from my fork too?

If so please send me the command you use.

duncanmmacleod:
I think the following command will do what we want:

```bash
git filter-branch --force --index-filter 'git rm --cached --ignore-unmatch docs/lalsuite_dox.tag' --prune-empty --tag-name-filter cat -- --all
```
After that you can force push the changes to your remote:

```bash
git push origin --force --all
```

I'm not sure what will happen to this PR, we might have to open a new one to get the right commits recognised.

duncanmmacleod:
I have successfully cloned, built, tested, and generated the sphinx docs for this new development.

The sphinx documentation is full and very readable, and the `gwpy-plot` command-line interface is easy to use. This is an excellent new feature set, thanks @areeda, a few beers coming your way.

I am ready to merge this upstream if you are.

duncanmmacleod:
Caveat, I see the image linking problem, we should probably fix that before merging.

areeda:
For the record here:

* The fixes you put in for the IFO solved my problems.
* The patch you sent for the image linking problem worked.
* As far as I can tell we are ready for release.

I will start the list of enhancements to the command line tools for the next release soon.

PR #33, opened 2015-02-27
Allow custom IFO prefix for Channels

### LINKED ISSUE #31: Time series fetch seems to have trouble with with odd prefixes
The following code snippet fails with the latest pull from dev, it did work in previous versions:
```python
chan = 'LVE-EX:X3_810BTORR.mean,m-trend'
dur = 36000
start = 1072539420
data = TimeSeries.fetch(chan, start, start+dur, verbose=True)

produces the error:
```AttributeError: 'Channel' object has no attribute '_ifo' ```

This is one of the examples in gwpy-plot documentation.  The plot from a previous version looks like: 

https://ligodv.areeda.com/gwpy/cli_examples/cli-ts-02.png

### PR DESCRIPTION
**Pull request by duncanmmacleod**
Old LIGO FMCS channels have odd prefixes, but are still valid channels, e.g.

```
'LVE-EX:X3_810BTORR.mean,m-trend'
```

This PR introduces a unittest to make sure these channels can be parsed, and amends the `Channel.MATCH` regular expression to allow custom prefixes.

This should fix #31.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ before_install:
   - sudo apt-get --assume-yes --allow-unauthenticated install lscsoft-archive-keyring
   - sudo apt-get update -qq
   - sudo apt-get --assume-yes install libhdf5-dev lalframe-python ldas-tools-framecpp-python
+  - sudo apt-get --assume-yes install python-nds2-client
   # install build dependencies
   - travis_retry pip install -q tornado jinja2 GitPython
   # install cython
--- a/gwpy/detector/channel.py
+++ b/gwpy/detector/channel.py
@@ -108,13 +108,13 @@ class Channel(object):
             name = str(name)
         # make a new channel
         # strip off NDS stuff for 'name'
-        self.name = str(name).split(',')[0]
         # parse name into component parts
         try:
             parts = self.parse_channel_name(name)
         except (TypeError, ValueError):
-            pass
+            self.name = str(name)
         else:
+            self.name = str(name).split(',')[0]
             for key, val in parts.iteritems():
                 try:
                     setattr(self, key, val)
--- a/gwpy/detector/channel.py
+++ b/gwpy/detector/channel.py
@@ -286,7 +286,10 @@ class Channel(object):
 
         :type: `str`
         """
-        return self._ifo
+        try:
+            return self._ifo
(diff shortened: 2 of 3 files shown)


duncanmmacleod:
@areeda: can you comment on whether we need to support essentially arbitrary IFO prefixes (as implemented here), or just make an exception for the `LVE-EX`-style channels?

areeda:
Well, I can certainly comment but not authoritatively.

I think that if we plan to support any NDS 1 & 2 servers it will be hard to predict what the channel prefix (AKA ifo) could be.  For example the 40m at CIT uses C1 as their prefix.  The NDS1 test stand at LLO uses X1 as theirs.  I'm not sure about LOSC but I think they have their own frame types and perhaps channel names.

I don't think an unrecognized prefix should be an error, although I don't see a good way to predict what servers to try if we don't recognize the prefix.

duncanmmacleod:
Ok, so I've reverted my original change to the channel name parsing.

Now this PR just fixes the `AttributeError` you get if the IFO is not present. Instead the IFO defaults to `None` which means that the CIT NDS2 server will get selected by default.

@areeda, is that good enough?

areeda:
I think the the HVE and LVE are curently active channels that are interesting the vacuum people.
By my count for LVE 2,464 out of 2,487 are being acquired now. HVE: seems to have all 3080 active.

If this reverts to the the way it used to work, I think it's good enough.  The only problem I see is when people request current data,  they will have to wait until it gets to CIT. (using the command line tools, from Python they can specify the server, I believe).

duncanmmacleod:
Ok, I am beginning to think we need an interferometer-independent `site` property of the `Channel` object, which can be parsed for the `HVE` or `LVE` channels. 

Then the new `site` property can be used to determine the NDS2 server to use, rather than `ifo`. @areeda, if you agree, I would merge this pull request, and I can work on this in a new issue.

areeda:
I am about to test the PR now.
I've been struggling with how this can work best.
It seems as though we need a per channel way to determine which server list to use.  It's hard without a database but I think your heuristic with requested time will work well.

An option might be a RESTful call to access ldvw's database.  I'll look into that from my end and we can maybe add it later as something to try first but not depend on.

Optimizing multiple NDS requests should give us a considerable performance increase.

PR #36, opened 2015-03-03
GPS epoch-based server selection

### LINKED ISSUE #27: Select server for nds transfer based on time of the data
Data at the sites LLO/LHO are archived to tape after about a month, perhaps 3 weeks.  Data can take hours to retrieve from tape if multiple people are using it.

LDR sometimes gets behind so while data usually arrives at CIT in 10 minutes, we seen it take a week when something nasty happens.

A solution, or at least improvement would be to have the choice of server to try first depend on the age of the data being requested.  A good cut-off value seems to be 2 weeks.  If the data is older than that NDS2 server at CIT should be preferred over the server at the appropriate observatory.

duncanmmacleod:
This PR introduces the `epoch` keyword argument to the NDS2 server selection algorithm (`gwpy.io.nds.host_resolution_order`). Additionally, this introduces a series of related unittests.

This fixes #27.

### DIFF
--- a/gwpy/io/nds.py
+++ b/gwpy/io/nds.py
@@ -29,6 +29,7 @@ import warnings
 import nds2
 
 from .. import version
+from ..time import to_gps
 from .kerberos import *
 
 __author__ = "Duncan Macleod <duncan.macleod@ligo.org>"
@@ -89,7 +90,8 @@ class NDSWarning(UserWarning):
 warnings.simplefilter('always', NDSWarning)
 
 
-def host_resolution_order(ifo, env='NDSSERVER'):
+def host_resolution_order(ifo, env='NDSSERVER', epoch='now',
+                          lookback=14*86400):
     """Generate a logical ordering of NDS (host, port) tuples for this IFO
 
     Parameters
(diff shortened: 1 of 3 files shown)


duncanmmacleod:
@areeda, can you take a quick look through the code, and see if it makes sense? From a LDVW point-of-view, this should be a transparent improvement on the NDS2 server selection which defaults to selection the CIT server if the start time of the fetch request is more than two weeks ago.

areeda:
I can't seem to find the code you're referring to.  I merged with upstream/master and ran the gwpy-plot tests (images for documentation) and am now running the automated ldvw tests.

It looks to me like I should be looking in nds.py: host_resolution_order() but I don't see the tests for age of the request.  Nor do I see anything in timeseries.py:fetch

@duncanmmacleod  Where should I be looking?

duncanmmacleod:
@areeda, you will need to clone this branch as we did before:

```bash
cd /path/to/gwpy/repo
git checkout -b test_nds2_server_selection
git pull https://github.com/duncanmmacleod/gwpy.git nds2_server_selection
```

Once you are done testing, you can then delete the test branch:

```bash
git checkout master
git branch -D test_nds2_server_selection
```

areeda:
D'Oh

PR #45, opened 2015-03-09
cli: fixes for matplotlib units error on py26

duncanmmacleod:
This PR fixes a python2.6/matplotlib bug where astropy Quantities are pushed through the matplotlib units interface, resulting in a `SystemError`.

This single commit just makes sure a couple of `set_ylim` calls receive float-like objects, and not Quantities.


duncanmmacleod:
@areeda, can glance at this and give me an ok to merge?

PR #47, opened 2015-03-09
New over-dense spectrogram method (TimeSeries.spectrogram2)

### LINKED ISSUE #25: Spectrogram calculation is sub-optimal
The current `Spectrogram` calculation (from a `TimeSeries`) uses non-overlapping chunks, with each one being used to calculate a PSD. This means the edges of each chunk are not represented properly in the overall result, and that the resolution of the resulting `Spectrogram` is low.

We need to rethink the way the `Spectrogram` is calculated, mainly for plotting purposes.

duncanmmacleod:
This PR introduces a new over-dense spectrogram method for calculating and displaying short-duration spectrograms. Also included is a new example, with the following code:

```python
gwdata = TimeSeries.fetch('L1:OAF-CAL_DARM_DQ', 'Feb 28 2015 06:02:05', 'Feb 28 2015 06:02:10')
specgram = gwdata.spectrogram2(fftlength=0.15, overlap=0.14) ** (1/2.)
medratio = specgram.ratio('median')
plot = medratio.plot(norm='log', vmin=0.5, vmax=10)
plot.set_yscale('log')
plot.set_ylim(40, 8192)
plot.add_colorbar(label='Amplitude relative to median')
plot.set_title('L1 $h(t)$ with noise interference')
plot.show()
```

which produces the following output:

spectrogram2

There are minor problems with precision errors making their way up to determine the length of the spectrogram, but since this is mainly for visualisation purposes, I'm not too fussed.

This should fix #25.

### DIFF
--- a/examples/spectrogram/spectrogram2.py
+++ b/examples/spectrogram/spectrogram2.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+
+# Copyright (C) Duncan Macleod (2013)
+#
+# This file is part of GWpy.
+#
+# GWpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# GWpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


duncanmmacleod:
@areeda, can you try this out on LDVW and let me know if it meets your needs.

You will likely need to hand-extend the fetch() request to over-fetch data to cover the required span of the plot, rather than shortening the plot (which I think is what happens now).

PR #62, opened 2015-03-23
Long dur specgram

### LINKED ISSUE #61: Long duration spectrograms are setting stride incorrectly in gwpy-plot
I've been making 12hr spectrograms and they look blocky. I believe the problem is with the stride calculation.

areeda:
Fixed #61 Long durations spectrograms were setting stride wrong in the cli program


areeda:
Fixed merge conflicts

PR #79, opened 2015-04-16
Two new examples: coherence and coherence_spectrogram

duncanmmacleod:
This PR introduces two new examples, one for `TimeSeries.coherence` and the other for `TimeSeries.coherence_spectrogram`.

All choices for data and times were ruthlessly copied from Joe's CLI examples.


maxisi:
@duncanmmacleod are these coherence plots exposed to gwsumm?

PR #83, opened 2015-04-22
Multiple optimisations for TimeSeries methods

duncanmmacleod:
This PR introduces a number of critical optimisations for routines in GWpy. The summary is as follows:

- `gwpy.segments` is now only imported at the function level for `Channel` objects and all `Array` sub-classes (include, e.g, `TimeSeries`) - this import is slow because it relies on `gwpy.table.ligolw` imports
- `TimeSeries.read(format='framecpp') now intelligently handles accessing the TOC for a frame, and does it as seldom as possible
- `TimeSeries.spectrogram` now calculates a single window function up-front, rather than passing the default argument and having `scipy` recalculate the window function every time `welch` is called
- `TimeSeries.read(cache)` will now try to use `lalframe`-based serial access for up to 4 channels since `lalframe` is much faster than `frameCPP.py` - this is set in `gwpy.timeseries.io.cache.MAX_LALFRAME_CHANNELS`

Additionally, a number of bugs have been squashed as they were discovered.

### DIFF
--- a/examples/timeseries/statevector.py
+++ b/examples/timeseries/statevector.py
@@ -44,6 +44,7 @@ bits = [
 ]
 
 data = StateVector.fetch('L1:ISI-ETMX_ODC_CHANNEL_OUT_DQ', 'May 22 2014 14:00', 'May 22 2014 15:00', bits=bits)
+data = data.astype('uint32')  # hide
 
 # For this example, we wish to :meth:`~StateVector.resample` the data to a
 # much lower rate, to make visualising the state much easier:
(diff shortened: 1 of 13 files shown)


duncanmmacleod:
@areeda: can you take a quick look through this and sanity check what I've done here. This should improve performance for `gwpy-plot` as well as anything else, so once it's merged we should get it onto `ldvw` fairly quickly.

duncanmmacleod:
It's easier to merge now and sort the bugs out later once their upstream

PR #101, opened 2015-06-23
table.io.trigfind: raise exception if no channel directory

### LINKED ISSUE #72: SnglBurstTable.fetch does not return error messages
Trying to get Omicron triggers using the fetch method of SnglBurstTable, I found that no error message is returned if the channel is not analyzed by Omicron (or if the channel name is wrong). Also no error message is given if the name of the ETG is wrong, and the help message doesn't list the possible ETGs, or specify formatting (e.g. omicron vs Omicron).

Examples:

SnglBurstTable.fetch('H1:NOT-A_CHANNEL','omicron',1111230000,1111240000)

SnglBurstTable.fetch('H1:LSC-DARM_IN1_DQ','iota',1111230000,1111240000)

Both return an empty list.

duncanmmacleod:
This PR adds an error condition to `table.io.trigfind.find_trigger_urls` (used by `SnglBurstTable.fetch`) when a channel-level directory is not found, meaning this channel has never been processed by the given ETG.

This fixes #72.

### DIFF
--- a/gwpy/table/io/trigfind.py
+++ b/gwpy/table/io/trigfind.py
@@ -59,6 +59,13 @@ def find_trigger_urls(channel, etg, gpsstart, gpsend, verbose=False):
     trigform = ('%s-%s_%s-%s-*.xml*'
                 % (ifo, re_dash.sub('_', channel), etg.lower(), '[0-9]'*10))
 
+    # test for channel-level directory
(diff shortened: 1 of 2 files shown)


duncanmmacleod:
@andrew-lundgren: can you confirm this should resolve your issue #72?

PR #128, opened 2015-08-07
Bump SciPy requirement to 0.16.0

duncanmmacleod:
This PR bumps the minimum required scipy version to 0.16, and astropy to 1.0.

[Scipy 0.16](https://pypi.python.org/pypi/scipy/) is the first version to support second-order-section filter decomposition, so is critical for proper time-domain filtering.

@areeda: can you please review the implications for LDVW and let me know if this change is ok?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_install:
   # install numpy
   - travis_retry pip install -q numpy==1.9.1
   # install scipy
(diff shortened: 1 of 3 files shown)


areeda:
Tested on https://ldvw-dev.ligo.caltech.edu and it works well.

PR #129, opened 2015-08-10
New DataQualityFlag.pad method

duncanmmacleod:
This PR introduces the `DataQualityFlag.pad` method, and an associated test case.

This method applies either a `(start, end)` tuple, or reads such a tuple from the `DataQualityFlag.padding` property.


duncanmmacleod:
The padding method should accept both `start` and `end` padding as forward-in-time, meaning negative numbers should be given for `start` if meant to pad out (back in time).

PR #130, opened 2015-08-12
TimeSeries.whiten: new method

duncanmmacleod:
This PR introduces two new methods for the `TimeSeries`:

- `TimeSeries.detrend`: a simple wrapper around `scipy.signal.detrend` to return a `TimeSeries`
- `TimeSeries.whiten`: a function to whiten a `TimeSeries` against its own ASD

The new methods are accompanied by unit tests and a new example for whitening a `TimeSeries` to see glitches at higher frequency.

The example reads in some data from NDS, and whitens it to show two glitches at high frequency, the resulting image is:

whiten

### DIFF
--- a/examples/timeseries/whiten.py
+++ b/examples/timeseries/whiten.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python
+
+# Copyright (C) Duncan Macleod (2013)
+#
+# This file is part of GWpy.
+#
+# GWpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


duncanmmacleod:
@andrew-lundgren: can you review the method to make sure it makes sense, and give me the go ahead to push this upstream?

duncanmmacleod:
[Here](https://ldas-jobs.ligo-la.caltech.edu/~duncan.macleod/gwpy/pulls/130/examples/timeseries/whiten.html) is a preview of the online version of new example.

PR #141, opened 2015-09-16
TimeSeries.read: fix bug in reading with default GWF format lalframe

duncanmmacleod:
This PR improves the unit tests for `TimeSeries.read()` to check that each GWF format works with and without the `format` keyword argument being given.

Given that this test currently fails for `format='lalframe'`, subsequent commits will fix the problems with that format.

Thanks to @mattpitkin for his patience in reporting and helping me diagnose the problem.

### DIFF
--- a/gwpy/tests/test_timeseries.py
+++ b/gwpy/tests/test_timeseries.py
@@ -31,6 +31,7 @@ from numpy import testing as nptest
 from scipy import signal
 
 from astropy import units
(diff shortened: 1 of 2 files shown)


mattpitkin:
Well done figuring this one out. I'll pull the changes and test them myself in the morning.

mattpitkin:
I've tested this too and it works for me. Thanks.

PR #142, opened 2015-09-16
Update travis-ci integration to new infrastructure

duncanmmacleod:
This PR attempts to update the CI configuration to the new travis-ci.org infrastructure.

The new `build_lal.sh` script build's the LSCSoft dependencies from source, instead of using custom apt installs.

This may be deprecated if the lscsoft deb source gets whitelisted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,39 +2,30 @@ language: python
 python:
   - "2.6"
   - "2.7"
+addons:
+  apt:
+    packages:
+      - libc6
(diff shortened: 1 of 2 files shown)


duncanmmacleod:
Ok, it seems that the necessary dependencies of LAL are not provided by default, so we need the lscsoft debian repository before proceeding. I have opened travis-ci/apt-source-whitelist#148 to get this approved.

PR #156, opened 2015-10-23
Fixed bug in Series.append treatment of xindex

### LINKED ISSUE #148: UnitConversionError during TimeSeries.append() usage
When using spectrogram.append() or timeseries.append() i often get an exception saying Valueerror: TimeSeries units do not match: None vs undef " or "Timeseries units do not match: None vs undef". 
The issue seems to be caused by line [441 of gwpy/series.py](https://github.com/gwpy/gwpy/blob/master/gwpy/data/series.py#L441) where a value without unit is copied in a valued value. The issue seems to be easily solved by changing 
``` python
            try:
                self.xindex[-other.shape[0]:] = other.xindex.value
``` 
with 
``` python
            try:
                self.xindex[-other.shape[0]:] = other.xindex
```

duncanmmacleod:
This PR fixes a bug in the treatment of `Series.xindex` when calling `Series.append`. Fixes were made to the method, and the test suite.

This fixes #148.

### DIFF
--- a/gwpy/data/series.py
+++ b/gwpy/data/series.py
@@ -438,7 +438,7 @@ class Series(Array):
             else:
                 self.xindex[:-other.shape[0]] = self.xindex[other.shape[0]:]
             try:
-                self.xindex[-other.shape[0]:] = other.xindex.value
+                self.xindex[-other.shape[0]:] = other._xindex
             except AttributeError:
                 del self.xindex
             try:
--- a/gwpy/tests/test_array.py
+++ b/gwpy/tests/test_array.py
@@ -245,8 +245,10 @@ class SeriesTestCase(CommonTests, unittest.TestCase):
     def test_append(self):
         """Test the `Series.append` method
         """
+        # create arrays
         ts1 = self.create()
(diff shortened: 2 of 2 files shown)


duncanmmacleod:
@mikelovskij, can you check out this PR branch and verify that it fixes the bug you reported?

mikelovskij:
After some troubles (caused by me trying to install the version of gwpy in the master branch instead of the right branch), I can confirm that this fix solves the issue for bot the spectrogram and timeseires monitors (other monitors didn't have this issue). 
Now the question is: is there any way for me to try it in the control room machines while using your virtual environment (I had troubles in installing some of the gwpy requirements in my user folder on those machines). 
Thanks.

duncanmmacleod:
The only real way to confirm is to install your own virtualenv in your home directory, as I have done. You should be able to just call

```bash
bash ~duncan.macleod/git/gwpysoft/gwpysoft-init ~/opt/gwpysoft ~duncan.macleod/git/gwpysoft/packages.txt
```

to install the virtual environment all in one go. Should work, but no guarantees.

I'll consider your note as confirmation that this can be merged.

PR #171, opened 2015-11-05
Return more default hosts

### LINKED ISSUE #170: Backup hosts don't get used if $NDSSERVER doesn't have the data
Installing gwpy on a 40m machine here. Our `$NDSSERVER` environment variable is set to point at our local frame builder, which doesn't mirror any aLIGO data. 

`gwpy.io.nds.host_resolution_order` only tries to use the server in `$NDSSERVER`, so if that server doesn't have the channel, `fetch` just fails, rather than trying the other known servers.

Of course, specifying the `host` kwarg works, but I'm lazy :P I'm about to propose a PR that is one way to fix this.

e-q:
This just removes the polling of the default host list from the conditional that checks for the `$NDSSERVER` environment variable, to fix #170.

I think this should be harmless to other users, but I could be wrong.

### DIFF
--- a/gwpy/io/nds.py
+++ b/gwpy/io/nds.py
@@ -123,20 +123,19 @@ def host_resolution_order(ifo, env='NDSSERVER', epoch='now',
                 port = int(port)
             if (host, port) not in hosts:
                 hosts.append((host, port))
-    # otherwise, return the server for this IFO and the backup at CIT
+    # If that host fails, return the server for this IFO and the backup at CIT
+    if to_gps('now') - to_gps(epoch) > lookback:
+        ifolist = [None, ifo]
     else:
-        if to_gps('now') - to_gps(epoch) > lookback:
-            ifolist = [None, ifo]
+        ifolist = [ifo, None]
+    for difo in ifolist:
+        try:
+            host, port = DEFAULT_HOSTS[difo]
+        except KeyError:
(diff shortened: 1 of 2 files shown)


e-q:
Naturally, a bunch of tests that check the output of this function have failed. Oops. 
I can clean these up soon.

duncanmmacleod:
@e-q: this looks fine in principle, as long as the `NDSSERVER` contents get searched first, which I'm pretty sure is what you have. The build/test on travis-ci.org failed because of a couple of random things, but also because your change causes a test of the `host_resolution_order` method to fail.

Can you update the tests under `/gwpy/tests/test_io.py` to pass with your new code.

You can run the test suite before pushing as

```bash
py.test -v gwpy/tests/test_io.py
```

from the root of the repo (you probably have to `pip install pytest` first).

duncanmmacleod:
Just saw that you already noted the test failure, my bad, carry on, look forward to a new push.

e-q:
Weird, all the tests pass on my local copy. Any ideas what the problem is? Something with travis, maybe?

duncanmmacleod:
ok @e-q, this seems to pass the tests just fine. Am I ok to merge?

e-q:
Yep, thanks!

duncanmmacleod:
No, thank you.

Issue #122, opened 2015-07-22
Using "Jet" colormap in spectrograms can mislead the eye


e-q:
The current default colormap in matplotlib is "jet," which has the potential to mislead the eye into seeing features that the data may not contain. When we look at a colormap, we expect changes in lightness to be proportional to changes in the data, but this isn't the case in jet. This strikes me as especially important when looking for transients/glitches in our interferometer data. 

Because of this issue with jet, matplotlib itself is going to be changing their default colormap in v1.5, and they've put together a page with more detail about this issue, with examples of the new default: http://bids.github.io/colormap/

So, for users that upgrade themselves to matplotlib 1.5, this will automatically change. However, it may be advantageous to specify a more perceptually uniform colormap within the gwpy code for user who do not.

duncanmmacleod:
Thanks for bringing this up @e-q. I was already planning to make some test images using the new viridis colour map, and if people like it, I'm open to changing the default. It will take some convincing to get it past some users more resistant to change.

PR #173, opened 2015-11-05
Improved compatibility with matplotlib 1.5

duncanmmacleod:
Matplotlib 1.5 introduces a bunch of new colormaps, and some other changes to setting colors. This PR removes all explicit references to the 'jet' colormap in favour of whatever the default is in the `rcParams` structure ('jet' for mpl 1.x, 'viridis' for mpl 2.0).

This PR also fixes a quirk in the `SegmentAxes.draw` meaning mpl 1.5 would draw blue boxes around tick labels.

### DIFF
--- a/gwpy/plotter/__init__.py
+++ b/gwpy/plotter/__init__.py
@@ -23,7 +23,7 @@ all be easily visualised using the relevant plotting objects, with
 many configurable parameters both interactive, and in saving to disk.
 """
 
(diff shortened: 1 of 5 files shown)


duncanmmacleod:
This PR should fix #122, at least in that it removes explicit dependencies on jet, and will use the default cmap 'viridis' in mpl>=2.0.

PR #181, opened 2015-11-13
Cross spectral density

paulaltin:
Adding support for calculating cross spectral densities using new `csd` method, as well as in spectrograms using the `cross` keyword argument.

### DIFF
--- a/gwpy/spectrum/scipy_.py
+++ b/gwpy/spectrum/scipy_.py
@@ -79,3 +79,21 @@ def rayleigh(timeseries, segmentlength, noverlap=0, **kwargs):
                     name='Rayleigh spectrum of %s' % timeseries.name)
 
 register_method(rayleigh)
+
+
+def csd(timeseries, othertimeseries, segmentlength, noverlap=None, **kwargs):
+    """Calculate the CSD using scipy's csd method (which uses Welch's method)
+    """
+    # get module
+    signal = import_method_dependency('scipy.signal')
+    # calculate CSD
+    f, csd_ = signal.csd(timeseries.value, othertimeseries.value, noverlap=noverlap,
+                         fs=timeseries.sample_rate.decompose().value,
+                         nperseg=segmentlength, **kwargs)
+    # generate Spectrum and return
+    unit = scale_timeseries_units(timeseries.unit,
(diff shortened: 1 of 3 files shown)


duncanmmacleod:
thanks for this @paulaltin, the only request I have is to move the `cross` kwarg further down the list of kwargs for the updated `TimeSeries.spectrogram` method. I want users to be able to run a spectrogram as

```python
timeseries.spectrogram(4, 2, 1)
```

i.e. specifying stride, fftlength and overlap without keyword arguments. Can you move that kwarg further down the list? Actually, I might like to hide that kwarg inside the `**kwargs` dict, to encourage users to use the `csd_spectrogram`, and have the cross functionality in the main `spectrogram` method as a hidden option.

paulaltin:
OK, that should be fixed now. Thanks for pointing it out, I didn't even realise you could do that :-)

duncanmmacleod:
@paulaltin: thanks, it'll be interesting to see if this fixes the problem with the test suite failing. Also, could you write up a unit test for these new methods and add them to `/gwpy/tests/test_timeseries.py`? If you're not sure about it I can take care of it once this PR is merged.

duncanmmacleod:
ok, I found the problem with the unit tests. You have modified the multiprocessed function `_specgram()` defined [here](https://github.com/paulaltin/gwpy/blob/csd/gwpy/timeseries/timeseries.py#L471) but haven't modified the calls [here](https://github.com/paulaltin/gwpy/blob/csd/gwpy/timeseries/timeseries.py#L483). This is causing the sub-processes to crash resulting in the main processes hanging for some reason. If you can fix the calls to pass a third argument, that should resolve this problem.

duncanmmacleod:
you probably didn't catch this because you haven't tested the new methods while using `nproc > 1`.

paulaltin:
OK, makes sense. But I don't really understand that syntax at all, sorry – what do I need to add to `args`? Is is just `, cross`, or do I need that stuff in brackets as well?

paulaltin:
Ah right, I think I get it now.

duncanmmacleod:
ok, this looks great now, thanks @paulaltin

PR #187, opened 2015-11-16
Bug in bartlett window for spectrograms

### LINKED ISSUE #186: Bug with `bartlett` window in spectrogram
There seems to be a bug when using the `bartlett` window to make a `spectrogram`. When I run the example on this page: https://gwpy.github.io/docs/v0.1/examples/spectrogram/plot.html with `method='bartlett'` (and without an `overlap` argument), it returns `TypeError: welch() got multiple values for keyword argument 'noverlap'`.

​I think this can be fixed by adding `kwargs.pop('noverlap', None)` to line 57 of spectrum/scipy_.py.

paulaltin:
Fixes #186 which would raise a TypeError when using the bartlett window to make a spectrogram.


duncanmmacleod:
thanks @paulaltin

PR #192, opened 2016-01-08
table/io/omicron: Fix bug reading triggers in ROOT file

andrew-lundgren:
This tries to fix a bug in the code that reads triggers from Omicron's ROOT format. The GetEntry method in ROOT needs to be given an index. As it is now, it just gets the same trigger (probably the first, by default) repeatedly.

The fix is simple, but I don't have a good development install of gwpy so I haven't been able to test it.


duncanmmacleod:
@andrew-lundgren, the change builds and runs fine, thanks for this addition.

PR #204, opened 2016-02-09
Get FFTW from apt for travis build-and-test

duncanmmacleod:
This PR attempts to get fftw from apt, rather than building from src.


duncanmmacleod:
test fails because fftw3 is required.

PR #209, opened 2016-02-18
Fixed bug in `Plot`, build manager manually if needed

### LINKED ISSUE #207: TimeSeries for plot is missing the module named externals
same as I follow the page https://gwpy.github.io/docs/stable/examples/timeseries/public.html

I do it in the jupyter

Everything is fine, I grab the data from the site, and parse the text as a list of floats, and generate a TimeSeries by supplying the necessary metadata.

but when I 
plot = ts.plot()
it failed, and message said,
ImportError: No module named externals

I am sure I have installed all required packages and moreover I also tested the 
matplotlib in the jupyter for a simple test, it is fine.

default

duncanmmacleod:
This PR fixes #207 by creating a figure manager manually if the `new_figure_manager_given_figure` method isn't present in the `backend_mod`.

### DIFF
--- a/gwpy/plotter/core.py
+++ b/gwpy/plotter/core.py
@@ -23,6 +23,7 @@ import numpy
 
 from matplotlib import (backends, figure, pyplot, colors as mcolors,
                         _pylab_helpers)
+from matplotlib.backend_bases import FigureManagerBase
 from matplotlib.axes import SubplotBase
 from matplotlib.cbook import iterable
 from matplotlib.ticker import LogLocator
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@orozcohsu can you check out this PR branch and confirm that this fixes the problem you are having with jupyter?

orozcohsu:
@duncanmmacleod  
thank you, the problem seems to have been resolved
I can see the graph
default

I check out #209 not #207 you mentioned

but a little strange, 
I type 
**plot = ts.plot() **
default

it shows the graph
I though that is going to generate graph after I type **plot.show()**

PR #235, opened 2016-03-25
spectrum/core.py: add ifft method

### LINKED ISSUE #234: Spectrum should have ifft method
The `TimeSeries` class has an `fft` method. It would be nice if the `Spectrum` had an `ifft`method which served as an inverse. Pull request incoming.

cpankow:
Fixes #234 . I've tested that this returns almost what we want with two issues:

 * I'm having trouble setting the sample rate from `df`, so I set `dx` instead, but the units come out as inverse Hz. Not ideal.
 * I'm struggling to understand the normalization convention used in `TimeSeries.fft`, so the output of `ifft` is not currently normalized correctly.

### DIFF
--- a/gwpy/spectrum/core.py
+++ b/gwpy/spectrum/core.py
@@ -159,6 +159,9 @@ class Spectrum(Series):
         from ..timeseries import TimeSeries
         nout = (self.size - 1) * 2
         # Undo normalization from TimeSeries.fft
+        # The DC component does not have the factor of two applied
+        # so we account for it here
+        dift[0] *= 2
         dift = npfft.irfft(self.value * nout / 2)
         new = TimeSeries(dift, epoch=self.epoch, channel=self.channel,
                        unit=self.unit * units.Hertz, dx=1/self.dx/nout)
--- a/gwpy/spectrum/core.py
+++ b/gwpy/spectrum/core.py
@@ -22,6 +22,7 @@
 import warnings
 from copy import deepcopy
 
+from numpy import fft as npfft
 from scipy import signal
 
 from astropy import units
@@ -133,17 +134,9 @@ class Spectrum(Series):
         from ..plotter import SpectrumPlot
         return SpectrumPlot(self, **kwargs)
 
-    def ifft(self, nfft=None):
+    def ifft(self):
         """Compute the one-dimensional discrete inverse Fourier
         transform of this `Spectrum`.
-
-        Parameters
-        ----------
-        nfft : `int`, optional
-            length of the desired Fourier transform.
-            Input will be cropped or padded to match the desired length.
-            If nfft is not given, the length of the `Spectrum`
-            will be used
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
I can't run this new method:

```python
>>> a = TimeSeries.fetch('L1:GDS-CALIB_STRAIN,rds', 1126259430.0, 1126259434.0)
>>> b = a.fft()
>>> c = b.ifft()
  File "/Users/duncan/Library/Python/2.7/lib/python/site-packages/gwpy-build-py2.7.egg/gwpy/spectrum/core.py", line 166, in ifft
    dift = npfft.irfft(self.value)
NameError: global name 'npfft' is not defined
```

Please test until `a.fft().ifft() == a` within numerical precision.

duncanmmacleod:
The following patch should give you what you need:

```diff
diff --git a/gwpy/spectrum/core.py b/gwpy/spectrum/core.py
index cdddad5..9a3ebff 100644
--- a/gwpy/spectrum/core.py
+++ b/gwpy/spectrum/core.py
@@ -22,6 +22,8 @@
 import warnings
 from copy import deepcopy
 
+from numpy import fft as npfft
+
 from scipy import signal
 
 from astropy import units
@@ -163,10 +165,13 @@ class Spectrum(Series):
         correct.
         """
         from ..timeseries import TimeSeries
-        dift = npfft.irfft(self.value)
-        nout = (len(dft) - 1) * 2
-        new = TimeSeries(dift, epoch=self.epoch, channel=self.channel,
-                       unit=self.unit, dx=1/self.dx/nout)
+        nout = (self.size - 1) * 2
+        x = self.value * nout
+        x[1:] /= 2.0
+        dft = npfft.irfft(x)
+        new = TimeSeries(dft, epoch=self.epoch, channel=self.channel,
+                         unit=self.unit * units.Hertz,
+                         epoch=self.epoch, dx=1/self.dx.value/nout)
         return new
 
     def zpk(self, zeros, poles, gain):
```

The only thing not recovered is the epoch of the `TimeSeries`, but that's due to a bug in the `Spectrum` which I can fix alongside.

duncanmmacleod:
@cpankow: the normalisation in `TimeSeries.fft` doesn't apply `*2` to the zero-frequency element, so we shouldn't do that here:

```diff
diff --git a/gwpy/spectrum/core.py b/gwpy/spectrum/core.py
index ce77252..870d0d6 100644
--- a/gwpy/spectrum/core.py
+++ b/gwpy/spectrum/core.py
@@ -159,7 +159,9 @@ class Spectrum(Series):
         from ..timeseries import TimeSeries
         nout = (self.size - 1) * 2
         # Undo normalization from TimeSeries.fft
-        dift = npfft.irfft(self.value * nout / 2)
+        x = self.value * nout
+        x[1:] /= 2.
+        dift = npfft.irfft(x)
         new = TimeSeries(dift, epoch=self.epoch, channel=self.channel,
                        unit=self.unit * units.Hertz, dx=1/self.dx/nout)
         return new
```

duncanmmacleod:
thanks @cpankow!

PR #255, opened 2016-04-12
New TimeSeries.q_transform method

duncanmmacleod:
This PR introduces a gwpy implementation to the original-MATLAB Q-transform scan, via the `TimeSeries.q_transform` instance method. Included are tests and an example, which produces this image:

qscan

**[NB: these data are from the S6 blind injection, and do not contain an astrophysical signal]**

Thanks to @scottcoughlin2014 for the original MATLAB-python conversion as part of the [GravitySpy](//github.com/mzevin1/GravitySpy/commits/PyOmega) project

### DIFF
--- a/examples/timeseries/qscan.py
+++ b/examples/timeseries/qscan.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python
+
+# Copyright (C) Duncan Macleod (2013)
+#
+# This file is part of GWpy.
+#
+# GWpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


duncanmmacleod:
@andrew-lundgren, you should be able to check out this PR branch if you want to play around. It's a bit rough as it is, but works pretty well. Please propose changes as issues or PRs.

duncanmmacleod:
the coverage decrease is mainly because we can't actually get to the data from the travis-ci.org nodes (no authentication). This should be fixed in a subsequent PR to use the new `TimeSeries.fetch_open_data` for testing.

PR #262, opened 2016-04-16
Fix Jupyter ShimWarnings - bump astropy version.

andrew-lundgren:
There are a lot of warnings when using jupyter. According to https://github.com/astropy/astropy/issues/4169 , this was fixed in astropy 1.0.5. I've tested that upgrading astropy directly using pip fixes the problem, but I haven't tested a clean install.


duncanmmacleod:
thanks @andrew-lundgren

PR #263, opened 2016-04-16
timeseries.whiten: don't produce NaN when zeros in asd

andrew-lundgren:
timeseries.whiten: don't produce NaN when zeros in asd

I ran into trouble in q_transform due to NaNs in the whitened data. This was happening because some values of the asd were zero, making the inverse blow up. The patch deals with this the simplest way, just replacing all zero values with the minimum positive value.

### DIFF
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1422,6 +1422,9 @@ class TimeSeries(TimeSeriesBase):
                            method=method, window=window, **kwargs)
         if isinstance(asd, units.Quantity):
             asd = asd.value
+ 
+        # remove zeros from asd before inverting - replace with min of asd
+        asd[asd==0.] = asd[asd>0.].min()
         invasd = 1. / asd
         # build window
         nfft = int((fftlength * self.sample_rate).decompose().value)


andrew-lundgren:
Split single-tile Q transform into its own function and add a test plot

The second patch (Split single-tile Q transform) splits part of the whole q_transform method into its own function, to make it easier to test and to be used in other code. The third patch (Make a plot) uses this to do a test that the Q transform reports the correct f and Q, and can be used as a base for doing other investigations.

P.S. Apparently pushing commits to a branch adds them to the pull request. I didn't mean the second and third commits to be part of the same pull request. If they need to be split up, I can take care of that.

duncanmmacleod:
@andrew-lundgren, please split up your pull requests, especially since this second addition causes the build and test suite to fail.

duncanmmacleod:
I'm not 100% convinced about this solution, as it will end up hiding a valid mathematical complaint from the user. It might just be easier to replace nans in the `TimeSeries.q_transform` method when they come out the other end.

andrew-lundgren:
You're right that it's a valid complaint. The ASD should not be zero anywhere (except f=0 and the Nyquist bin, possibly) for any kind of real data. I think this is a bug in the LAL spectrum routine, or its Python binding, that makes it have bins with zero power. That should be fixed.

But if an ASD with zeros is used for whitening, then the result will be entirely infinities and NaNs. There's no way to just replace a few bad values in the `q_transform`, they'll all be bad.

duncanmmacleod:
@andrew-lundgren, I think you misunderstand what I'm aiming at. I mean to not replace zeroes in the ASD in `TimeSeries.whiten`, but do the replacement internal to `TimeSeries.q_transform` before the ASD gets passed into `whiten()`. In that way the zeroes don't impact the whitening inside the Q-transform, but also independent use of `whiten` presents warnings/errors where appropriate.

duncanmmacleod:
this has been inactive for a while, I will close now, but would be happy to receive a fresh pull request for this issue in the future

PR #266, opened 2016-04-18
TimeSeries.q_transform: improved frequency recovery

### LINKED ISSUE #265: qtransform spectrogram seems to misrepresent frequencies
I tried doing a Qscan of a few sine-Gaussians in white noise. The resulting spectrogram looks good, but the centers of the injections on the spectrogram are consistently below the central frequency. Here's a test plot, with four injections and lines through the central frequencies:

omega_frequencies

The code to produce this is:

```python
from numpy import *
from numpy.random import normal
from scipy import signal

tlen_sec = 32
srate = 4096
inj_halflen=4
inj_step = 0.25

data = normal(0.,0.1,size=int(srate*tlen_sec))

time = arange(-inj_halflen,inj_halflen, 1./srate)

# Injection parameters
Q = 30.
bw = 3.3245/Q
frequencies = [32.,64.,128.,256.]

injs = []
for num, ff in enumerate(frequencies):
    inj = signal.gausspulse(time, fc=ff, bw=bw)
    idx=(len(data)-len(inj))/2
    offset=idx + num*int(inj_step*srate)
    data[offset:offset+len(inj)] += inj

data = TimeSeries(data, sample_rate=srate, name='fake', epoch=0)

result = data.q_transform(frange=(10,500))

plot = result.crop(tlen_sec/2.- inj_step, tlen_sec/2.+ inj_step*len(frequencies)).plot(figsize=[8, 6])
ax = plot.gca()
ax.set_yscale('log')
for ff in frequencies:
    ax.axhline(ff)
plot.add_colorbar(clim=[0, 400], label='Normalized energy')

plot.show()
```

duncanmmacleod:
interpolation happens at exact frequencies, so need to operate on proper frequencies for both the `QRow` and the desired frequencies, then shift down by `fres/2.` for the output `Spectrogram` bins

Fixes #265. With this code, I get the following equivalent output for the example given in #265:

figure_1

### DIFF
--- a/gwpy/tests/test_timeseries.py
+++ b/gwpy/tests/test_timeseries.py
@@ -565,7 +565,7 @@ class TimeSeriesTestCase(TimeSeriesTestMixin, SeriesTestCase):
             self.assertIsInstance(qspecgram, Spectrogram)
             self.assertTupleEqual(qspecgram.shape, (32000, 2560))
             self.assertAlmostEqual(qspecgram.q, 11.31370849898476)
-            self.assertAlmostEqual(qspecgram.value.max(), 37.158404056005764)
+            self.assertAlmostEqual(qspecgram.value.max(), 37.157012691452067)
 
 
 class StateVectorTestCase(TimeSeriesTestMixin, SeriesTestCase):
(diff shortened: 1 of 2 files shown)


duncanmmacleod:
@andrew-lundgren, please confirm this seems to fix the issue you were seeing.

andrew-lundgren:
@duncanmmacleod Yes, it fixes the issue. I also checked the central times and those look good. One thing we might want to think about is whether the interpolation should be in frequency or log frequency, given how Omega places its tiles. But I think it's good the way it is for now.

PR #330, opened 2016-12-09
Use pytest-xdist to multiprocess tests in travis

duncanmmacleod:
This PR modifies the travis tests to use [`pytest-xdist`](http://pytest.org/dev/xdist.html) to multiprocess the tests with 2 CPUs, allowing (naively) a 50% speed up. This will be most useful on personal/LDG systems where a lot of tests hang on data access.


duncanmmacleod:
So, it seems that multi-processing the tests works fine, but breaks the coverage. Damn.

PR #339, opened 2017-01-11
Low & band pass added to gwpy-plot

areeda:
A mall but useful change


gwpy/cli/timeseries.py
```diff
@@ -39,8 +39,10 @@ def init_cli(self, parser):
         return
 
     def get_ylabel(self, args):
-        """Text for y-axis label"""
-        return 'Counts'
+        """Text for y-axis label,  check if channel defines it"""
+        ret = 'Counts'
+
+        return ret
```
duncanmmacleod: @areeda, what's the motivation behind this change?

areeda: Many channels at LLO now have real units in the frame files. The mechanism to propagate units from the models to the frames is working now.  I thought I could get real units from  NDS but not yet so I backed out the changes.

PR #342, opened 2017-01-14
Added --style command-line option to gwpy-plot

duncanmmacleod:
This PR adds a `--style` option for `gwpy-plot` documented as

```
  --style FILE          path to custom matplotlib style sheet, see
                        http://matplotlib.org/users/style_sheets.html#style-
                        sheets for details of how to write one
```

The argument is a path to a matplotlib style sheet, as documented [here](http://matplotlib.org/users/style_sheets.html#style-sheets). This will allow users to customise their plots ad nauseam without manually editing python code.

In line with this I have removed the custom `rcParams` settings in `gwpy.cli.cliproduct`. These can be reinstated (e.g. on LDVW) with the following style file

```
figure.dpi: 100
font.family: sans-serif
font.size: 16
font.weight: book
legend.loc: best
lines.linewidth: 1.5
text.usetex: True
agg.path.chunksize: 10000
```

### DIFF
--- a/bin/gwpy-plot
+++ b/bin/gwpy-plot
@@ -127,6 +127,8 @@ if __name__ == '__main__':
     # If they requested interactive mode and run from ipython
     # this makes it easier
     if prod.is_interactive:
+        # import pyplot so that user has access to it
+        from matplotlib import pyplot as plt  # pylint: disable=unused-import
         plot = prod.plot
         # pull raw data and plotted results from product for their use
         timeseries = prod.timeseries
--- a/bin/gwpy-plot
+++ b/bin/gwpy-plot
@@ -115,11 +115,18 @@ if __name__ == '__main__':
                            "Please try again with --help.")
     prod = actions[args.func]
     prod.log(2, ('%s called' % args.func))
+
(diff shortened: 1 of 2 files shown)


areeda:
Looking at the changesI think this is a very usefull enhancement. A couple of questions though:

I don't see how or where the default style is specified.

I'm not sure why the import ( from matplotlib import pyplot as plt) was removed. Perhaps it was not apparent what this is for. The idea (perhaps crazy) was that this program could be run from ipython and the plot object manipulated directly. Since most of the documentation and tutorials that I've read abbreviate pyplot as plt I thought that would be convenient.

The only one I'm concerned about is the default style I suppose that can be specified in the external file but we would have to add that to the installation.

duncanmmacleod:
@areeda, the default style is specified in [`/gwpy/plotter/__init__.py`](https://github.com/duncanmmacleod/gwpy/blob/8c6fe045e531183c61285176e3024fb7001ea494/gwpy/plotter/__init__.py#L45). Any customisation would then be done on the command line, and I would encourage you to have such a file for LDVW on your end.

PR #355, opened 2017-01-30
Working version of gwpy-wav

andrew-lundgren:
A standalone command-line tool for grabbing some data from a channel and outputting a wave file. Can whiten or band-pass, and also change the output sample rate. Next I plan to implement linear pitch-shifting.

### DIFF
--- a/bin/gwpy-wav
+++ b/bin/gwpy-wav
@@ -39,14 +39,15 @@ parser.add_argument('-w','--whiten',action='store_true',help='Whiten the data ag
 parser.add_argument('-l','--flow',type=float,default=20.,help='Low-frequency cutoff (Hz). Defaults to 20 Hz because that is the bottom of the human hearing range.')
 parser.add_argument('-u','--fhigh',type=float,help='High-frequency cutoff (Hz).')
 parser.add_argument('-r','--change-sample-rate',type=float,default=1.,help='Multiply sample rate by a factor to speed up the resulting sound.')
-#parser.add_argument('-p','--pitch-shift',type=float,description='Apply a linear pitch shift (Hz).')
+parser.add_argument('-p','--pitch-shift',type=float,help='Apply a linear pitch shift (Hz).')
+parser.add_argument('--pad',type=int,default=2,help='Padding in integer seconds of extra data to request in order to do the filtering. ASD used for whitening is twice this length. Longer values can also make bandpassing more effective.')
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@andrew-lundgren, thanks for this. Two questions

- why is the `TimeSeries.whiten` method not being used? If you have an alternate whitening method, we should be able to integrate it into `TimeSeries.whiten` as an option
- can the pitch shifting be made into a method of the `TimeSeries` object?

Also, please run `pep8 bin/gwpy-wav` against your code and patch to match the formatting guidelines.

duncanmmacleod:
While I'm thinking about it, can the `wavefile.write` operation be integrated into `TimeSeries.write`?

duncanmmacleod:
also, please run `pyflakes` to expose unused imports

andrew-lundgren:
I want to apply band-passing and whitening at the same time, and then possibly pitch-shift in the frequency domain. It's easier to just write the operations out explicitly rather than use the gwpy whiten method.
We could probably make the wavfile.write part of TimeSeries. Should it include auto-scaling of the loudness?

bin/gwpy-wav
```diff
@@ -0,0 +1,113 @@
...
+from numpy import roll
+from numpy.fft import rfft, irfft
+from scipy.signal import firwin, hann
+from scipy.io import wavfile
+
+__author__ = 'andrew lundgren'
+__email__ = 'andrew.lundgren@ligo.org'
+
+parser = argparse.ArgumentParser(description=__doc__, prog='gwpy-wav')
+parser.add_argument('-c', '--chan', type=str, required=True,
```
duncanmmacleod: why are these options not arguments? i.e. use

```python
parser.add_argument('channel', help='Channel name')
```

andrew-lundgren: Wouldn't that make them positional arguments? I've done my best to match the command-line options for gwpy-plot.

duncanmmacleod: yes, positional arguments should be fine (we can make things work with LDVW with positional arguments)

duncanmmacleod:
@andrew-lundgren, this PR has sat untouched for a while, so I'm going to close it. If you still think this tool would be a good addition to gwpy, please open a fresh PR.

areeda:
@andrew-lundgren I'm working on a big re-factoring of LDVW that will require moving the Java code that creates .WAV files out of the tomcat servlets into a standalone program. I will look at making this part of LDVW.

PR #356, opened 2017-01-30
Additions to timeseries.py for Audio editing

dethodav:
Added in new addition to the timeseries class that allows for the creation of .wav files - wavwrite(). fshift() adds two methods to frequency shift the time series, while time_expand_central_freq() changes the playback time of the time series while choosing a frequency to preserve.


duncanmmacleod:
@dethodav, this PR has sat untouched for a while; if you still think this would be a useful addition to gwpy, please open a fresh PR.

PR #365, opened 2017-02-08
Addition of gwpy-sounds executable

dethodav:
The executable gwpy-sounds is based on gwpy-plot but produces audio files instead of plots. It takes advantage of the prebuilt infrastructure inside of cliproduct.py with a few additions that are audio specific. Note that as gwpy-sounds uses the exact same methods as gwpy-plot, the two executables could be merged simply by adding exceptions in cliproduct.MakePlot to account for the differences in audio and plot production.  Features that were previously introduced to the timeseries class in PR #356 (fshift and wavwrite) are and contained and utilized in this PR.


duncanmmacleod:
@dethodav, please look into the build failure, which is a result of your changes. You can run the test suit on your own machine from the root of the gwpy repository as follows:

```bash
py.test-2.7 -v gwpy/tests/
```

areeda:
I have looked at the pull request and just have a few comments.

The main program seems a little awkward in that we have to specify the audio command but there are no others, unlike the plot program where the command specifies one of many types of plots. This does allow us to add more commands to this program and it might be necessary, maybe, to use the command line builders as they exist.

In the cliproduct.py file the addition of secperfftw and overlapw seems unnecessary and it violates the design criterion, minimizing duplication to maintain consistency of the UI. Using the arg_freq method would be my choice.

I did not review the audio.py or the changes to timeseries.py

When I saw looked pretty good to me can't wait to try. I've also looked into adding ffmpeg to the LDVW machine. It will require permission from Stuart that another repo I don't think that's a problem. That should allow us to include MP3 as well as WAV formats. I doubt very much that the quality difference will be noticeable.

One thing we may want to consider for the next release could be to accept two channels and generate a stereo output file. The best audio I heard of the first detection did that with the strain channel from Hanford and Livingston.

dethodav:
@areeda I resolved the issue with the overlap in secpfft by setting up a new set of arguments for timeseries-like plots that prevents multiple cases of secpfft being used. Also, to resolve the "gwpy-sounds audio" command awkwardness, we could just incorporate the audio class into gwpy-plot in the same way as the other plot types so that it would be accessed via "gwpy-plot audio". Otherwise it looks like we would need to redesign how gwpy-sounds sets up its parser, which would also make it more difficult to add another class in the future.

duncanmmacleod:
@dethodav, this PR has sat untouched for a while; if you still think this would be a useful addition to gwpy, please open a fresh PR.

PR #369, opened 2017-02-16
Setting up writing functions for wav files

rpfisher:
Base io for wave files, and registering io functions for wav files to the TimeSeries class.

### DIFF
--- a/gwpy/io/wav.py
+++ b/gwpy/io/wav.py
@@ -1,0 +1,129 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) Duncan Macleod (2013)
+#
+# This file is part of GWpy.
+#
+# GWpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# GWpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with GWpy.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Write a `TimeSeries` to a WAV file
+
+"""
+
+from numpy import (savetxt, loadtxt)
+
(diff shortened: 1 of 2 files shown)


gwpy/timeseries/io/wav.py
```diff
@@ -0,0 +1,29 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with GWpy.  If not, see <http://www.gnu.org/licenses/>.
+
+"""This module attaches the WAV input output methods to the TimeSeries.
+"""
+
+from ...types.io import wav
```
duncanmmacleod: @rpfisher does this line work? given that you haven't changed `types.io` I suspect not.

gwpy/io/wav.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+from ..types import TimeSeries
+from .registry import (register_reader, register_writer, register_identifier)
+from .utils import identify_factory
+
+__author__ = "Ryan Fisher <ryan.fisher@ligo.org>"
+
+
+#def read_ascii(filepath, _obj=Series, xcol=0, ycol=1, delimiter=None,
```
duncanmmacleod: @rpfisher, why are these comment blocks here?

rpfisher: Left those comment blocks in to show the relative difference between wav and another io format.  This was to better enable you and Derek to review it.  We can remove the comments after you and Derek review the code if you prefer.

duncanmmacleod:
@rpfisher, I think it would be cleaner to have all of the new code in the `gwpy.timeseries.io.wav` module, rather than separating into `gwpy.io.wav`, can you combine it all together into a single new module?

rpfisher:
I think that would be less well designed for any future extension of the wav io operations.  Specifically, if we want to make it more broadly usable by other Series subclass objects that aren't themselves subclasses of timeseries.

duncanmmacleod:
I take your point, however, since this is currently restricted to the `TimeSeries` object, I would like it merged. In the future if someone proposes the kind of extensions you suggest, it can be separated appropriately.

rpfisher:
This whole PR is too buggy now that Duncan provided a nice simple testing script.  Closing and starting over!

PR #371, opened 2017-02-16
Remove python-cjson from requirements.txt

duncanmmacleod:
This PR removes the `python-cjson` package from `requirements.txt` since it doesn't support python3. It's only an optional dependency for glue anyway, and will hopefully be patched out.


duncanmmacleod:
this is a duplicate of #374.

PR #374, opened 2017-02-21
Don't install cjson/M2Crypto from requirements.txt

duncanmmacleod:
This PR removes `python-cjson` and `M2Crypto` from the `requirements.txt` file. These are dependencies for `glue`, but aren't available for python3.


duncanmmacleod:
This PR needs to wait on ligovirgo/dqsegdb#32, and [lalsuite/glue#5180](https://bugs.ligo.org/redmine/issues/5180).

PR #395, opened 2017-03-09
Improved TimeSeries read() and write() definitions

duncanmmacleod:
This PR simplifies the definitions of `read()` and write()` for the `gwpy.timeseries` classes, including allowing HDF5 to accept `start` and `end` keywords.


duncanmmacleod:
failures from travis-ci are unrelated to this change, so I will merge.

PR #396, opened 2017-03-09
Increase max number of datasets per plot

areeda:
Minor  change, it is an arbitrary constant plus we adjust legend font if there's a lot of them.


gwpy/cli/cliproduct.py
```diff
@@ -659,13 +658,22 @@ def annotate_save_plot(self, arg_list):
...
                 try:
                     leg.remove()
                 except NotImplementedError:
                     leg.set_visible(False)
+            elif self.n_datasets <= 5 and leg:
+                self.plot.gca().legend(loc='best')
+            elif self.n_datasets > 5 and leg:
+                self.plot.subplots_adjust(right=0.75)
+                #self.plot.gca().legend(bbox_to_anchor=(1.0, 1.0))
+                self.plot.gca().legend(bbox_to_anchor=(1.5, 1), loc='right', borderaxespad=0.)
```
duncanmmacleod: @areeda, I think it makes more sense here to use 

```python
self.plot.gca().legend(bbox_to_anchor=(1.001, 1), loc='left', borderaxespad=0)
```

to left-align the legend to the right-hand axes, rather than right-aligning to the edge of the figure (which I think is what you are doing now).

areeda:
2 changes in this PR (I missed one)
1) constant that set the max plots per canvas raised to ridiculous level (no limit)
2) change decision between spectrogram & spectrogram2

areeda:
close was a mistake

areeda:
LDVW changed the default of intensity scaling from percentile to absolute.  This exposed problems with default scaling.  Fixed in this PR.

areeda:
This PR IS being broken into individual updates.

PR #491, opened 2017-07-06
Pedantic changes to the style/syntax in gwpy.cli

duncanmmacleod:
This PR makes some pedantic changes to the style/syntax used in the `gwpy.cli` module, mainly to bring it inline with the visual style used in the rest of the package.

There are (should be) no functional changes to any part of the module.


duncanmmacleod:
I have rebased this pull request against the current `upstream/develop` branch, since it was a bit out of date.

duncanmmacleod:
@areeda, can you review this, please, and give an OK (or otherwise)

PR #493, opened 2017-07-07
Removed debugging print statement in timeseries.io.losc

duncanmmacleod:
This PR removes what a `print` statement that was presumably for debugging something.


duncanmmacleod:
the CI failure is trivially unrelated (there is no way I can imagine that this change would impact anything), so this PR is ok to merge.

PR #494, opened 2017-07-07
Clean up of gwpy.signal.fft.ui

duncanmmacleod:
This PR cleans up the `gwpy.signal.fft.ui` module a wee bit, mainly with reference to hacky checks that we're not including the wrong kwargs in 'rayleigh' or other methods.


duncanmmacleod:
Failures in the CI are unrelated to the new code, but deprecations in numpy 0.13, and similar. So, this can be merged.

PR #503, opened 2017-07-26
Q-transform

areeda:
Added Q-xform  product and removed limits on # of channels


areeda:
I've resolved the merge conflicts but there seems to be some file corruption.  I need to go through my test suite.
Please hold off on the merge request

areeda:
OK we have tested the changes and now I believe we're ready.

duncanmmacleod:
@areeda, fd082ce seems to have replaced some symlinks with copies of files, presuming this wasn't deliberate, it would be nice to undo this.

areeda:
I'm not sure how this happened let alone how to undo it.
Any suggestion to shorten my search?

duncanmmacleod:
@areeda, my recommendation would be to `git rebase -i` to reorder the commits to put fd082ce on top, then `git reset --head HEAD^` to delete it from the history.

However, given that this PR now has 76 commits in it, I am really tempted to get you on the phone and restart this PR fresh on a feature branch, with only the commits that are really functional.

PR #511, opened 2017-08-03
registering GravitySpy fetch functionality and GravitySpyTable Class

scottcoughlin2014:
Improvement to GravitySpyTable class should be made. However, the main thrust of this PR is to a.) get GravitySpy registered with fetch and to b.) have a GravitySpyTable class of some kind.


duncanmmacleod:
@scottcoughlin2014, thanks for this! Are you able to add a test function to the `/gwpy/tests/test_table.py` module? If not, don't worry, we can work together to add one.

scottcoughlin2014:
I will appease robot and fix up a couple things

scottcoughlin2014:
@duncanmmacleod can the PEP8 robot tell me if my syntax is to its liking again? The code I have submitted passes the python2.7 checks!

areeda:
Scotty, most python implementations include pep8. Ii usually run it like
pep8 *.py

scottcoughlin2014:
Hey @areeda good to know! I should add this to Gravity Spy then!

duncanmmacleod:
@scottcoughlin2014, the CI failed because you have included some python3-incompatible code, namely references to `izip_longest`. Please update your code to use `six` to provide cross-version compatibility.

duncanmmacleod:
the coveralls check will continue to fail without the addition of unit tests, but we can worry about that later.

scottcoughlin2014:
@duncanmmacleod will do!

gwpy/table/io/gravityspy.py
```diff
@@ -0,0 +1,111 @@
...
+        e.args = ('pandas is required to download triggers',)
+        raise
+
+    try:
+        from sqlalchemy.engine import create_engine
+    except ImportError as e:
+        e.args = ('sqlalchemy is required to download triggers',)
+        raise
+
+    if (not QUEST_SQL_USER) or (not QUEST_SQL_PASSWORD):
```
duncanmmacleod: @scottcoughlin2014, it makes more sense to me to access these variables directly from `os.environ` at run time (rather than when the module gets imported), so that users can set them whenever they like. Also, does it not make sense to add `user` and `password` keyword arguments so they can be provided directly to the method?

Also, why are these variables called `QUEST_SQL_xxx`, does it make sense at this time to rename them `GRAVITYSPY_SQL_USER` and `GRAVITYSPY_SQL_PASSWORD`, at least for the GWpy interface?

PR #513, opened 2017-08-09
Check 'not PY2' instead of PY3

duncanmmacleod:
This PR modifies usage of `six.PY2` and `six.PY3` so that the relevant statements work for python > 3. Its nice to be ahead of the game.


duncanmmacleod:
CI failure is unrelated, will be resolved by #512.

PR #587, opened 2017-12-06
Removing some of the absurd URL construction and also transitioning to using wget and not ligo.org module.

scottcoughlin2014:
Mostly, this leverages the public Zooniverse URLs for all these omega scans. Come O3 I will need to have a stable way to upload an image to Zooniverse, immediately retrieve the public url for that image and add it to the database to maintain the robust usability of this method. But for O1 and O2 it will be a far faster more effective way to get the data.

Scotty


duncanmmacleod:
@scottcoughlin2014, there are still lingering `./download` references, can you change those all to use `os.path` (or just don't use `./`, which is `posix`-specific).

gwpy/table/gravityspy.py
```diff
@@ -72,47 +73,78 @@ def download(self, **kwargs):
...
+        # If someone wants labelled samples they are
+        # Definitely asking for the training set but
+        # may hve forgotten
+        if LabelledSamples and not TrainingSet:
+            TrainingSet = 1
+
+        # Let us check what columns are needed
+        cols_for_download = ['imgUrl1', 'imgUrl2', 'imgUrl3', 'imgUrl4']
+        cols_for_download_ext = ['Label', 'SampleType', 'ifo', 'uniqueID']
+        duration_values = np.atleast_2d(np.array(['0.5', '1.0', '2.0', '4.0']))
```
duncanmmacleod: @scottcoughlin2014, why not just use 

```python
    duration_values = np.array([['0.5', '1.0', '2.0', '4.0']])
```
?

duncanmmacleod:
@scottcoughlin2014, it looks like `zip_longest` is no longer required, if so, please remove the `import`.

[you can quickly scrape for unused imports etc using `pyflakes`]

duncanmmacleod:
@scottcoughlin2014, are there more commits coming for this PR?

PR #625, opened 2018-01-09
Removed ez_setup.py from repo

duncanmmacleod:
This PR removes the ez_setup.py bootstrap script from the repo, simplifying the `setup.py` script a bit. Users can be expected to have an installation of `setuptools` already.


duncanmmacleod:
hopefully the CI failures from this will be resolved when combined with the changes from #624

PR #661, opened 2018-02-01
Allowing GravitySpyTable to mimic https://gravityspytools.ciera.northwestern.edu/search/

scottcoughlin2014:
Adding an optional download argument to allow for downloading only specific image durations instead of all 4 images for every glitch. In addtion. one can now GravitySpyTable.search(uniqueID='') in order to preform the same query as is avialble in https://gravityspytools.ciera.northwestern.edu/


gwpy/table/gravityspy.py
```diff
@@ -177,3 +189,44 @@ def _download_single_image(url):
...
+        # and can permantly be 'sbc538'
+        username = 'sbc538'
+        url = url + 'username={0}&'.format(username)
+        url = url + 'howmany={0}&'.format(howmany)
+        # zooID can also be permanantly set to zero.
+        zooID = '0'
+        url = url + 'zooid={0}&'.format(zooID)
+        url = url + 'imageid={0}'.format(uniqueID)
+
+        return GravitySpyTable.from_pandas(pandas.read_json(url))
```
duncanmmacleod: @scottcoughlin2014, do we really need to use `pandas` for this construction?

scottcoughlin2014: :( But it was so super fun! Alright give me a sec I am sure astropy can probably do it...but maybe not? :)

duncanmmacleod:
```

...

        with get_readable_fileobj(url) as f:
            return GravitySpyTable(json.load(f))
```

gwpy/table/gravityspy.py
```diff
@@ -177,3 +189,44 @@ def _download_single_image(url):
...
+            '/search/similarity_search_restful_API/?'
+        # The first varaible username for the url is unimportant
+        # and can permantly be 'sbc538'
+        username = 'sbc538'
+        url = url + 'username={0}&'.format(username)
+        url = url + 'howmany={0}&'.format(howmany)
+        # zooID can also be permanantly set to zero.
+        zooID = '0'
+        url = url + 'zooid={0}&'.format(zooID)
+        url = url + 'imageid={0}'.format(uniqueID)
```
duncanmmacleod: it might be cleaner to build the URL using something like

```python
        parts = {
            'username': username,
            'howmany': howmany,
            'zooid': 0,
            'imageid': uniqueID,
        }
        search = '&'.join('{}={}'.format(key, value) for
                          key, value in parts.items())
        url = '{}/?{}'.format(base, search)
```

which might make it easier to modify/add/remove URL query components.

scottcoughlin2014: Excellent will make this change and the above shortly

duncanmmacleod:
@scottcoughlin2014, the new unit tests are failing on LDG reference OSs because there is a bug in astropy. The bug has since been fixed with astropy 2.0.4 but that's not available as a system package on the LDG. So, what you should probably do is add a skip decorator to the `TestGravitySpyTable` class:

```python
@utils.skip_minimum_version('astropy', '2.0.4')
class TestGravitySpyTable(TestTable):
...
```

This will allow your tests to be executed on travis most of the time.

duncanmmacleod:
@scottcoughlin2014, the coverage dropped because of a bug in pytest, but it manifests because of a quirk in the way you have defined your tests. Since `GravitySpyTable` inherits from `EventTable`, `TestGravitySpyTable` should inherit from `TestEventTable`, not `TestTable`. Can you make that change?

That should resolve the coverage problems, I think

duncanmmacleod:
The CI failures are unrelated to this PR, and will be fixed by #671.

PR #672, opened 2018-02-14
Handle unit of '*' from NDS2

areeda:
Detect units of '*' and override so we can calculate spectra

### DIFF
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -806,6 +806,7 @@ class CliProduct(object):
             if (un != 'undef') & (un != '') & (un != '*'):
                 all_units.add(un)
             else:
+                ts.override_unit('')
                 all_units.add('Counts')
 
         if len(all_units) == 1:
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -517,8 +517,8 @@ class CliProduct(object):
                     data = TimeSeries.fetch(chan, start, start+self.dur,
                                             verbose=verb)
 
-                if data.unit in self.BAD_UNITS:
-                    data.override_unit('count')
+                if str(data.unit) in self.BAD_UNITS:
+                    data.override_unit('undef')
 
                 if highpass > 0 and lowpass == 0:
                     data = data.highpass(highpass)
@@ -809,8 +809,10 @@ class CliProduct(object):
 
         if len(all_units) == 1:
             self.units = label_to_latex(all_units.pop())
+        elif len(all_units) > 1:
+            self.units = 'Multiple units'
         else:
-            self.units = 'Counts'
+            self.units = 'undef'
 
         if args.ylabel:
             ylabel = label_to_latex(args.ylabel)
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -58,7 +58,7 @@ class CliProduct(object):
     """
 
     __metaclass__ = abc.ABCMeta
-    BAD_UNITS = {'undef', '', '*'}
+    BAD_UNITS = ['undef', '', '*']
 
     def __init__(self):
 
@@ -518,7 +518,7 @@ class CliProduct(object):
                                             verbose=verb)
 
                 if data.unit in self.BAD_UNITS:
-                    data.override_unit('COUNTS')
+                    data.override_unit('count')
 
                 if highpass > 0 and lowpass == 0:
                     data = data.highpass(highpass)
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@areeda, thanks for the PR. The CI is failing because other tests (unrelated to this change) are out-of-date. Can you rebase your branch against the current `upstream/develop`?

```
git fetch upstream/develop
git rebase -i upstream/develop
```

duncanmmacleod:
@areeda, what's the relevance of `agg` to this PR?

areeda:
Sorry reused a branch to debug, such a small change I lazily didn't create a new one.
I can do that iif you like.

areeda:
Are those CI failures significant?

duncanmmacleod:
@areeda, the CI failures are obscuring the impact of the new additions, so if you can rebase as I posted then `git push --force` the CI should rerun and we might be able to see a bit better.

gwpy/cli/cliproduct.py
```diff
@@ -41,7 +41,7 @@
 
 from dateutil.parser import parser
 
-from ..timeseries import TimeSeries
+from gwpy.timeseries import TimeSeries
```
duncanmmacleod: @areeda, what's the reason for this change?

gwpy/cli/cliproduct.py
```diff
@@ -801,13 +803,14 @@ def annotate_save_plot(self, args):
         all_units = set()
         for ts in self.timeseries:
             un = str(ts.unit)
-            if (un != 'undef') & (un != ''):
+            if (un != 'undef') & (un != '') & (un != '*'):
```
duncanmmacleod: it might be cleaner to set up a list in the module header that contains all of the proscribed units,

```python
BAD_UNITS = {'undef', '', '*'}
```

then just do something like

```python
            if str(ts.unit) in BAD_UNITS:
                ts.override_unit('')
            else:
                all_units.add(str(ts.unit))

File: gwpy/cli/cliproduct.py (hunk 1 above)
Comment (areeda) [reply gwpy/cli/cliproduct.py]: I think it's a tad more obvious plusI can copy/pasteinto ipython
No big deal I can reverse it if you likr

File: gwpy/cli/cliproduct.py (hunk 2 above)
Comment (areeda) [reply gwpy/cli/cliproduct.py]: I've been wondering whether they should be handled at this level at all? It was originally put in there to avoid confusion with the transition. The only real problem is when we do arithmetic on the units and for the most part that is being handled in the timeseries object I believe.

I like your suggestion and I will make the change.

File: gwpy/cli/cliproduct.py (hunk 1 above)
Comment (duncanmmacleod) [reply gwpy/cli/cliproduct.py]: yes, please revert. since this module is part of `gwpy`, it should use relative imports for relative objects as much as possible.

File: gwpy/cli/cliproduct.py [hunk 3]
```diff
@@ -801,13 +803,14 @@ def annotate_save_plot(self, args):
...
                 all_units.add(un)
             else:
+                ts.override_unit('')
                 all_units.add('Counts')
 
         if len(all_units) == 1:
-            self.units = all_units.pop()
+            self.units = label_to_latex(all_units.pop())
         else:
             self.units = 'Counts'
```
duncanmmacleod: @areeda, since 'Counts' is an actual physical unit, I don't think we should use it as the default. Perhaps something like

```python
        if len(all_units) == 1:
            self.units = label_to_latex(all_units.pop())
        elif len(all_units) > 1:
            self.units = 'multiple units'
        else:
            self.units = 'undef'
```

gwpy/cli/cliproduct.py
duncanmmacleod: @areeda, the units parser knows about `'undef'`, and `''` is a legitimate way of specifying 'dimensionless', so please only include `'*'` in your `BAD_UNITS` `set`. We might consider handling that at the unit parser level, as you say.

gwpy/cli/cliproduct.py
```diff
@@ -58,6 +58,7 @@ class CliProduct(object):
     """
 
     __metaclass__ = abc.ABCMeta
+    BAD_UNITS = ['undef', '', '*']
```
duncanmmacleod: @areeda, can you change this to just

```python
    BAD_UNITS = {'*',}
```

since the other two are legitimate unit choices? (also, `a in {set}` is slightly more efficient than `a in {list}` (but only really for large lists).

areeda: these are in latest push

PR #690, opened 2018-02-28
Preserve order when uniquifying start times in gwpy.cli

### LINKED ISSUE #341: The order of plot lines and legend items is not the order given by the user
Using GWpy plots in the ligoDV-web interface I cannot specify the order that items should appear in the legend or the plot itself currently. I would like them to appear in the order I enter them. First entered gets plotted first and is first in the legend. Then in order after that. The attached screenshot shows that the order specified by me was sent to GWpy in the command but the plot came out with a different order (not chronological either). 
ldvw-spectrum-order

duncanmmacleod:
This PR fixes #341 by preserving order when uniquifying the `start_time` list in `gwpy.cli`.

cc: @areeda

### DIFF
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -456,7 +456,11 @@ class CliProduct(object):
                 % self.min_timeseries)
 
         if len(arg_list.start) > 0:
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@areeda, can you take a look at this?

PR #705, opened 2018-03-09
Demodulation method for TimeSeries objects

alurban:
This is my first pass at implementing a demodulation method for `TimeSeries` objects. The proposed method takes in a frequency and stride, and returns a new `TimeSeries` object with `dt=stride` which represents the original signal's average magnitude and phase at the given frequency as a complex number.

Also included in this pull request is a unit test which creates a pure sinusoid and then checks that `demod` correctly measures the amplitude and phase within acceptable precision. (Note, the error in this calculation decreases for arbitrary amplitude and phase as the stride increases.)

### DIFF
--- a/gwpy/tests/test_timeseries.py
+++ b/gwpy/tests/test_timeseries.py
@@ -1168,16 +1168,16 @@ class TestTimeSeries(TestTimeSeriesBase):
         demod = data.demodulate(f, stride=stride, exp=True)
         assert demod.unit == data.unit
         assert len(demod) == duration // stride
-        assert_allclose(numpy.abs(demod.value), amp, rtol=1e-5)
-        assert_allclose(numpy.angle(demod.value), phase, rtol=1e-5)
+        utils.assert_allclose(numpy.abs(demod.value), amp, rtol=1e-5)
+        utils.assert_allclose(numpy.angle(demod.value), phase, rtol=1e-5)
 
         # test with exp=False
         mag, ph = data.demodulate(f, stride=stride)
         assert mag.unit == data.unit
         assert len(mag) == len(ph)
         assert ph.unit == 'deg'
-        assert_allclose(mag.value, amp, rtol=1e-5)
-        assert_allclose(ph.value, numpy.rad2deg(phase), rtol=1e-5)
+        utils.assert_allclose(mag.value, amp, rtol=1e-5)
+        utils.assert_allclose(ph.value, numpy.rad2deg(phase), rtol=1e-5)
 
     def test_whiten(self):
         # create noise with a glitch in it at 1000 Hz
--- a/gwpy/tests/test_timeseries.py
+++ b/gwpy/tests/test_timeseries.py
@@ -1158,14 +1158,16 @@ class TestTimeSeries(TestTimeSeriesBase):
         # create a timeseries that is simply one loud sinusoidal oscillation
         # at a particular frequency, then demodulate at that frequency and
         # recover the amplitude and phase
-        amp, phase, f = 1., 0., 30
-        t = numpy.linspace(0, 10, 10*16384)
-        data = amp * numpy.cos(2*numpy.pi*f*t + phase)
-        data = TimeSeries(data, unit='strain', times=t)
-        assert data.demodulate(f).unit == data.unit
-        assert len(data.demodulate(f)) == 10
-        assert all(x <= 1e-4 for x in numpy.abs(data.demodulate(f).value - amp))
-        assert all(x <= 1e-4 for x in (numpy.angle(data.demodulate(f)) - phase))
+        amp, phase, f = 1., numpy.pi/4, 30
+        duration, sample_rate, stride = 600, 4096, 60
+        t = numpy.linspace(0, duration, duration*sample_rate)
+        data = TimeSeries(amp * numpy.cos(2*numpy.pi*f*t - phase),
+                                   unit='', times=t)
+        demod = data.demodulate(f, stride=stride)
+        assert demod.unit == data.unit
+        assert len(demod) == duration // stride
+        assert_allclose(numpy.abs(demod.value), amp, rtol=1e-5)
+        assert_allclose(numpy.angle(demod.value), phase, rtol=1e-5)
 
     def test_whiten(self):
         # create noise with a glitch in it at 1000 Hz
--- a/gwpy/tests/test_timeseries.py
+++ b/gwpy/tests/test_timeseries.py
@@ -1163,11 +1163,21 @@ class TestTimeSeries(TestTimeSeriesBase):
         t = numpy.linspace(0, duration, duration*sample_rate)
         data = TimeSeries(amp * numpy.cos(2*numpy.pi*f*t - phase),
                                    unit='', times=t)
-        demod = data.demodulate(f, stride=stride)
+
+        # test with exp=True
+        demod = data.demodulate(f, stride=stride, exp=True)
         assert demod.unit == data.unit
         assert len(demod) == duration // stride
         assert_allclose(numpy.abs(demod.value), amp, rtol=1e-5)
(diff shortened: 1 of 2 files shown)


gwpy/timeseries/timeseries.py
```diff
@@ -1290,6 +1290,41 @@ def rms(self, stride=1):
         return self.__class__(data, channel=self.channel, t0=self.t0,
                               name=name, sample_rate=(1/float(stride)))
 
+    def demod(self, f, stride=1):
```
duncanmmacleod: @alurban, why not call it `demodulate()`?

gwpy/timeseries/timeseries.py
```diff
@@ -1290,6 +1290,41 @@ def rms(self, stride=1):
...
+        data = numpy.zeros(nsteps, dtype=complex)
+        mixed = 2 * numpy.exp(2*numpy.pi*1j*f*self.times.value) * self.value
+        for step in range(nsteps):
+            # find step TimeSeries
+            idx = int(stridesamp * step)
+            idx_end = idx + stridesamp
+            stepseries = mixed[idx:idx_end]
+            demod_ = numpy.average(stepseries)
+            data[step] = demod_
+        return self.__class__(data, channel=self.channel, t0=self.t0,
```
duncanmmacleod: @alurban, I would prefer to use `type(self)` rather than `self.__class__`, purely for aesthetic reasons, I think. There is precedent for both in `timeseries.py` already.

gwpy/tests/test_timeseries.py
```diff
@@ -1154,6 +1154,18 @@ def test_rms(self, losc):
...
+        # create a timeseries with two loud lines with different phases
+        # at different frequencies
+        amp, phase, f = 1., 0., 30
+        t = numpy.linspace(0, 10, 10*16384)
+        data = amp * numpy.cos(2*numpy.pi*f*t + phase)
+        data = TimeSeries(data, unit='strain', times=t)
+        assert data.demod(f).unit == data.unit
+        assert len(data.demod(f)) == 10
+        assert all(x <= 1e-4 for x in numpy.abs(data.demod(f).value - amp))
+        assert all(x <= 1e-4 for x in (numpy.angle(data.demod(f)) - phase))
```
duncanmmacleod: @alurban, is the purpose of these `all` calls to assert that the result is close to zero? If so it makes more sense to me to use something like

```python
        data = TimeSeries(amp * numpy.cos(2*numpy.pi*f*t + phase),
                                        unit='', times=t)
        demod = data.demod(f)
        assert demod.size == 10
        assert demod.unit == data.unit
        utils.assert_allclose(numpy.abs(demod.value), amp)
        utils.assert_allclose(numpy.angle(demod.value), phase)
```

gwpy/timeseries/timeseries.py
```diff
@@ -1290,6 +1290,41 @@ def rms(self, stride=1):
...
+
+        stride : `float`
+            stride (seconds) between calculations
+
+        Returns
+        -------
+        demod : `TimeSeries`
+            a new `TimeSeries` containing magnitude and phase trends
+            (stored as a complex number) with dt=stride
+        """
```
duncanmmacleod: @alurban, can you add an example to the bottom of this docstring? See https://github.com/alurban/gwpy/blob/ea10c8076db5bfcabc1c628ac109c0b70c879c7b/gwpy/timeseries/timeseries.py#L1027 for an example of that.

gwpy/timeseries/timeseries.py
alurban: @duncanmmacleod, this is a good suggestion and is taken care of in a new commit (https://github.com/gwpy/gwpy/pull/705/commits/bd0cbc6f3ae1715b6e5f8474746c5900a7335cc0)

gwpy/timeseries/timeseries.py
alurban: I took a stab at this in the following commit: https://github.com/gwpy/gwpy/pull/705/commits/5576230c44e2aefaf0427b41ccd5194d17db2f3c

gwpy/tests/test_timeseries.py
alurban: @duncanmmacleod you're right that this is exactly what I was going for. I was unaware of `assert_allclose` before, it seems perfect for this test. I coded that up in this commit: https://github.com/gwpy/gwpy/pull/705/commits/1977ee56063d3b2fc1c4955ce253738353745736

duncanmmacleod:
@alurban, the CI is failing because `assert_allclose` isn't imported for `test_timeseries.py`. I think you want to use `utils.assert_allclose`, since `gwpy/tests/utils.py` is imported already.

FTR you should be able to test your changes before you push from the root of the repo via

```
python setup.py test --addopts "gwpy/tests/test_timeseries.py::TestTimeSeries::test_demodulate"
```

duncanmmacleod:
@alurban, the new code looks great, and the unit tests are passing. One last (optional) thing (that we can complete in a separate PR, probably) is whether you can come up with a practical example of this method, using open data from LOSC? Maybe something to do with a pulsar injection line, or a calibration line?

alurban:
Aha! That's exactly what I was looking for. Previous to this I was writing the test in a separate directory, then copying that script into the code block for test_demodulate. In doing so I failed to realize that `assert_allclose` was not an imported function. But problems like this should be much easier to avoid in the future, now that I know how to run the unit test from the command line. Thanks much!

alurban:
Regarding an example: yeah I was trying to think about that. I *could* try to compare calibration lines to the PCALY_TX channel, but I can't do that with LOSC data. Doing it with a pulsar injection line sounds like it might be cool though

duncanmmacleod:
@alurban, just because I'm pedantic, can you take care of the one remaining pep8 complaint:

```
$ pep8 gwpy/tests/test_timeseries.py
gwpy/tests/test_timeseries.py:1165:13: E128 continuation line under-indented for visual indent
```

This is reported by @pep8speaks above, but that bot edits the same comment over and again, so its hard to spot new pep8 issues after a few commits.

alurban:
I think I've resolved the pep8 issue, thanks for spotting it. I also noticed I was using the wrong sign convention in my complex oscillator, so I've fixed that as well.

gwpy/timeseries/timeseries.py
```diff
@@ -1290,6 +1290,63 @@ def rms(self, stride=1):
...
+        for step in range(nsteps):
+            # find step TimeSeries
+            idx = int(stridesamp * step)
+            idx_end = idx + stridesamp
+            stepseries = mixed[idx:idx_end]
+            demod_ = numpy.average(stepseries)
+            out.value[step] = demod_
+        if exp:
+            return out
+        else:
```
duncanmmacleod: @alurban, since the `if` statement includes `return`, you don't need the `else` at all. Also, I think I prefer using `override_unit()` compared to `_unit` (using a ternary instead of `if/else`).

If you agree, can you please modify this block to something like:

```python
        if exp:
            return out
        mag = numpy.abs(out)
        phase = numpy.angle(out, deg=deg).view(type(self))
        phase.__metadata_finalize__(out)
        phase.override_unit('deg' if deg else 'rad')
        return mag, phase

alurban:
Ooooh, swanky! I'm glad  you pointed out this functionality, particularly `override_unit`. I've updated this in my last commit and updated the unit test to also check the `deg=False` use case. This change seems a lot cleaner to me.

alurban:
@duncanmmacleod: just so you know, I'm working on an example for the docstring and will hopefully have it by the end of the day. After talking to TJ I thought it would be cool to look at how the 60 Hz line varies in amplitude over a random chunk of time.

duncanmmacleod:
Thanks @alurban!

alurban:
Sweet! Thanks @duncanmmacleod!

Issue #477, opened 2017-06-19
Wishlist: an 'injection' fct for TimeSeries add(data, t) 


ecm0:
it would be convenient to include to the methods of TimeSeries, a function add(data, t) that adds data at time t where data is a numpy array or a TimeSeries and t is a GPS date.

alurban:
@ecm0 I'm going to take this on and start putting together a pull request in the next few days

ecm0:
:+1:

PR #720, opened 2018-03-28
Support for injections in GWpy TimeSeries objects

alurban:
This is a first pass at a method that allows for injections with GWpy.

The pull request includes a new method, called `add()`, which takes in a `TimeSeries` object and sums it to `self` along their overlapping time samples.

If `self` and the new `TimeSeries` have inconsistent sample rates, the method will raise a `ValueError`. If the timeseries have non-overlapping time stamps, the method will raise a warning and return the original (unaltered) `TimeSeries`.

This pull request also includes a unit test which injects a pure sinusoid into silence and then checks that the units, length, and values are all as expected.

Lastly, the method's doctstring includes an example which downloads strain data and a simulated waveform (corresponding to GW150914) from LOSC, then adds them together. I think it would be nice to add a Q-scan to this example showing that the injection was successful, but I'll want to hear your thoughts first. (I know those Q transform plots can be slow, so I don't know how you feel about adding them to the documentation.)

This addresses issue 477 (https://github.com/gwpy/gwpy/issues/477) by @ecm0.

### DIFF
--- a/gwpy/tests/test_timeseries.py
+++ b/gwpy/tests/test_timeseries.py
@@ -1199,6 +1199,22 @@ class TestTimeSeries(TestTimeSeriesBase):
         assert ph.unit == 'rad'
         utils.assert_allclose(ph.value, phase, rtol=1e-5)
 
+    def test_add(self):
+        # create a timeseries out of an array of zeros
+        duration, sample_rate = 1, 4096
+        data = TimeSeries(numpy.zeros(duration*sample_rate), t0=0,
+                          sample_rate=sample_rate, unit='')
+
+        # create a second timeseries to add to the first
+        waveform = numpy.cos(2*numpy.pi*30*data.times.value)
+        waveform = TimeSeries(waveform, times=data.times.value)
+
+        # test that we recover this waveform when we add it to data
+        new_data = data.add(waveform)
+        assert new_data.unit == data.unit
+        assert len(new_data) == len(data)
+        utils.assert_allclose(new_data.value, waveform.value)
+
     def test_whiten(self):
         # create noise with a glitch in it at 1000 Hz
         noise = self.TEST_CLASS(
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1382,6 +1382,32 @@ class TimeSeries(TimeSeriesBase):
 
         Examples
         --------
+        It can often be useful to add a simulated signal to some data
+        stream. For example, we like to quantify our sensitivity to
+        gravitational wave signals by injecting simulated ones into otherwise
+        quiet stretches of data. To see this, we can download some data
+        from LOSC:
+
+        >>> from gwpy.timeseries import TimeSeries
+        >>> strain = TimeSeries.fetch_open_data('H1', 1131350417, 1131350467)
+
+        These data are reasonably quiet, but we can also grab a simulation
+        of what we expect a binary black hole waveform to look like and add
+        it to our strain data:
+
+        >>> import numpy, urllib2
+        >>> source = 'https://losc.ligo.org/s/events/GW150914/P150914/'
+        >>> filename = 'fig2-unfiltered-waveform-H.txt'
+        >>> download = urllib2.urlopen('%s/%s' % (source, filename))
+        >>> t, h = numpy.recfromtxt(download, skip_header=1, unpack=True)
+        >>> h *= 1e-21
+
+        We will need to store this as a `TimeSeries` and downsample to 4096 Hz:
+        >>> waveform = TimeSeries(h, t0=1131350417, sample_rate=16384)
+        >>> waveform = waveform.resample(strain.sample_rate.value)
+
+        Finally, we can add this simulation to our strain data:
+        >>> injected = strain.add(waveform)
 
         Notes
         -----
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1292,7 +1292,7 @@ class TimeSeries(TimeSeriesBase):
 
     def demodulate(self, f, stride=1, exp=False, deg=True):
         """Compute the average magnitude and phase of this `TimeSeries`
-           once per stride at a given frequency.
+        once per stride at a given frequency.
 
         Parameters
         ----------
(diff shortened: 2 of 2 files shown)


alurban:
The bit where it actually adds was in a `for` loop, which seems slow. I found a way to do this with array indexing that might be faster (I haven't done a speed test, but it is at least prettier).

I also failed to mention that there's a minor outdent in the doctstring for `demodulate` included in this pull request, because I too can be pedantic.

alurban:
Question: do we want this method's name to be `__add__()`, so that you invoke it via e.g.
```
new_timeseries = old_timeseries + waveform
```
rather than the current syntax,
```
new_timeseries = old_timeseries.add(waveform)
```
?

duncanmmacleod:
@alurban, thanks for putting this together. Notes

#### Method:

It should be enough to check that the `sample_rate` properties are the same (which you are doing) and that the `other.to` is an integer number of samples away from `self.to`: 

```python
    def add(...):
        self.is_compatible(other)  # checks units and sample size compatibility
        ot0 = other.t0.to(self.t0.unit)
        idx = ((ot0 - self.t0) / self.dt).value
        if not idx.is_integer():
            raise ValueError(....)
        elif idx < 0:
            raise ValueError(...)
        slice_ = slice(idx, idx + other.shape[0])
        out = self.copy()  # this is important!
        out.value[slice_] += other.value
        return out
```

This code is completely untested. One other thing we might want to do is to abstract this away from the time-domain, and make it a method of the `Series` object (which is the parent of `TimeSeries` and `FrequencySeries`) to provide the same functionality for the frequency domain.

#### Docstring:

You go the long way around to getting a `TimeSeries` waveform from LOSC, it should be enough to suggest something like this:

```python
        Create noise `TimeSeries`

        >>> from numpy import random
        >>> from gwpy.timeseries import TimeSeries
        >>> noise = TimeSeries(random.normal(scale=.5, size=16384), sample_rate=16384)

        Download waveform from LOSC:

        >>> from astropy.utils.data import get_readable_fileobj
        >>> url = 'https://losc.ligo.org/s/events/GW150914/P150914/fig2-unfiltered-waveform-H.txt'
        >>> with get_readable_fileobj(url) as f:
        ...     signal = TimeSeries.read(f)
        >>> signal.t0 = .5  # reset start time to line up with noise

        Combine the two:

        >>> data = noise + signal
        >>> plot = data.plot()
        >>> plot.show()
```

#### Tests:

You should also check that the original timeseries is unharmed by the operation, just in case. I suspect that this check would currently fail.

alurban:
@duncanmmacleod I'm wondering if we also want to supply users with a `taper` method. It would be fun to code up, and I could use a Planck-taper window by default? https://arxiv.org/abs/1003.2939

alurban:
@duncanmmacleod I've now included a `taper` method that implements the Planck taper window. (I checked that this is the same window function used in LAL.) This method has two arguments: `tau`, the taper length, and `side`, the side of the waveform to taper. `tau` defaults to 0.05 seconds, while `side` defaults to `'leftright'`, meaning both ends of the `TimeSeries` are tapered.

I also updated the example in the doctstring for `TimeSeries.add` so that the 150914 waveform is tapered. Here's a rendered version of the resulting plot (requires LIGO.ORG credentials): https://ldas-jobs.ligo.caltech.edu/~alexander.urban/gwpy_tests/timeseries.pdf)

**Most importantly:** happy weekend!

gwpy/signal/window.py
```diff
@@ -56,6 +56,8 @@ def canonical_name(name):
     >>> canonical_name('ksr')
     'kaiser'
     """
+    if name == 'planck':  # make sure to handle the Planck window function
```
duncanmmacleod: can you modify this to be

```python
    if name.lower() == 'planck':
```

to allow people to use `name='Planck'`?

gwpy/signal/window.py
```diff
@@ -83,7 +85,8 @@ def canonical_name(name):
     'hann': .5,
     'hamming': .5,
     'nuttall': .656,
-    'triang': .5
+    'triang': .5,
+    'planck': .01
```
duncanmmacleod: for my edification, can you explain the choice of `.01` here?

alurban: It's definitely somewhat arbitrary. By default, the `TimeSeries.taper()` method which wraps around `signal.window.planck()` automatically tapers to the second stationary point from either side of the signal, so its taper length can vary wildly. To come up with `.01`, I chose the duration of one cycle of a sinusoid at 100 Hz, which is roughly on the same order as what we would commonly taper out of our waveforms.

gwpy/signal/window.py
alurban: done in commit 21f24d31b156aa1cf435496a6d85b43c95efdaaf

gwpy/signal/window.py
duncanmmacleod: if the recommended overlap for this window is not actually known, I would rather omit it from this mapping, so that the user needs to manually select an overlap.

alurban: I'm happy enough with that setup. That has been done in commit f58afa24937a87d82ab7a664b488f2807c97a51f

ecm0:
Hi Alex, Duncan,

I did a few tests of the new injection mechanism in gwpy. There is something
misleading in the example given here: https://github.com/gwpy/gwpy/pull/720#issuecomment-377200522

You set `signal.t0 = .5` which I've understood as: I'm injecting signal at the time position t = 0.5 s
in noise. This is not what happens.

Instead, if signal.t0 = .5 and noise.t0 = 0 then one get different data.t0 if data = signal + noise or noise + signal. This is confusing and will likely lead to bugs.

I see two options -- one simple and straightforward but with limited functionality, the other request more development but is much more useful:

1. Request strict timeseries alignment before adding: t0, duration and sampling frequency should be equal

2. Request that the sampling freqs are equal and that the time support of the shortest timeseries is in complete overlap with the longest one which then fixes the t0 and duration of the result. The short timeseries is added at the position defined by its t0 (ie., from the closest index idx = np.round((ts2.t0-ts1.t0)/sampling_freq) in the Numpy array). 

data = signal + noise or data = noise + signal would produce the same result with Solution 2 and
it would allow to handle simply injections in most cases.

Does that make sense? What do you think?

alurban:
Hi Eric (@ecm0), thanks for your comment, it's very nice to hear from you as I believe you are the user who originally requested this functionality.

I should note that @duncanmmacleod's comment that you link to above is not the final use case that we ended up going with. For a worked example of that, I'll point you to [**this page**](https://gwpy.github.io/docs/stable/examples/timeseries/inject.html) in the GWPy documentation, which reflects the latest stable release (at the time of writing, 0.12.2). The short version is that, if you have two series objects `hoft` and `waveform` which represent raw data and a signal model, respectively, then you would perform an injection like this:

```python
hoft = ....  # prepare raw data
waveform =  # prepare signal model
waveform = waveform.taper()  # taper with the Planck-taper window
injected = hoft.inject(waveform)
```

This operation will add the two series, but only along their shared x-axis samples, which will be a (not necessarily proper) subset of `hoft.xindex`. If the (discrete) set of x-axis samples do not intersect, i.e. if the x-axis limits are disjoint or if the samples are offset by a non-integer multiple of the sample rate, then a warning will be raised and a copy of `hoft` will be returned. The objects `hoft` and `waveform` are otherwise free to have different `x0`.

Also, note that since it's now written as a method of the parent class, it will work the same in both time and frequency domain. The source code for this can be found [**here**](https://github.com/gwpy/gwpy/blob/develop/gwpy/types/series.py#L975-L991) and an example of a frequency-domain injection can be found [**here**](https://gwpy.github.io/docs/stable/examples/frequencyseries/inject.html).

I believe that this removes the ambiguity you're concerned about and is much more in line with what you're suggesting here, but please let me know if I'm mistaken.

ecm0:
Hi Alex, thanks for the quick feedback. I haven't tried this yet, but yes, this seems to resolve my concerns and looks like a solution that will work for the simulation I'm trying to do. Thanks!

alurban:
No problem @ecm0! Thanks for using this functionality 😄

We wound up calling the method `inject()` and not `__add__()` because, as you pointed out, the operation is not commutative. This way there's no ambiguity about which series is being added to which, but they can still have different durations.

PR #728, opened 2018-04-05
Support for frequency domain injections

alurban:
@duncanmmacleod This PR adds a method, `Series.inject()`, which allows for both time and frequency domain injections. Its invocation replaces `TimeSeries.inject()` (see pull request https://github.com/gwpy/gwpy/pull/720) by moving the `inject()` method into the parent `Series` class. In the frequency domain, the behavior is designed such that `other.frequencies` must be a subset of `self.frequencies`. In other words, signals that trail off past `self.f0` or `self.f0 + self.size*self.df` will raise a `ValueError`. The method's behavior in the time domain remains unchanged.

This PR also adds a unit test and two example use cases, which would be added to the `examples` directory.


alurban:
Note, the issue reported by @pep8speaks is on a line not touched by this pull request.

duncanmmacleod:
@alurban, this PR begs the question of whether the relevant inject method should be moved upstream to the `Series`, which is the parent class for both the `TimeSeries` and `FrequencySeries`, so we don't have to maintain two independent implementations of essentially the same method. Can you have a go at making it xunit-agnostic and moving it up?

alurban:
@duncanmmacleod, sure, I'll have a go at that this afternoon. It should be ready for you to look over tomorrow morning your time

alurban:
I've taken a stab at making the method xunit-agnostic. The one small hitch is that, if the `Series` are in the time domain, it will attempt to crop them to reflect the behavior in #720 . This currently gives the method a complexity of 6 which fails the codeclimate test. Are you ok with that?

I've also moved injection examples over to `examples/`, which seemed like the appropriate place to put them.

alurban:
I've managed to refactor a nested if statement in `types.series.Series.inject()` to reduce its cognitive complexity. After this change, the PR passes codeclimate's test. I've also checked that the new method reproduces the behavior of the old method you approved in #720 .

Issue #729, opened 2018-04-06
TimeSeries.fft() and FrequencySeries.ifft() are inconsistent


alurban:
There seems to be a discrepancy between `TimeSeries.fft()` and `FrequencySeries.ifft()`. To take the example in the latter’s doctoring:

```
In [1]: from gwpy.timeseries import TimeSeries

In [2]: timeseries = TimeSeries([1.0, 0.0, -1.0, 0.0], sample_rate=1.0)

In [3]: timeseries.fft().ifft() == timeseries
Out[3]: False
```

It seems the units disagree and the first sample is off by a factor of 2:

```
In [4]: timeseries.unit
Out[4]: Unit(dimensionless)

In [5]: timeseries.fft().ifft().unit
Out[5]: Unit("Hz")

In [6]: timeseries.value
Out[6]: array([ 1.,  0., -1.,  0.])

In [7]: timeseries.fft().ifft().value
Out[7]: array([ 2.,  0., -1.,  0.])
```

PR #730, opened 2018-04-06
Make FrequencySeries.ifft() consistent with TimeSeries.fft()

alurban:
I've taken a stab at altering the units and normalization of `FrequencySeries.ifft()` so that it reproduces the original timeseries after `TimeSeries.fft().ifft()`. I've also added a unit test to make sure this is the case.

This fixes https://github.com/gwpy/gwpy/issues/729.

### DIFF
--- a/gwpy/frequencyseries/core.py
+++ b/gwpy/frequencyseries/core.py
@@ -230,8 +230,7 @@ class FrequencySeries(Series):
         # Undo normalization from TimeSeries.fft
         # The DC component does not have the factor of two applied
         # so we account for it here
-        dift = npfft.irfft(self.value * nout)
-        dift /= 2
+        dift = npfft.irfft(self.value * nout) / 2
         new = TimeSeries(dift, epoch=self.epoch, channel=self.channel,
                          unit=self.unit, dx=1/self.dx/nout)
         return new
--- a/gwpy/frequencyseries/core.py
+++ b/gwpy/frequencyseries/core.py
@@ -231,9 +231,9 @@ class FrequencySeries(Series):
         # The DC component does not have the factor of two applied
         # so we account for it here
         dift = npfft.irfft(self.value * nout)
-        dift[1:] /= 2
+        dift /= 2
         new = TimeSeries(dift, epoch=self.epoch, channel=self.channel,
-                         unit=self.unit * units.Hertz, dx=1/self.dx/nout)
+                         unit=self.unit, dx=1/self.dx/nout)
         return new
 
     def zpk(self, zeros, poles, gain, analog=True):
--- a/gwpy/tests/test_frequencyseries.py
+++ b/gwpy/tests/test_frequencyseries.py
@@ -37,6 +37,7 @@ use('agg')  # nopep8
 
 from astropy import units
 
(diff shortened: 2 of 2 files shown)


alurban:
Note, the issue pointed out by @pep8speaks is on a line not touched by this PR.

alurban:
@duncanmmacleod I think I see where at least one mixup happened. As written, `TimeSeries.fft()` normalizes by dividing out `nfft`, leaving the resulting `FrequencySeries` with the same units as the original `TimeSeries`. However, `FrequencySeries.ifft()` presumed that there would be an extra 1/Hz and attempted to multiply `self.unit * unit('Hz')` to get rid of this, but didn't change the overall scale.

As I see it, we could choose between two conventions:
1. Both `TimeSeries.fft()` and `FrequencySeries.ifft()` return the same units as the original object. This would mean the `FrequencySeries` represents a discrete Fourier transform.
2. `TimeSeries.fft()` scales by `self.duration.to('s') / nfft` (or `self.dt.to('s')`), giving units of `self.unit / unit('Hz')`, while `FrequencySeries.ifft()` scales by `self.df.to('Hz')` giving the original units. This would mean the `FrequencySeries` represents a discretized version of a continuous Fourier transform.

The first option requires the fewest changes and is what is currently implemented by this pull request, but I'm happy to code up the latter. Which option would you prefer?

alurban:
There is a build failure in Travis CI related to MacPorts:

```
--->  Updating MacPorts base sources using rsync
DEBUG: system: /usr/bin/rsync -rtzvl --delete-after rsync://rsync.macports.org/macports/release/tarballs/base.tar /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs
rsync: failed to connect to rsync.macports.org: No route to host (65)
rsync error: error in socket IO (code 10) at /BuildRoot/Library/Caches/com.apple.xbs/Sources/rsync/rsync-51/rsync/clientserver.c(106) [receiver=2.6.9]
Command failed: /usr/bin/rsync -rtzvl --delete-after rsync://rsync.macports.org/macports/release/tarballs/base.tar /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs
Exit code: 10
```

I don't think this has anything to do with proposed code changes, but I also don't seem to have any way to trigger a re-run of the build without either making a new commit or closing this PR and opening another one.

duncanmmacleod:
@alurban, I have restarted that build job.

alurban:
@duncanmmacleod, thank you. The build now completes without failure, looks like it just needed to be kicked.

duncanmmacleod:
@alurban, it sounds like we should have the `fft()` and `ifft()` methods be just what they advertise, discrete (I)FFTs, so in the original units. Can you work that up?

alurban:
@duncanmmacleod, I tend to agree, and this is actually what the PR already implements. That is, in the following example:
```
data = TimeSeries([1., 0., -1., 0.], sample_rate=1.0)
data_fd = data.fft()
data_td = data_fd.ifft()
```
All three objects (`data`, `data_fd`, `data_td`) have the same units and `data == data_td`.

PR #733, opened 2018-04-13
Set default epoch=None for GPSScale

duncanmmacleod:
This PR modifies the `GPSScale` to use a default `epoch=None`, which simplifies setting and unsetting the epoch dynamically when rendering the figure.

This also adds better unit testing of the auto-gps system.

The long term plan is to move this logic upstream into the `Axes`, since any `Axes` object can have an `'auto-gps'` scale.


duncanmmacleod:
This was superseded by #790.

PR #744, opened 2018-04-25
 Alright python3.x you need your file to be opened wb not just w.

scottcoughlin2014:
This is because i never made a  3x env and try the download. it works now and the file is opened write-binary not just write.


duncanmmacleod:
@scottcoughlin2014, this looks trivial, but can you squash the commits together so that it looks a bit cleaned in the history:

```
git rebase -i upstream develop
```

then 

```
git push --force
```

Issue #758, opened 2018-05-14
Cannot resize `xindex` when appending array if `xindex` getter has been called


stefco:
For some reason, calling the `Series.xindex` or, equivalently for timeseries, the `TimeSeries.times` getters causes problems when trying to append an array to another. I will look into this and try to fix it. If the getter is not called, appending works fine:

```python
In [159]: t = TimeSeries.read('foo.hdf5')
     ...: # start = t.times[-1] + t.dt
     ...: # del start
     ...: start = 1175011260.0
     ...: # end = t.times[-1] + t.dt + Quantity(100, "s")
     ...: # del end
     ...: end = 1175011360.0
     ...: t2 = c.placeholder(t, start, end).copy()
     ...: t3 = t.append(t2)

In [160]: t = TimeSeries.read('foo.hdf5')
     ...: # start = t.times[-1] + t.dt
     ...: # del start
     ...: t.times
     ...: start = 1175011260.0
     ...: # end = t.times[-1] + t.dt + Quantity(100, "s")
     ...: # del end
     ...: end = 1175011360.0
     ...: t2 = c.placeholder(t, start, end).copy()
     ...: t3 = t.append(t2)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/dev/gwpy/gwpy/types/series.py in append(self, other, gap, inplace, pad, resize)
    677                 try:
--> 678                     self.xindex.resize((s[0],), refcheck=False)
    679                 except ValueError as exc:

ValueError: cannot resize this array: it does not own its data

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-160-45fdee9ffe53> in <module>()
      8 end = 1175011360.0
      9 t2 = c.placeholder(t, start, end).copy()
---> 10 t3 = t.append(t2)

~/dev/gwpy/gwpy/types/series.py in append(self, other, gap, inplace, pad, resize)
    680                     if 'cannot resize' in str(exc):
    681                         self.xindex = self.xindex.copy()
--> 682                         self.xindex.resize((s[0],))
    683                     else:
    684                         raise

ValueError: cannot resize an array that references or is referenced
by another array in this way.  Use the resize function

In [161]: t = TimeSeries.read('foo.hdf5')
     ...: # start = t.times[-1] + t.dt
     ...: # del start
     ...: t.xindex
     ...: start = 1175011260.0
     ...: # end = t.times[-1] + t.dt + Quantity(100, "s")
[...]

PR #759, opened 2018-05-15
fix Quantity resize in Series.append

stefco:
described in issue #758, https://github.com/gwpy/gwpy/issues/758

When trying to append to a `Series` instance that already has _xindex initialized with an `astropy.units.Quantity` value, an error is thrown due to the fact that `Quantity` only owns a view to its own data and cannot therefore be resized.

Fix this by trying to copy the underlying numpy array in `quantity.value`, resizing that copy, and then creating a new `Quantity` instance with the resized array and the original `quantity.unit`.

### DIFF
--- a/gwpy/types/series.py
+++ b/gwpy/types/series.py
@@ -678,8 +678,18 @@ class Series(Array):
                     self.xindex.resize((s[0],), refcheck=False)
                 except ValueError as exc:
                     if 'cannot resize' in str(exc):
-                        self.xindex = self.xindex.copy()
-                        self.xindex.resize((s[0],))
+                        pass  # we'll try again with a copy
+                try:
+                    self.xindex = self.xindex.copy()
+                    self.xindex.resize((s[0],))
+                except ValueError as exc2:
+                    # astropy.units.Quantity cannot be resized because
+                    # it does not own its own data, which breaks t-axis
+                    # resizing.
+                    if 'cannot resize' in str(exc2):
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@stefco, I finally found the time to look into this, and I think you're misunderstanding the issue. The problem is reference counting, not data ownership - the first `try/except` clause catches that. I can fix the issue with the following patch:

```diff
diff --git a/gwpy/types/series.py b/gwpy/types/series.py
index b9a5f6ff..a190471f 100644
--- a/gwpy/types/series.py
+++ b/gwpy/types/series.py
@@ -685,8 +685,8 @@ class Series(Array):
                     self.xindex.resize((s[0],), refcheck=False)
                 except ValueError as exc:
                     if 'cannot resize' in str(exc):
-                        self.xindex = self.xindex.copy()
-                        self.xindex.resize((s[0],))
+                        self._xindex = self.xindex.copy()
+                        self._xindex.resize((s[0],))
                     else:
                         raise
             else:
```

Using the direct attribute reference `_xindex` rather than the property reference `xindex` gets directly to the new object (the copied array) without creating another reference to it in the property methods.

Can you try this out and if it works simplify your PR to implement it?

stefco:
Yeah, i’ll try that out! Thanks!

duncanmmacleod:
Bump @stefco

duncanmmacleod:
This hasn't seen attention in too long, I'm closing. Please reopen if the changes are still relevant.

PR #768, opened 2018-05-18
Use ligo-segments for segment objects

duncanmmacleod:
This PR updates gwpy to use `ligo-segments` as the provider for the `segment` objects required by `gwpy.segments.segments` (not `glue.segments`).

This means we don't need glue at all, so this dependency has been reduced to 'recommended'.

This in turn allows a py36 version of the port for macports, which has been added to the travis configuration.


duncanmmacleod:
`ligo-segments` has not been approved for production deployment for the `ligo/base` images yet, so most of the docker-based CI jobs are failing. This PR should not be merged until [this ticket](https://bugs.ligo.org/redmine/issues/6150) is closed and a port for macports has been approved.

PR #780, opened 2018-06-01
Added windows build config for appveyor

duncanmmacleod:
This PR adds a basic appveyor configuration to build gwpy on windows and run the test suite.


duncanmmacleod:
Windows support will require `ligo-segments-1.0.2`, and probably a fix for the pickling issue with `gwpy.utils.mp`.

duncanmmacleod:
@alurban, @areeda, this PR touches `TimeSeries.q_transform`, so if you could take a quick look, that would be great. Specifically [these changes](https://github.com/gwpy/gwpy/pull/780/files#diff-ece5c3d84855990064f4e6b30442a803), which are supposed to have zero functional impact, and were a result of some debugging that ended up having nothing to do with that method.

gwpy/timeseries/timeseries.py
```diff
@@ -1705,32 +1705,30 @@ def q_transform(self, qrange=(4, 64), frange=(0, numpy.inf),
...
-            span = self.span
+            data = self
+
+        # perform FFT to feed into Q-transform
+        fdata = data.fft().value
 
+        # metadata
+        epoch = data.x0
+        span = data.span
         if outseg is None:
```
alurban: Nice! I like these changes for the following reasons:
* The `fdata`, `epoch`, and `span` variables are set outside of an `if-else` statement
* `fftlength` and `overlap`, which are `asd_kw` arguments, are not redundantly set if `whiten` is a `FrequencySeries` object
* More lines are removed than are added

The build tests are all passing which makes me confident that you haven't accidentally dropped any needed functionality. Have you re-run some of the docs pages to check that the output is sane, particularly [this one](https://gwpy.github.io/docs/stable/examples/timeseries/qscan.html)?

PR #800, opened 2018-06-14
Update to fftgram method. This now outputs a complex spectrogram

scottcoughlin2014:
Mostly things like PEP8 or whatever is going on with rebasing will need updating


duncanmmacleod:
@scottcoughlin2014, [`scipy.signal.spectrogram`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.spectrogram.html) is only available in `scipy>=0.16.0`. Since the LDG is pinned at 0.12.0 we should think carefully about how to proceed.

gwpy/timeseries/timeseries.py
```diff
@@ -476,48 +476,57 @@ def spectrogram2(self, fftlength, overlap=0, **kwargs):
...
+        [frequencies, times, values] = spectrogram(self,
+                                                   fs=self.sample_rate.value,
+                                                   window=window,
+                                                   nperseg=nfft,
+                                                   noverlap=noverlap,
+                                                   mode='complex')
+
+        # The shape is incorrect for the Spectrogram object
+        for freq_idx in range(frequencies.size):
+            out[:, freq_idx] = values[freq_idx, :]
```
duncanmmacleod: @scottcoughlin2014, can we just do

```python
            out[:] = values.T
```

?

scottcoughlin2014:
@duncanmmacleod ya i think this is somewhat problematic but there has been some discussion of making the scipy version newer on the clusters maybe? I thought I remembered seeing some email but maybe not, its just that it a nice fast method that already has window and other flexibility cooked in.

gwpy/timeseries/timeseries.py
scottcoughlin2014: possibly for some reason I thought that was not right but one sec

duncanmmacleod:
@scottcoughlin2014, I don't really object to having this method require a particular version of scipy, there is precedent in other places, but we should protect against the `ImportError` that will present to those poor souls on scipy < 0.16 and tell them what the real problem is; can you take care of that?

scottcoughlin2014:
@duncanmmacleod Ill do this today

scottcoughlin2014:
@duncanmmacleod looks like I need to as a unit test make sure the coverage does not decrease, however I do not think that the build failure has to do with this

duncanmmacleod:
@scottcoughlin2014 I have restarted the failed CI jobs, and yes, please add a unit test.

scottcoughlin2014:
@duncanmmacleod can I have a check on this? the code climate issues seem to be function is too long by 1 line and that I have too many argument both seem like not a big deal. I have written unit tests as you can see I increased percentage by a whopping 0.2 percent

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -650,6 +650,27 @@ def test_spectrogram2(self, losc):
...
 
+    @utils.skip_minimum_version('scipy', '0.16')
+    def test_fftgram(self, losc):
+        fgram = losc.fftgram(1)
+        fs = int(losc.sample_rate.value)
+        [f, t, sxx] = signal.spectrogram(losc, fs,
+                                         window='hann',
+                                         nperseg=fs,
+                                         mode='complex')
+        numpy.testing.assert_array_equal(losc.t0.value + t, fgram.xindex.value)
```
duncanmmacleod: `assert_array_equal` can be access as `utils.assert_array_equal`, which may make things a little cleaner.

gwpy/timeseries/timeseries.py
```diff
@@ -476,48 +476,61 @@ def spectrogram2(self, fftlength, overlap=0, **kwargs):
...
-            # find step TimeSeries
-            idx = stride * step
-            idx_end = idx + stride
-            stepseries = self[idx:idx_end]
-            # calculated FFT and stack
-            stepfft = stepseries.fft()
-            out[step] = stepfft.value
-            if step == 0:
-                out.frequencies = stepfft.frequencies
+        [frequencies, times, sxx] = spectrogram(self,
```
duncanmmacleod: the use of square brackets here looks very matlab, can you please remove them (they are not required)?

gwpy/timeseries/timeseries.py
```diff
@@ -476,48 +476,61 @@ def spectrogram2(self, fftlength, overlap=0, **kwargs):
...
+                                                nperseg=nfft,
+                                                noverlap=noverlap,
+                                                mode='complex',
+                                                **kwargs)
+        out = Spectrogram(sxx.T,
+                          name=self.name, unit=self.unit,
+                          xindex=self.t0.value + times,
+                          yindex=frequencies)
+
         return out
```
duncanmmacleod: if you are just assigning then returning, I would assign and return in one step:

```python
        return Spectrogram(sxx.T, ...)
```

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -650,6 +650,27 @@ def test_spectrogram2(self, losc):
         # note: bizarre stride length because 4096/100 gets rounded
         assert sg.dt == 0.010009765625 * units.second
 
+    @utils.skip_minimum_version('scipy', '0.16')
+    def test_fftgram(self, losc):
+        fgram = losc.fftgram(1)
+        fs = int(losc.sample_rate.value)
+        [f, t, sxx] = signal.spectrogram(losc, fs,
```
duncanmmacleod: please remove the square brackets.

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -650,6 +650,27 @@ def test_spectrogram2(self, losc):
         # note: bizarre stride length because 4096/100 gets rounded
         assert sg.dt == 0.010009765625 * units.second
 
+    @utils.skip_minimum_version('scipy', '0.16')
+    def test_fftgram(self, losc):
+        fgram = losc.fftgram(1)
+        fs = int(losc.sample_rate.value)
```
duncanmmacleod: is this call to `int` required?

gwpy/timeseries/timeseries.py
scottcoughlin2014: its not my fault I started with MATLAB as a language :)

scottcoughlin2014:
@duncanmmacleod could i get a restart of the CI? not sure what happened

scottcoughlin2014:
@duncanmmacleod So I think this thing is ready seems like all checks except codebeat are happy

duncanmmacleod:
@scottcoughlin2014, is this ready to go?

scottcoughlin2014:
@duncanmmacleod I have been using it and the tests and what not all seem good to go to me.

duncanmmacleod:
Thanks @scottcoughlin2014.

Issue #818, opened 2018-06-23
gwpy-plot: Backend setting prevents headless ops


areeda:
Error is:
```/usr/local/ldvw/bin/gwpy-plot:92: UserWarning:
This call to matplotlib.use() has no effect because the backend has already
been chosen; matplotlib.use() must be called *before* pylab, matplotlib.pyplot,
or matplotlib.backends is imported for the first time.

The backend was *originally* set to u'TkAgg' by the following code:
  File "/usr/local/ldvw/bin/gwpy-plot", line 32, in <module>
    from gwpy import __version__
```
Fix on the way.

PR #819, opened 2018-06-23
Agg problem Issue #818

areeda:
It looks like there was a recent update to `gwpy.__version__` that imports at least 1 of `matplotlib.pyplot`, or `matplotlib.backends` which prevents us from setting the backend to something that works in a headless environment.

This PR moves our matplotlib.use() to as early as possible in gwpy-plot

### DIFF
--- a/bin/gwpy-plot
+++ b/bin/gwpy-plot
@@ -24,10 +24,16 @@ import time
 
 import os
 import sys
-from importlib import import_module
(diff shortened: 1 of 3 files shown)


duncanmmacleod:
@areeda, I thought that the pyplot issue was resolved by 07823845371ed55f1108481c13ed2b105cdefb4f, can you rebase your testing code against the upstream `develop` branch and reproduce the issue?

areeda:
This PR is being superseded.

PR #822, opened 2018-06-24
Fixed some broken tests exposed during conda build

duncanmmacleod:
This PR fixes two bugs in testing exposed only when testing under conda, where glue is available, but lalframe and libframe are not, and where the tests are run from the installed package, not from the git repo.


duncanmmacleod:
@scottcoughlin2014, this patch modifies the tests for the `GravitySpyTable` object, can you please review?

scottcoughlin2014:
@duncanmmacleod The file appears to have been properly transferred from the `json` file to the test module itself. since we are not testing the `json` module ability to read json but whether the contents of the file are the same as the query this patch seems appropriate and i approve.

duncanmmacleod:
thanks @scottcoughlin2014

Issue #821, opened 2018-06-24
gwpy-plot spectrogram fails if zero ffts and missing --nopct option


areeda:
I may be the only one who uses it but the option is there so that the color bar limits can be set to a percentile value of the actual spectrogram.

 I don't understand why the histogram based background estimation works so well in the Java version and so poorly here.

duncanmmacleod:
Migrated to https://gitlab.com/gwpy/gwpy/-/issues/821, please join us on gitlab to continue your contribution!

PR #825, opened 2018-06-25
Spectrogram log zero Issue #821, 820 & 818

areeda:
Note this PR includes changes in PRs 817 & 819
It addresses signals with data but have long periods of zeroes so that one of the columns in the spectrogram is all zeroes


gwpy/cli/spectrogram.py
```diff
@@ -166,15 +166,21 @@ def scale_axes_from_data(self):
                 args.ymin, args.ymax)
 
         # auto scale colours
-        from numpy import percentile
+        import numpy as np
```
duncanmmacleod: @areeda, is this change necessary?

duncanmmacleod:
@areeda, this PR seems to have the combined commits from #817 and #819 as well as anything new for this PR. Can you rebase/clean up the branch to just have the necessary commits for this patch?

areeda:
I believe I have addressed all of the issues we talked about this morning.
 I'm not sure what to do about that codeclimate issue it seems  not like that function although I don't see it as more complex than many of the others.

duncanmmacleod:
@areeda, this PR now includes more commits than it did before, but admittedly most of those are already merged, so they don't impact things too much. However, this does make it hard to review this PR for what it is.

Specifically, this PR still includes 815dbeaa77699aab3a85da2adf8ed8f316c1329d, which I maintain we don't need, so if that can be excised so that this PR can continue forward, that would be good.

areeda:
This PR is being superseded.

PR #835, opened 2018-07-05
Cumulative changes from gwpy-plot testing

areeda:
Sorry for the "merge" commits rebase failed before each one.
Annotation for changed files:

**bin/gwpy-plot**
* move test for no DISPLAY to top for matplotlib.use()

**gwpy/cli/cliproduct.py**
* :rtype annotation added to log() docs

**gwpy/cli/qtransform.py**
* `def fformat(x):  # float format` failed if passed a str arg

**gwpy/cli/spectrogram.py**
* Deal with zeroes in resulting spectrogram and log color scale


gwpy/cli/qtransform.py
```diff
@@ -148,6 +149,9 @@ def fformat(x):  # float format
                 return '[{0}]'.format(', '.join(map(fformat, x)))
             if isinstance(x, Quantity):
                 x = x.value
+            elif isinstance(x, str):
```
duncanmmacleod: how does `fformat` ever get called with a `str`? if its via `qrange` I think a better switch would be to set `type=float` in the `--qrange` option definition in line 101.

areeda: This is one of your changes. I needed this to go further.
I'll look into it.

duncanmmacleod: thanks @areeda, I should have been a bit clearer, reading back my message it sounds rude, sorry. I'm just surprised that this hasn't been caught before, can you post a command-line that reproduces it?

duncanmmacleod:
@areeda, is there anything else to go into this PR?

PR #861, opened 2018-08-03
Minor error handling updates

areeda:
Automated system testing showed a few errors and unhelpful error messages. This pull request makes minor changes to a bunch of files that hopefully make the errors more informative.

### DIFF
--- a/bin/gwpy-plot
+++ b/bin/gwpy-plot
@@ -118,4 +118,4 @@ if INTERACTIVE:
 run_time = time.time() - PROG_START
 prod.log(1, 'Program run time: %.1f' % run_time)
 if prod.got_error:
-    sys.exit(2)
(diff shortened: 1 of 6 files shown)


areeda:
I'm not sure I understand the issues brought up by codeclimate and codebeat.
Are these holding up simple request?
Do we comply with suggestions like function longer than 25 lines or no nested if-else?

duncanmmacleod:
@areeda, codeclimate and codebeat are mainly complaining about complexity, since you have introduced a bunch more `if` statements, which increase the nesting level of some blocks. If you can have a quick look at the code and work out if any of the `if/elif/else` statements can be combined to reduce the nesting, or to remove clauses, that would be great.

But it's not a blocker, so I'm happy to merge whenever you like.

PR #865, opened 2018-08-07
Updated gwpy.io.datafind to use new gwdatafind package

duncanmmacleod:
This PR updates `gwpy.io.datafind` (and `gwpy.timeseries.core`) to use the new [`gwdatafind`](//pypi.org/project/gwdatafind) package instead of `glue.datafind`. I also included some improvements to the tests for `gwpy.io.datafind` and a patch to `gwpy.io.gwf` to accommodate reading files from `unicode` on python2.

`gwdatafind` is not yet available under macports, or in the LIGO `apt` or `yum` repositories, so this PR should probably wait until those uploads are complete and the package can be added to the relevant optional dependencies lists in CI.


duncanmmacleod:
`gwdatafind` is now available under Macports, see macports/macports-ports#2349.

Issue #451, opened 2017-05-05
Applying FIR filters to long timeseries


andrew-lundgren:
Scipy seems to lack a method for applying FIR filters to very long time series (meaning longer than can be efficiently FFTed). One standard way to do this is the [overlap-save method](https://en.wikipedia.org/wiki/Overlap-save_method). It goes through the data by chunks, overlapping enough to avoid the corruption from filter wraparound.

I've implemented this using standard numpy and scipy calls. The sample code is below. I think this should be made a method of the TimeSeries class, maybe called `.convolve`, since it's equivalent to `scipy.signal.convolve`.

If this is OK, I can prepare a patch to `TimeSeries` in a more gwpy style. This could be used to close #358.

```python
def overlap_save(in1, in2, fftlen=16384):
    """Convolve a very long data series with a short one
        using the overlap-save method. Result should be
        nearly identical to scipy.signal.convolve with
        mode='same'.
        
        Inputs:
        in1 (numpy array): long data series
        in2 (numpy array): short data series (e.g. FIR filter)
        fftlen (int, optional): number of points to use for FFT
        
        Output:
        Numpy array with same length as in1"""
    # If data is short enough, just do a single convolve
    if len(in1) <= 2*fftlen:
        return sig.convolve(in1, in2, mode='same')
    
    # Find how much is corrupted by filter wraparound
    pad = int(np.ceil(len(in2)/2.))
    step = int(fftlen) - 2*pad
    # Create room for the result
    result=np.zeros(len(in1), dtype=in1.dtype)
    # Do the first chunk on its own
    result[:fftlen-pad]=sig.fftconvolve(in1[:fftlen-pad], in2, mode='same')
    # Walk through chunk by chunk
    for idx in xrange(2*step, len(in1), step):
        tmp = sig.fftconvolve(in1[idx-fftlen:idx], in2, mode='same')
        result[idx-fftlen+pad:idx-pad] = tmp[pad:-pad]
    # Clean up the last piece which may be anything smaller than step
    # It's important that we made sure to have at least one step in
    # the loop above, otherwise idx is undefined
    len_final = len(in1) - idx
    tmp = sig.fftconvolve(in1[-fftlen:], in2, mode='same')
    result[-len_final-pad:] = tmp[-len_final-pad:]
    
    return result
```

alurban:
[...]

Issue #358, opened 2017-01-31
TimeSeries.whiten does not apply FIR filter correctly


andrew-lundgren:
The gwpy.TimeSeries.whiten method uses an incorrect method for applying FIR filters. This causes discontinuities if the whitening filter has narrow features that cause a wide impulse response, even when overlap is used. I've uploaded a Jupyter notebook illustrating the problem:

[Notebook](https://github.com/andrew-lundgren/gwpy/blob/whiten_bug/examples/test_gwpy_whiten.ipynb)

duncanmmacleod:
@andrew-lundgren, I (finally) took a look at your notebook. Given the shape of the Hann window, should we not expect problems if you select `overlap=0`? If you select `overlap=8` (50%), I get a more sensible result.

PR #868, opened 2018-08-13
New whitening scheme based on inverse spectrum truncation

### LINKED ISSUE #867: TimeSeries.whiten() can easily introduce filter transients
While putting together the [`gwdetchar-omega`](https://github.com/gwdetchar/gwdetchar) tool, I discovered that in many cases the default whitening scheme in GWPy can easily introduce filter transients in many channels. An example is shown below:

170817_old_whitening

These data are from the Livingston calibration channel `L1:CAL-PCALY_RX_PD_OUT_DQ` near the time of binary neutron star merger GW170817. This channel shows reflected laser light in the photon calibrator at the Y-end station, which we use to independently monitor the calibration lines. To make this plot I used the default whitening with `fftlength=2`, `overlap=1`. Clearly, there are filter transients that recur once per overlap (i.e. once per second) near 330 and 1087 Hz, which are the locations of two of the calibration lines.

As @andrew-lundgren pointed out in #358, this behavior is not ubiquitous but it is relatively common. I believe it can be remedied by allowing users to select an alternative whitening scheme that designs and uses an FIR filter, similar to the one in [PyCBC](http://pycbc.org/pycbc/latest/html/_modules/pycbc/types/timeseries.html#TimeSeries.whiten), which is based on [FINDCHIRP](https://arxiv.org/abs/gr-qc/0509116).

I have coded this up and will post a merge request momentarily.

alurban:
This pull request addresses a common issue with filter transients in the current whitening method by replacing it with an alternative scheme which designs and applies an FIR (finite impulse response) filter using inverse spectrum truncation.

Other new features added include:
* A method `FrequencySeries.interpolate()` and functions `timeseries._fft_length_default`, `signal.filter_design.truncate_transfer`, and `signal.filter_design.truncate_impulse`, all used to design the FIR filter
* New keyword arguments for filter duration (in the time domain) and highpass corner frequency
* In `TimeSeries.whiten()`, a default value of `fftlength`which  causes it to be set to `max(2, int(2048 // self.sample_rate.decompose().value))`
* Automatic cropping of the beginning and end of the whitened timeseries by a segment of length `filter_duration`
* Updates to the documentation for `TimeSeries.whiten()` addressing these changes
* Updates to `q_transform()` that address these recent changes, including a new default `fftlength` that mirrors the one for `TimeSeries.whiten`
* Unit tests for new functionality, a unit test for `signal.window.planck()`, and a bug fix in the unit test for `FrequencySeries.inject()`

Note, these changes make it so that a simple call to `TimeSeries.whiten()` with no other arguments returns something sensible in most cases. It is also backwards compatible.

This fixes #867. It is also related to #358 and #451.

### DIFF
--- a/gwpy/frequencyseries/frequencyseries.py
+++ b/gwpy/frequencyseries/frequencyseries.py
@@ -22,6 +22,7 @@
 import warnings
 from copy import deepcopy
 
+import numpy
 from numpy import fft as npfft
 from scipy import signal
 
@@ -268,6 +269,39 @@ class FrequencySeries(Series):
             >>> data2 = data.zpk([100]*5, [1]*5, 1e-10)
         """
         return self.filter(zeros, poles, gain, analog=analog)
+
+    def interpolate(self, df, left=0, right=0):
+        """Interpolate this `FrequencySeries` to a new resolution.
+
+        Parameters
+        ----------
+        df : `float`
+            desired frequency resolution of the interpolated `FrequencySeries`,
+            in Hz
+
+        left : `float`
+            value to return for f < flow, default: 0
+
+        right : `float`
+            value to return for f > fhigh, default: 0
+
+        Returns
+        -------
+        out : `FrequencySeries`
+            the interpolated version of the input `FrequencySeries`
+
+        See Also
+        --------
+        :func:`numpy.interp`
+            for the underlying 1-D linear interpolation scheme
+        """
+        N = (self.size - 1) * (self.df.decompose().value / df) + 1
+        fsamples = numpy.arange(0, numpy.rint(N)) * df
+        out = numpy.interp(fsamples, self.frequencies.value, self.value,
+                           left=left, right=right).view(type(self))
+        out.__array_finalize__(self)
+        out.df = df
+        return out
 
     def filter(self, *filt, **kwargs):
         """Apply a filter to this `FrequencySeries`.
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -52,6 +52,22 @@ def _update_doc_with_fft_methods(func):
     return func
 
 
+def _fft_length_default(dt):
+    """Choose an appropriate FFT length (in seconds) based on a sample rate
+
+    Parameters
+    ----------
+    dt : `~astropy.units.Quantity`
+        the sampling time interval, in seconds
+
+    Returns
+    -------
+    fftlength : `int`
+        a choice of FFT length, in seconds
+    """
+    return int(max(2, numpy.ceil(2048 * dt.decompose().value)))
+
+
 # -- TimeSeries ---------------------------------------------------------------
 
 class TimeSeries(TimeSeriesBase):
@@ -1428,14 +1444,16 @@ class TimeSeries(TimeSeriesBase):
         out *= planck(out.size, nleft=nleft, nright=nright)
         return out
 
-    def whiten(self, fftlength, overlap=0, method='scipy-welch',
-               window='hanning', detrend='constant', asd=None, **kwargs):
-        """White this `TimeSeries` against its own ASD
+    def whiten(self, fftlength=None, overlap=0, method='scipy-welch',
+               window='hanning', detrend='constant', asd=None,
+               fduration=1, highpass=None, **kwargs):
+        """Whiten this `TimeSeries` using inverse spectrum truncation
 
         Parameters
         ----------
-        fftlength : `float`
-            number of seconds in single FFT
+        fftlength : `float`, optional
+            FFT integration length (in seconds) for ASD estimation,
+            default: choose based on sample rate
 
         overlap : `float`, optional
             number of seconds of overlap between FFTs, defaults to the
@@ -1447,15 +1465,25 @@ class TimeSeries(TimeSeriesBase):
 
         window : `str`, `numpy.ndarray`, optional
             window function to apply to timeseries prior to FFT,
+            default: ``'hanning'``
             see :func:`scipy.signal.get_window` for details on acceptable
             formats
 
         detrend : `str`, optional
             type of detrending to do before FFT (see `~TimeSeries.detrend`
-            for more details)
+            for more details), default: ``'constant'``
 
-        asd : `~gwpy.frequencyseries.FrequencySeries`
-            the amplitude-spectral density using which to whiten the data
+        asd : `~gwpy.frequencyseries.FrequencySeries`, optional
+            the amplitude spectral density using which to whiten the data,
+            overrides other ASD arguments, default: `None`
+
+        fduration : `float`, optional
+            duration (in seconds) of the time-domain FIR whitening filter,
+            defaults to fftlength
+
+        highpass : `float`, optional
+            highpass corner frequency (in Hz) of the FIR whitening filter,
+            default: `None`
 
         **kwargs
             other keyword arguments are passed to the `TimeSeries.asd`
@@ -1476,36 +1504,35 @@ class TimeSeries(TimeSeriesBase):
         scipy.signal
 
         Notes
-        -----"""
+        -----
+        The `window` argument is used in both ASD estimation and FIR filter
+        design.
+
+        Due to filter corruption at the boundaries, a segment of length
+        `fduration` will automatically be cropped from the beginning and
+        end of the output.
+
+        For more on inverse spectrum truncation, see arXiv:gr-qc/0509116.
+        """
         # build whitener
+        fftlength = fftlength if fftlength else _fft_length_default(self.dt)
         if asd is None:
             asd = self.asd(fftlength, overlap=overlap,
                            method=method, window=window, **kwargs)
-        if isinstance(asd, units.Quantity):
-            asd = asd.value
-        invasd = 1. / asd
-        # build window
-        nfft = int((fftlength * self.sample_rate).decompose().value)
-        noverlap = int((overlap * self.sample_rate).decompose().value)
-        # format window
-        if type(window).__module__ == 'lal.lal':
-            window = window.data.data
-        elif not isinstance(window, numpy.ndarray):
-            window = signal.get_window(window, nfft)
-        # create output series
-        nstride = nfft - noverlap
-        nsteps = 1 + int((self.size - nfft) / nstride)
-        out = numpy.zeros(nsteps * nstride + noverlap).view(type(self))
-        out.__metadata_finalize__(self)
-        out._unit = self.unit
-        del out.times
-        # loop over ffts and whiten each one
-        for i in range(nsteps):
-            x = i * nstride
-            y = x + nfft
-            in_ = self[x:y].detrend(detrend) * window
-            out.value[x:y] += npfft.irfft(in_.fft().value * invasd)
-        return out
+        asd = asd.interpolate(1./self.duration.decompose().value)
+        # truncate the edges and apply a hard highpass
+        ncorner = int(highpass / asd.df.decompose().value) if highpass else 0
+        invasd = filter_design.truncate_transfer(1./asd.value, ncorner=ncorner)
+        # truncate and window in the time domain
+        tdw = npfft.irfft(invasd)
+        ntaps = int((fduration * self.sample_rate).decompose().value)
+        tdw = filter_design.truncate_impulse(tdw, ntaps=ntaps, window=window)
+        # filter in the frequency domain, enforcing zero-phase
+        invasd = (self.duration.value/2) * numpy.abs(npfft.rfft(tdw))
+        in_ = self.detrend(detrend)
+        out = type(self)(npfft.irfft(in_.fft().value * invasd))
+        out.__array_finalize__(self)
+        return out.crop(*out.xspan.contract(fduration))
 
     def detrend(self, detrend='constant'):
         """Remove the trend from this `TimeSeries`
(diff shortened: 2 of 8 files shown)


alurban:
@duncanmmacleod, I've compared the whitening schemes by plotting the timeseries in figure 1 (original scheme in orange, new scheme in blue). In figure 2 the difference between the two is shown in blue, while the bandpassed difference is shown in orange to demonstrate that the filter artifacts are not produced by the new scheme. (You can see some 1 Hz beat notes / windowing artifacts in the orange timeseries.)

170817_compare_whitening
170817_whitening_difference


On a separate note, it looks like there are a few issues with this PR. I'll have a look at those in short order. One issue is python 3 readiness, which I think i can resolve with modest effort. The other issues are from `codeclimate` and related to complexity. There's not much I can do about this as the arguments I've added are all necessary, except perhaps the `highpass` argument, and the large complexity and number of lines come from supporting two separate whitening schemes. I didn't want to delete your existing scheme, presumably you want to keep it in there? Do you have any thoughts?

For my purposes, this alternative whitening scheme would be very nice to have in 0.12.0 as I want to use it for `gwdetchar-omega`.

gwpy/timeseries/timeseries.py
```diff
@@ -1428,14 +1428,15 @@ def taper(self, side='leftright'):
         out *= planck(out.size, nleft=nleft, nright=nright)
         return out
 
-    def whiten(self, fftlength, overlap=0, method='scipy-welch',
-               window='hanning', detrend='constant', asd=None, **kwargs):
-        """White this `TimeSeries` against its own ASD
+    def whiten(self, fftlength=2, overlap=0, method='scipy-welch',
+               window='hanning', detrend='constant', asd=None,
+               fir=False, filter_duration=1, highpass=0, **kwargs):
```
duncanmmacleod: I think I would be more comfortable with `highpass=None` as the default, to make it more obvious that the default is to not apply a FD highpass.

duncanmmacleod: is it reasonable to tie the `filter_duration` to `fftlength` and remove the former option? they seem redundant to me, but I understand they are currently used in separate contexts.

gwpy/timeseries/timeseries.py
```diff
@@ -1428,14 +1428,15 @@ def taper(self, side='leftright'):
         out *= planck(out.size, nleft=nleft, nright=nright)
         return out
 
-    def whiten(self, fftlength, overlap=0, method='scipy-welch',
-               window='hanning', detrend='constant', asd=None, **kwargs):
-        """White this `TimeSeries` against its own ASD
+    def whiten(self, fftlength=2, overlap=0, method='scipy-welch',
```
duncanmmacleod: why was a default added here? and why this value?

gwpy/timeseries/timeseries.py
alurban: Makes sense to me, and that's a pretty easy change, so I'll make that the default.

gwpy/frequencyseries/frequencyseries.py
```diff
@@ -269,6 +270,37 @@ def zpk(self, zeros, poles, gain, analog=True):
...
         return self.filter(zeros, poles, gain, analog=analog)
 
+    def interpolate(self, df, left=0, right=0):
+        """Interpolate this `FrequencySeries` to a new resolution.
+
+        Parameters
+        ----------
+        df : `float`
+            desired frequency resolution of the interpolated `FrequencySeries`,
+            in Hz
```
duncanmmacleod: can you leave a blank line between parameter definitions in the docstring, it makes things a little easier to read IMO

gwpy/timeseries/timeseries.py
alurban: I prefer to keep `fftlength` and `filter_duration` as separate options. In this new scheme, the former is a feature of the ASD calculation, while the latter is a feature of the actual filter that doesn't necessarily relate to `fftlength` at all. This is one reason why I've introduced a `FrequencySeries.interpolate` method, which is very similar to what [PyCBC](http://pycbc.org/pycbc/latest/html/_modules/pycbc/psd/estimate.html#interpolate) does.

gwpy/frequencyseries/frequencyseries.py
alurban: Sure, I'll make this change as well

gwpy/timeseries/timeseries.py
```diff
@@ -1476,11 +1488,48 @@ def whiten(self, fftlength, overlap=0, method='scipy-welch',
...
+            window = signal.get_window(window, filter_length)
+            tdw[0:trunc_start] *= window[int(filter_length/2):filter_length]
+            tdw[trunc_stop:self.size] *= window[0:int(filter_length/2)]
+            tdw[trunc_start:trunc_stop] = 0
+            # apply truncated filter in the frequency domain
+            invasd = numpy.abs(npfft.rfft(tdw))  # enforce zero-phase
+            in_ = self.detrend(detrend)
+            out = numpy.zeros(self.size).view(type(self))
+            out.__metadata_finalize__(self)
+            out._unit = self.unit
```
duncanmmacleod: you should be able to use `out.__array_finalize__` to set the units and other metadata in one call, I haven't updated other methods to do this yet.

gwpy/timeseries/timeseries.py
```diff
@@ -1476,11 +1488,48 @@ def whiten(self, fftlength, overlap=0, method='scipy-welch',
...
+            out.value[:] += npfft.irfft(in_.fft().value * invasd)
+            out.value[:] *= self.duration.value / 2  # re-normalise
```
duncanmmacleod: please combine these into a single operation (unless there's a good reason not to)

gwpy/timeseries/timeseries.py
alurban: I added the default because under the new scheme, `fftlength` isn't used at all if `fir=True` and `asd` is not `None`. I chose a default of 2 seconds because that's the smallest use case I see for this parameter, and I see it quite often.

gwpy/timeseries/timeseries.py
```diff
@@ -1476,11 +1488,48 @@ def whiten(self, fftlength, overlap=0, method='scipy-welch',
...
+            trunc_start = int(filter_length / 2)
+            trunc_stop = self.size - int(filter_length / 2)
+            window = signal.get_window(window, filter_length)
+            tdw[0:trunc_start] *= window[int(filter_length/2):filter_length]
+            tdw[trunc_stop:self.size] *= window[0:int(filter_length/2)]
+            tdw[trunc_start:trunc_stop] = 0
+            # apply truncated filter in the frequency domain
+            invasd = numpy.abs(npfft.rfft(tdw))  # enforce zero-phase
+            in_ = self.detrend(detrend)
+            out = numpy.zeros(self.size).view(type(self))
```
duncanmmacleod: it seems cleaner to just do something like

```python
out = type(self)(npfft.irfft(in_.fft().value * invasd) * self.duration.value/2)
out.__array_finalize__(self)
```

gwpy/timeseries/timeseries.py
```diff
@@ -1476,11 +1488,48 @@ def whiten(self, fftlength, overlap=0, method='scipy-welch',
...
+            out.__metadata_finalize__(self)
+            out._unit = self.unit
+            out.value[:] += npfft.irfft(in_.fft().value * invasd)
+            out.value[:] *= self.duration.value / 2  # re-normalise
+            del tdw, invasd, in_, out.times
```
duncanmmacleod: why are these `del` required?

alurban: I think I was emulating what's done in another method, can't remember which one. Are those arrays deleted from memory automatically after the function returns?

duncanmmacleod: As soon as there are no references to them, the garbage collector will remove them automatically, so as long as they aren't linked to from the returned `out` array, this should not be necessary.

gwpy/timeseries/timeseries.py
duncanmmacleod: Ok, sounds reasonable.

alurban:
@duncanmmacleod, I've addressed your review comments and removed the old whitening scheme entirely. The new whitening method has the same name (`TimeSeries.whiten()`) and same keyword arguments, with two new ones for filter duration and highpass. This should satisfy the requirement of being backward compatible.

However, now there are two build errors in `test_q_transform`. I'll have a look at that shortly.

alurban:
@duncanmmacleod, I've resolved WIP status and am ready for review. I fixed the errors by automatically cropping filter corruptions out of the whitened timeseries, then adjusting the `q_transform` and various unit tests to account for this.

alurban:
@duncanmmacleod, for illustration, here are the whitened timeseries (figure 1) and a Q-spectrogram (figure 2) of `L1:GDS-CALIB_STRAIN` at the end of GW170817. I made these using the most recent version of the new whitening method with `fftlength=8`, `overlap=4`, `qrange=(93.1, 93.1)`, and all other options default.

170817_whitened_timeseries
170817_strain

I'm quite happy with this, even with a 1-second filter the glitch is well-resolved and doesn't really ring off or corrupt the filter much at all. You can even see where the glitch isn't quite loud enough to surpass a few lines in the spectrum, and the BNS trace is clearly visible as well.

After I have some ice cream, I'll try squashing these commits.

gwpy/timeseries/timeseries.py
```diff
@@ -1428,14 +1428,17 @@ def taper(self, side='leftright'):
...
         return out
 
-    def whiten(self, fftlength, overlap=0, method='scipy-welch',
-               window='hanning', detrend='constant', asd=None, **kwargs):
-        """White this `TimeSeries` against its own ASD
+    def whiten(self, fftlength=2, overlap=0, method='scipy-welch',
+               window='hanning', detrend='constant', asd=None,
+               fduration=1, highpass=None, **kwargs):
+        """Whiten this `TimeSeries` against its own ASD using inverse spectrum
+           truncation
```
duncanmmacleod: Can you shorten this docstring to

```
Whiten this `TimeSeries` using inverse spectrum truncation
```

? This will make it fit on one line, and the whitening doesn't have to use 'its own ASD'.

gwpy/timeseries/timeseries.py
```diff
@@ -1476,36 +1489,43 @@ def whiten(self, fftlength, overlap=0, method='scipy-welch',
...
+        tdw[0:trunc_start] *= window[trunc_start:filter_length]
+        tdw[trunc_stop:self.size] *= window[0:trunc_start]
+        tdw[trunc_start:trunc_stop] = 0
+        # apply truncated zero-phase filter in the frequency domain
+        invasd = (self.duration.value/2) * numpy.abs(npfft.rfft(tdw))
+        in_ = self.detrend(detrend)
+        out = type(self)(npfft.irfft(in_.fft().value * invasd))
+        out.__array_finalize__(self)
+        return out.crop(out.t0.decompose().value + fduration,
+                        (out.t0 + out.duration).decompose().value - fduration)
```
duncanmmacleod: it would be cleaner here to use `out.xspan`, which does all decompositions internally, something like

```python
return out.crop(*out.xspan.contract(fduration))
```

gwpy/timeseries/timeseries.py
alurban: This is fixed in commit 66e9982, along with a pesky split infinitive elsewhere in that doctoring.

gwpy/timeseries/timeseries.py
alurban: I like that a lot better as well! This is fixed in commit 66e9982.

alurban:
@duncanmmacleod, I think we're ready for primetime. I've refactored `TimeSeries.whiten()` such that the filter truncation is done with two new functions, `signal.filter_design.truncate_transfer` and `signal.filter_design.truncate_impulse` which smoothly truncate (and, optionally, highpass) the transfer function and impulse response of the desired filter. I've also added a global function, `timeseries._fft_length_default`, which sets the default FFT length and is invoked by `TimeSeries.whiten()` and `TimeSeries.q_transform()`. Unit tests for the new filter design functions are also included.

On a separate note, I managed to reproduce the original Q-spectrogram for GW150914 by using a filter duration of 2 seconds, the same as the FFT length:

150914_strain

I still think the default filter length of 1 second is a sensible default though, and I did use it to reproduce the Q-spectrogram for GW170817 above.

gwpy/frequencyseries/frequencyseries.py
```diff
@@ -269,6 +270,38 @@ def zpk(self, zeros, poles, gain, analog=True):
...
+            value to return for f > fhigh, default: 0
+
+        Returns
+        -------
+        out : `FrequencySeries`
+            the interpolated version of the input `FrequencySeries`
+
+        See Also
+        --------
+        :func:`numpy.interp` for the underlying 1-D linear interpolation scheme
```
duncanmmacleod: This usage of `see also` doesn't work, numpydoc expects the format to be:

```
See Also
--------
:func:`numpy.interp`
    for the undelying 1-D linear interpolation scheme
```
I think. The new [numpydoc docs](https://numpydoc.readthedocs.io/en/latest/format.html#sections) suggest that you might be able to use a simple colon-separated key, value listing, but I haven't tested that.

alurban: Good catch! This is addressed in commit 58c95a9.

duncanmmacleod:
Thanks @alurban!

Issue #881, opened 2018-08-20
gwpy-plot no longer accepts relative epoch


areeda:
--epoch is treated as an absolute time.  We used to be able to set it relative to start time.

duncanmmacleod:
@areeda, please confirm that this was fixed by #882.

duncanmmacleod:
Migrated to https://gitlab.com/gwpy/gwpy/-/issues/881, please join us on gitlab to continue your contribution!

PR #882, opened 2018-08-20
Relative epoch

areeda:
Addresses issue #881 if epoch is small it is relative to min start time

### DIFF
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -32,7 +32,6 @@ import sys
 from collections import OrderedDict
 from functools import wraps
 
-from ..time import LIGOTimeGPS
 from six import add_metaclass
 
 from matplotlib import rcParams
@@ -893,7 +892,7 @@ class TimeDomainProduct(CliProduct):
         else:
             epoch = args.epoch
             if (epoch < 1e8):
-                args.epoch = LIGOTimeGPS(epoch + min(starts))
+                args.epoch = epoch + min(starts)
 
         if args.xmax is None:
             args.xmax = max(starts) + args.duration
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -891,7 +891,7 @@ class TimeDomainProduct(CliProduct):
         if args.epoch is None:
             args.epoch = args.xmin
         else:
-            epoch = float(args.epoch)
+            epoch = args.epoch
             if (epoch < 1e8):
                 args.epoch = LIGOTimeGPS(epoch + min(starts))
 
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -893,7 +893,7 @@ class TimeDomainProduct(CliProduct):
         else:
             epoch = float(args.epoch)
             if (epoch < 1e8):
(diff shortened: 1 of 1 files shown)


gwpy/cli/cliproduct.py
```diff
@@ -888,6 +890,11 @@ def _finalize_arguments(self, args):
             args.xmin = min(starts)
         if args.epoch is None:
             args.epoch = args.xmin
+        else:
+            epoch = float(args.epoch)
```
duncanmmacleod: @areeda, why is `args.epoch` cast to a new variable here? why not just

```python
        if args.epoch is None:
            args.epoch = args.xmin
        elif args.epoch < 1e8:  # user gave epoch relative to start
            args.epoch += min(starts)
```

areeda: Does that work with LIGOTimeGPS? I thought it was necessary.

duncanmmacleod: Yes, that should work, `LIGOTimeGPS` is compatible with floats.

duncanmmacleod:
@areeda, can you test and implement my simpler change as suggested above? I don't think there's a need for separate casting of epoch to a new variable via `LIGOTimeGPS`.

areeda:
I think I hve done what Duncan suggested.

duncanmmacleod:
@areeda, your latest change isn't quite what I had in mind; the code exactly as below is slightly more efficient a change, IMO, please give that a go:

```diff
diff --git a/gwpy/cli/cliproduct.py b/gwpy/cli/cliproduct.py
index d333553d..caa3aedd 100644
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -877,7 +877,8 @@ class TimeDomainProduct(CliProduct):
         """
         group = super(TimeDomainProduct, cls).arg_xaxis(parser)
         group.add_argument('--epoch', type=to_gps,
-                           help='center X axis on this GPS time')
+                           help='center X axis on this GPS time, may be'
+                                'absolute date/time or delta')
         return group

     def _finalize_arguments(self, args):
@@ -888,6 +889,11 @@ class TimeDomainProduct(CliProduct):
             args.xmin = min(starts)
         if args.epoch is None:
             args.epoch = args.xmin
+        elif args.epoch < 1e8:
+            args.epoch += min(starts)
         if args.xmax is None:
             args.xmax = max(starts) + args.duration
         return super(TimeDomainProduct, self)._finalize_arguments(args)
```

gwpy/cli/cliproduct.py
```diff
@@ -888,6 +889,11 @@ def _finalize_arguments(self, args):
             args.xmin = min(starts)
         if args.epoch is None:
             args.epoch = args.xmin
+        else:
+            epoch = args.epoch
+            if (epoch < 1e8):
+                args.epoch += min(starts)
```
duncanmmacleod: @areeda, this can still be simplified, such that no new variables are used at all:

```diff
@@ -888,6 +889,8 @@ def _finalize_arguments(self, args):
             args.xmin = min(starts)
         if args.epoch is None:
             args.epoch = args.xmin
+        elif args.epoch < 1e8:
+            args.epoch += min(starts)
```

Can you please implement this instead?

areeda: How's this one?

PR #883, opened 2018-08-24
Introducing a TimeSeries.convolve() method based on the overlap-save algorithm

### LINKED ISSUE #451: Applying FIR filters to long timeseries

alurban:
@duncanmmacleod, this PR introduces a `TimeSeries.convolve()` method which efficiently applies an FIR filter to a long timeseries using the [overlap-save](https://en.wikipedia.org/wiki/Overlap–save_method) algorithm. It also adds a unit test for `TimeSeries.convolve()` which prepares a timeseries and then checks that the input data are unchanged by convolving with a delta function.

@andrew-lundgren, this fixes #451.

### DIFF
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -835,6 +835,17 @@ class TestTimeSeries(_TestTimeSeriesBase):
         tmax = whitened.times[whitened.argmax()]
         nptest.assert_almost_equal(tmax.value, glitchtime)
 
+    def test_convolve(self):
+        data = self.TEST_CLASS(
+            signal.hann(1024), sample_rate=512, epoch=-1
+        )
+        filt = numpy.array([1, 0])
+
+        # check that the 'valid' data are unchanged by this filter
+        convolved = data.convolve(filt)
+        assert convolved.size == data.size
+        utils.assert_allclose(convolved.value[1:-1], data.value[1:-1])
+
     def test_detrend(self, losc):
         assert not numpy.isclose(losc.value.mean(), 0.0, atol=1e-21)
         detrended = losc.detrend()
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1500,8 +1500,8 @@ class TimeSeries(TimeSeriesBase):
         --------
         TimeSeries.asd
             for details on the ASD calculation
-        scipy.signal.fftconvolve
-            for details on the convolution algorithm used here
+        TimeSeries.convolve
+            for details on convolution with the overlap-save method
         gwpy.signal.filter_design.fir_from_transfer
             for FIR filter design through spectrum truncation
 
@@ -1511,7 +1511,8 @@ class TimeSeries(TimeSeriesBase):
         and in preventing spectral leakage in the output.
 
         Due to filter settle-in, a segment of length `0.5*fduration` will be
-        corrupted at the beginning and end of the output.
+        corrupted at the beginning and end of the output. See
+        `~TimeSeries.convolve` for more details.
 
         The input is detrended to give the whitened `TimeSeries` zero mean,
         and the output is normalised to give it unit variance.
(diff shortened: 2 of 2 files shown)


duncanmmacleod:
@alurban, can you take a look at the failing tests? it looks like this presumes some minimal version of scipy that is beyond what the LDG clusters can support. I do plan to move away from supporting those super-old versions of numpy and scipy, but would like to at least hit 1.0.0 with LDG compatibility.

alurban:
@duncanmmacleod, the basic issue was that prior to scipy 0.19.0, `scipy.signal.convolve()` only ever used direct convolution. I've attempted to address this by switching to `scipy.signal.fftconvolve`, which (1) still exists in subsequent scipy versions, (2) always convolves in the frequency domain, and (3) is faster than direct convolution for filters longer than ~500 taps. (In scipy >= 0.19.0, calls to `signal.convolve()` automatically invoke `fftconvolve` when the filter is long enough.)

I've also tried to address concerns from coveralls and codeclimate by cleaning up the use cases and docstrings of new functions in `gwpy.signal.filter_design`.

gwpy/signal/filter_design.py
```diff
@@ -182,7 +183,7 @@ def truncate_transfer(transfer, ncorner=None):
     return out
 
 
-def truncate_impulse(impulse, ntaps, window='hanning'):
+def truncate_impulse(impulse, ntaps, window='hamming'):
```
duncanmmacleod: what's the motivation for this change?

alurban: I'm actually not sure why that's there, I can change it back if you like? (A Hanning window was the default for the old `TimeSeries.whiten` method.)

duncanmmacleod: @alurban, yes please, lets keep this PR as clean as possible

alurban: @duncanmmacleod, ok, I'll make that change shortly

alurban: This is done with the latest commit (which I've squashed with the main commit)

alurban:
@duncanmmacleod, I've noticed that the new whitening filter corrupts the left and right boundaries of the input signal with huge wings that go up to enormous values. I was cropping these off, but I've just pushed a change to `TimeSeries.convolve()` that instead smoothly tapers off the left and right boundaries of the input before FFT'ing, then returns a `TimeSeries` the same length as the input (and with the same `t0`). So, the whitened timeseries smoothly rises up from 0 at both ends, and the result is not cropped. This better reproduces the functionality of the original `TimeSeries.whiten()`.

In a moment I'll update the unit tests to reflect this.

alurban:
@duncanmmacleod, the offending Travis CI build failure looks like a transient SSL timeout, probably unrelated to recent changes. Can it be restarted?

alurban:
@duncanmmacleod, I believe I've worked out all the gotchas and this PR is ready for proper review. Below are plots of (1) the raw and whitened timeseries from Hanford for 64 seconds surrounding GW150914; (2) the whitened timeseries, zoomed in and bandpassed to show the BBH signal; and (3) the Q-transform spectrogram showing the BBH trace. All three plots were made with whitening defaults under the new changes. Note that in the spectrogram, the low-frequency power of the signal is restored, better reproducing the published spectrogram for GW150914.

150914_compare
150914_whitened_timeseries
150914_strain

alurban:
@duncanmmacleod, per your suggestion in a private slack message, I've posted an issue (#890) and a separate pull request (#891) for all the changes from this one that do not explicitly introduce significant new functionality. That PR can be merged for a 0.12.1 point release, while this one should be saved for 0.13.0/1.0.0 since `TimeSeries.convolve()` certainly is a significant new feature.

alurban:
@duncanmmacleod, after you merged #891 I rebased this branch against `upstream/develop` and resolved a few (minor and expected) conflicts. After pushing those changes, all the build checks have passed and this PR is ready whenever you are.

alurban:
@duncanmmacleod, I've rebased this branch against `upstream/develop` to get all the functionality introduced in gwpy-0.12.2.

alurban:
@duncanmacleod, yes, please merge at your discretion and I’ll rebase #905 so that it’s ready as well.

PR #884, opened 2018-08-25
Formatting qrange problem fixed

areeda:
Data type of qrange and frange changed from  str  to float because of labeling issues

### DIFF
--- a/gwpy/cli/qtransform.py
+++ b/gwpy/cli/qtransform.py
@@ -97,8 +97,10 @@ class Qtransform(Spectrogram):
         group = parser.add_argument_group('Q-transform options')
         group.add_argument('--plot', nargs='+', type=float, default=[.5],
                            help='One or more times to plot')
-        group.add_argument('--frange', nargs=2, help='Frequency range to plot')
-        group.add_argument('--qrange', nargs=2, help='Search Q range')
+        group.add_argument('--frange', nargs=2, type=float,
+                           help='Frequency range to plot')
+        group.add_argument('--qrange', nargs=2, type=float,
+                           help='Search Q range')
 
         group.add_argument('--nowhiten', action='store_true',
                            help='do not whiten input before transform')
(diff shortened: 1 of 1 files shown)


gwpy/cli/qtransform.py
```diff
@@ -149,6 +151,8 @@ def fformat(x):  # float format
                 return '[{0}]'.format(', '.join(map(fformat, x)))
             if isinstance(x, Quantity):
                 x = x.value
+            elif isinstance(x, str):
+                x = float(x)
```
duncanmmacleod: if you make the first change, casting the input `args.frange` and `args.qrange` to a `float`, under what circumstances is `x` a `str`?

gwpy/cli/qtransform.py
```diff
@@ -149,6 +151,11 @@ def fformat(x):  # float format
                 return '[{0}]'.format(', '.join(map(fformat, x)))
             if isinstance(x, Quantity):
                 x = x.value
+            elif isinstance(x, str):
+                self.log(0, 'WARNING: fformat called with a string. This has '
+                            'been depricated and may disappear '
+                            'in a future release.')
```
duncanmmacleod: @areeda, can we try using `warnings.warn(..., DeprecationWarning)` instead, so that the warnings can be caught easily during testing?

duncanmmacleod: warnings print the line that caused the warning to be raised, which here would just be the warning itself. I think that's a small price to pay for the ability to promote this warning to an error during testing to find if any code actually will touch this clause.

duncanmmacleod: or, we could just do something like this:

```python
        if isinstance(x, Quantity):
            x = x.value
        return '{0:.2f}'.format(float(x))
```

and cast everything to float regardless. This would only then error when non-`float()`able values present, which is an bug anyway.

PR #885, opened 2018-08-26
gwpy/cli/timeseries.py: limit sample rate to 3 decimals in plot title

areeda:
I got tired of minute trends reporting Fs to 10 decimal digits.  I verified VIrgo sample rates are not slower than .001 Hz

### DIFF
--- a/gwpy/cli/timeseries.py
+++ b/gwpy/cli/timeseries.py
@@ -57,7 +57,7 @@ class TimeSeries(TimeDomainProduct):
     def get_title(self):
         suffix = super(TimeSeries, self).get_title()
         # limit significant digits for minute trends
-        rates = {int((ts.sample_rate.value + 5e-4) * 1000) / 1000. * units.Hz
+        rates = {round(ts.sample_rate.value,3) * units.Hz
                  for ts in self.timeseries}
         fss = '({0})'.format('), ('.join(map(str, rates)))
         return ', '.join([
--- a/gwpy/cli/timeseries.py
+++ b/gwpy/cli/timeseries.py
@@ -25,7 +25,6 @@
 from .cliproduct import TimeDomainProduct
 from ..plot import Plot
 from ..plot.tex import label_to_latex
-from astropy import units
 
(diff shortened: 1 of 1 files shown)


gwpy/cli/timeseries.py
```diff
@@ -55,7 +56,9 @@ def get_suptitle(self):
 
     def get_title(self):
         suffix = super(TimeSeries, self).get_title()
-        rates = {ts.sample_rate for ts in self.timeseries}
+        # limit significant digits for minute trends
+        rates = {int((ts.sample_rate.value + 5e-4) * 1000) / 1000. * units.Hz
+                 for ts in self.timeseries}
         fss = '({0})'.format('), ('.join(map(str, rates)))
```
duncanmmacleod: @areeda, is there a reason not to use the builtin `round` function here?

gwpy/cli/timeseries.py
```diff
@@ -55,7 +56,9 @@ def get_suptitle(self):
 
     def get_title(self):
         suffix = super(TimeSeries, self).get_title()
-        rates = {ts.sample_rate for ts in self.timeseries}
+        # limit significant digits for minute trends
+        rates = {round(ts.sample_rate.value, 3) * units.Hz
+                 for ts in self.timeseries}
```
duncanmmacleod: I'm sorry @areeda, it turns out that Quantities have built-in support for rounding, so can you try using this:

```python
        rates = {ts.sample_rate.round(3) for ts in self.timeseries}
```

areeda: How's this/

gwpy/cli/timeseries.py
```diff
@@ -25,6 +25,7 @@
 from .cliproduct import TimeDomainProduct
 from ..plot import Plot
 from ..plot.tex import label_to_latex
+from astropy import units
```
duncanmmacleod: @areeda, can you please remove this import, which I don't think is needed anymore

PR #891, opened 2018-09-06
Address issues with whitening defaults

### LINKED ISSUE #890: Current whitening defaults considered harmful
Introduced in GWPy 0.12.0, the new `TimeSeries.whiten()` method is causing minor confusion with users who expect the output timeseries to have the same length and starting time as the input, like the older method gave. For a 0.12.1 point release, `whiten()` should return a non-cropped output. To prevent huge wings from spectral leakage during filter settle-in, a segment of data half the length of the whitening filter will need to be windowed on the left and right boundaries of the input.

A separate but related issue is that, with default settings, `TimeSeries.whiten()` suppresses the lower frequency signal power in GW150914. If the default filter length in `TimeSeries.whiten()` is bumped up to 2 seconds (same as the default FFT length for ASD measurement), then the published spectrogram is much better reproduced.

Finally, the guts of `TimeSeries.whiten()` can probably be refactored a bit so that all of the filter design is done in `~gwpy.signal.filter_design`, and so that the actual application of the filter is done using `scipy.signal.fftconvolve`, anticipating #883.

alurban:
@duncanmmacleod, this PR includes the following changes:

* Window the first and last half-filter-length of input data before convolving, and return a `TimeSeries` the same length as the input (this is mostly to stay more in line with the behavior of the old whitener)
* Update the documentation’s whitening example to reflect this
* Change default whitening filter length to 2 seconds, which better reproduces the published GW150914 Q-spectrogram
* Refactor FIR filter design into its own function, `signal.filter_design.fir_from_transfer()`, including a unit test
* Use `scipy.signal.fftconvolve()` to apply the whitening filter, anticipating #883
* Edit unit tests affected by changes to `TimeSeries.whiten()`

This fixes #890.

### DIFF
--- a/examples/timeseries/whiten.py
+++ b/examples/timeseries/whiten.py
@@ -55,4 +55,7 @@ plot.axes[1].set_ylabel('Whitened amplitude', fontsize=16)
 plot.show()
 
 # Here we see two large spikes that are completely undetected in the raw
-# `TimeSeries`, but are very obvious in the whitened data.
+# `TimeSeries`, but are very obvious in the whitened data. We can also see
+# tapering effects at the boundaries as the whitening filter settles in,
+# meaning that the first and last ~second of data are corrupted and should
+# be discarded before further processing.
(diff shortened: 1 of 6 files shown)


alurban:
@duncanmmacleod, I've rebased and force-pushed, and build checks are now in progress.

PR #894, opened 2018-09-07
WIP: Fix for failed tests on windows with pytest-3.8.0

duncanmmacleod:
This PR attempts to fix the failing tests on appveyor when using pytest-3.8.0. Something changed with the way warnings are collected...


duncanmmacleod:
These fixes have not worked, but I think I have isolated the issue to something in pytest that only impacts python2.7, see pytest-dev/pytest#3954. Hopefully that issue will get resolved or a workaround will be provided.

duncanmmacleod:
I think with pytest-3.9.0 this is no longer an issue.

PR #903, opened 2018-09-12
Fix normalization for whitening and FIR filter design

### LINKED ISSUE #902: FIR filter amplitude is inconsistent
In `signal.filter_design.fir_from_transfer()`, the overall amplitude scaling of the final filter is off by a factor of `Nsize * Ntaps / 2`, where `Nsize` is the length of the input array and `Ntaps` the requested number of taps. The fix should be very simple, and will impact the amplitude of whitening filters designed in TimeSeries.whiten().

@duncanmmacleod, this should be a 0.12.1 milestone. Sorry for the confusion! I’ll post a PR as soon as I can.

alurban:
@duncanmmacleod, this PR adjusts the normali(s)ation in `signal.filter_design.fir_from_transfer` so that the resulting filter's frequency response has the same overall amplitude as the original transfer function. **Crucially, this means FIR filter design will use the `numpy` normalization.** To compensate for this, the PR also fixes normalization in `TimeSeries.whiten()` so that the output timeseries has zero mean and unit variance:

150914_spectra_compare-2

150914_compare

150914_spectra

First plot shows the original spectrum in blue and the frequency response of the filter in orange; second plot shows the original timeseries on top and the whitened timeseries (note the y-axis scale) on the bottom; third plot compares the original strain ASD to that of the whitened data.

This fixes #902.

### DIFF
--- a/gwpy/signal/filter_design.py
+++ b/gwpy/signal/filter_design.py
@@ -252,7 +252,7 @@ def fir_from_transfer(transfer, ntaps, window='hanning', ncorner=None):
     impulse = npfft.irfft(transfer)
     impulse = truncate_impulse(impulse, ntaps=ntaps, window=window)
     # wrap around and normalise to construct the filter
-    out = numpy.roll(impulse, int(ntaps/2 - 1))[0:ntaps] / impulse.size
+    out = numpy.roll(impulse, int(ntaps/2 - 1))[0:ntaps] * ntaps/2
     return out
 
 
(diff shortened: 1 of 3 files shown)


duncanmmacleod:
@alurban, can you add some documentation of the normalisations in the relevant docstrings?

alurban:
@duncanmmacleod, good suggestion, this is done in commit e0ecdab.

alurban:
thanks @duncanmmacleod!

PR #905, opened 2018-09-13
Introducing a TimeSeries.correlate() method

### LINKED ISSUE #892: Add support for matched filtering and cross-correlation
It would be useful to have `TimeSeries` methods for matched filtering and/or cross-correlation. These can be built around the `TimeSeries.convolve()` method introduced in #883 once that PR is merged. For one possible use case and an idea for implementation, see gwdetchar/gwdetchar#95.

alurban:
@duncanmmacleod, this pull request makes the following changes:

* Introduce a method, `TimeSeries.correlate()`, which convolves `self` with the phase reverse of a given signal, yielding the matched-filter SNR
* The new method whitens both the input and the matched-filter if requested, using an ASD of the input followed by two calls to `TimeSeries.whiten()`
* Add a unit test and documentation example for the new method (the latter requires `LIGO.ORG` credentials)
* Update `TimeSeries.whiten()` to normalize by expected standard deviation, rather than observed
* Update the unit test for `TimeSeries.whiten()`

Note, the SNR timeseries returned by `TimeSeries.correlate()` has the same length and timestamps as `self`. Both `self` and the matched-filter argument are detrended (to give the output zero mean), and the output SNR is normalized by its expected standard deviation (to give unit variance). Users will have to remember to take the absolute value of the output if they don't care about minus signs.

This should be considered a target for gwpy 0.13.0/1.0.0.

This fixes #892.

### DIFF
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1514,8 +1514,9 @@ class TimeSeries(TimeSeriesBase):
         corrupted at the beginning and end of the output. See
         `~TimeSeries.convolve` for more details.
 
-        The input is detrended to give the whitened `TimeSeries` zero mean,
-        and the output is normalised to give it unit variance.
+        The input is detrended and the output normalised such that, if the
+        input is stationary and Gaussian, then the output will have zero mean
+        and unit variance.
 
         For more on inverse spectrum truncation, see arXiv:gr-qc/0509116.
         """
@@ -1533,7 +1534,7 @@ class TimeSeries(TimeSeriesBase):
         # condition the input data and apply the whitening filter
         in_ = self.copy().detrend(detrend)
         out = in_.convolve(tdw, window=window)
-        return out / out.value.std()
+        return out * numpy.sqrt(2 * in_.dt.decompose().value)
 
     def convolve(self, fir, window='hanning'):
         """Convolve this `TimeSeries` with an FIR filter using the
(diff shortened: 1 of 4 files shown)


alurban:
@duncanmmacleod, the example introduced in `examples/timeseries/correlate.py` produces the following plots:

* SNR timeseries of h(t) cross-correlated with the PSL channel
whistle_snr

* Q-scan of h(t)
whistle_hoft

* Q-scan of PSL photodiode
whistle_psl

@tjma12, you may also be interested in this functionality for various reasons.

alurban:
@duncanmmacleod, I've rebased this branch against `upstream/develop` to get all the functionality introduced in gwpy-0.12.2. Note, this PR is ready for review but still marked as WIP because it depends on #883, which should be merged first.

gwpy/timeseries/timeseries.py
```diff
@@ -1601,6 +1601,87 @@ def convolve(self, fir, window='hanning'):
...
+            if overlap is None:
+                overlap = recommended_overlap(window) * fftlength
+            psd = self.psd(fftlength, overlap, window=window, **psd_kw)
+            # overwhiten (with errors on division by zero)
+            with numpy.errstate(all='raise'):
+                in_ = self.copy().whiten(window=window, asd=psd,
+                                         fduration=wduration,
+                                         highpass=highpass)
+        else:
+            in_ = self.copy()
```
duncanmmacleod: why is this `copy()` required?

alurban: I think I was attempting to be safe by not performing any operations in-place. Would it be better to replace `in_ = self.copy().whiten(...)` with `self.whiten(...)` on line 1674, and to replace `in_ = self.copy()` with `in_ = self` on line 1678?

alurban: This is done in the most recent commit, let me know if you like this change better.

alurban:
@duncanmmacleod, after talking to @andrew-lundgren, I've updated this PR to more gracefully normalize both the whitening filter and the matched-filter SNR based on the _expected_ standard deviation rather than the _observed_ one. The upshot is that `TimeSeries.whiten()` and `TimeSeries.correlate()` will return Guassian-distributed data with zero mean and unit variance **only if** the input is Guassian and stationary, and will show deviations above Guassian noise elsewise. I've also updated the unit tests for `TimeSeries.whiten()` and `TimeSeries.correlate` which should now be more complete.

PR #921, opened 2018-10-10
Require lscsoft-glue >= 1.60.0

### LINKED ISSUE #928: Update lscsoft-glue dependence to 1.60+
The `glue.segments` module is also being deprecated in favor of `ligo.segments`.

alurban:
This fixes #928.

cc @duncanmmacleod

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -10,7 +10,7 @@
 # extras
 pymysql
 pyRXP
-lscsoft-glue
+lscsoft-glue >= 1.60.0
 lalsuite
 dqsegdb
 gwdatafind


duncanmmacleod:
@alurban, thanks for this, unfortunately `lscsoft-glue-1.60.0` is not available from pypi.python.org yet, I have requested this from @rpfisher. Once that is available I can rerun all of the CI.

alurban:
thanks @duncanmmacleod, sounds like this will affect gwdetchar/gwdetchar#167 and gwdetchar/hveto#89 as well.

alurban:
@duncanmmacleod, `lscsoft-glue-1.60.0` has been uploaded to PyPI, so this should now be good to roll.

duncanmmacleod:
@alurban, given that `gwpy` has no dependence upon `glue.segments` why is this change necessary?

alurban:
@duncanmmacleod, you are definitely a much better arbiter of whether it's necessary than I am, I was just hunting down any reference to `lscsoft-glue` in repositories with DetChar tools and changing them with reckless abandon. I'm happy enough to have this PR closed without merging if GWPy doesn't need it

duncanmmacleod:
@alurban, in that case I suggest closing this ticket without merging, but we can come back to it at a later date if we need to.

PR #925, opened 2018-10-16
Fixed fetching data with nds1

duncanmmacleod:
This PR fixes a few bugs in fetching data using the `nds2` client from an **NDS1** server.


SeanDS:
I noticed the trends bug a couple of days ago as well, on our NDS1 frame builder at the AEI. I'd be happy to test this but I installed everything via conda so it seems simpler to wait till the next release.

PR #933, opened 2018-10-18
Fixed optimisation regression in framecpp.py

duncanmmacleod:
This PR fixes a significant speed regression introduced by #924.


duncanmmacleod:
The coverage reduction is due to a failure uploading coverage to coveralls.

PR #953, opened 2018-10-30
Refactoring of unix CI jobs

duncanmmacleod:
This PR refactors the unix CI jobs between CircleCI and TravisCI, mainly moving the docker stuff to CircleCI, which has better support.


duncanmmacleod:
The coverage reduction from this change is down to two things:

1. none of pip, conda, or macports can install `frameCPP` (yet)
1. gw-openscience.org is down (at time of writing), so none of the LOSC-data-based tests can run

I think (1) will persist a little while until the conda packages for `python-ldas-tools-framecpp` are finished. However, the python2.7 tests runnin on circleCI _are_ testing that interface, but they just can't report to coveralls.

PR #972, opened 2018-11-20
Fixing both the scraping of env at load and the error you get if you perform a sql query that returns an empty result

### LINKED ISSUE #855: Gravityspy functions should scrape environment at execution time, not import
@scottcoughlin2014, currently the GravitySpy functions that scrape the environment when the module is imported for the first time, rather than at execution time:

https://github.com/gwpy/gwpy/blob/879b044f0c39c79f3d19512ba465cd61c9dea93a/gwpy/table/io/gravityspy.py#L93-L96

Using `user=None, passwd=None` as defaults and scraping the environment inside the function will allows users to manipulate their environment (via `os.environ`) on-the-fly if they wish.

scottcoughlin2014:
This fixes #855.

### DIFF
--- a/gwpy/table/io/gravityspy.py
+++ b/gwpy/table/io/gravityspy.py
@@ -95,6 +95,15 @@ def get_connection_str(db='gravityspy',
                        user=None,
                        passwd=None):
     """Create string to pass to create_engine
+
+    Note:
+        If not supplied, user and passwd are
+        resolved by scraping your environement for
+        ``GRAVITYSPY_DATABASE_USER`` and
+        ``GRAVITYSPY_DATABASE_PASSWD``.
+
+        *Both* user and passwd are required otherwise
+        it is ignored.
     """
     if (not user) or (not passwd):
         user = os.getenv('GRAVITYSPY_DATABASE_USER', None)
--- a/gwpy/table/io/gravityspy.py
+++ b/gwpy/table/io/gravityspy.py
@@ -96,14 +96,31 @@ def get_connection_str(db='gravityspy',
                        passwd=None):
     """Create string to pass to create_engine
 
-    Note:
-        If not supplied, user and passwd are
-        resolved by scraping your environement for
-        ``GRAVITYSPY_DATABASE_USER`` and
-        ``GRAVITYSPY_DATABASE_PASSWD``.
+    Parameters
+    ----------
+    db : `str`, default: ``gravityspy``
+        The name of the SQL database your connecting to.
 
-        *Both* user and passwd are required otherwise
-        it is ignored.
+    host : `str`, default: ``gravityspy.ciera.northwestern.edu``
+        The name of the server the database you are connecting to
+        lives on.
+
+    db : `str`, default: None
+        Your username for authentication to this database.
+
+    db : `str`, default: None
+        Your password for authentication to this database.
+
+    .. note::
+
+       `user` and `passwd` should be given together, otherwise they will be
+       ignored and values will be resolved from the
+       ``GRAVITYSPY_DATABASE_USER`` and ``GRAVITYSPY_DATABASE_PASSWD``
+       environment variables.
+
+    Returns
+    -------
+    conn_string : A SQLAlchemy engine compliant connection string
     """
     if (not user) or (not passwd):
         user = os.getenv('GRAVITYSPY_DATABASE_USER', None)
--- a/gwpy/table/io/gravityspy.py
+++ b/gwpy/table/io/gravityspy.py
@@ -105,10 +105,10 @@ def get_connection_str(db='gravityspy',
         The name of the server the database you are connecting to
         lives on.
 
-    db : `str`, default: None
+    user : `str`, default: `None`
         Your username for authentication to this database.
 
-    db : `str`, default: None
+    passwd : `str`, default: `None`
         Your password for authentication to this database.
 
     .. note::
@@ -120,7 +120,8 @@ def get_connection_str(db='gravityspy',
 
     Returns
     -------
-    conn_string : A SQLAlchemy engine compliant connection string
+    conn_string : `str`
+        A SQLAlchemy engine compliant connection string
     """
     if (not user) or (not passwd):
         user = os.getenv('GRAVITYSPY_DATABASE_USER', None)
--- a/gwpy/table/io/gravityspy.py
+++ b/gwpy/table/io/gravityspy.py
@@ -92,10 +92,14 @@ def get_gravityspy_triggers(tablename, engine=None, **kwargs):
 
 def get_connection_str(db='gravityspy',
                        host='gravityspy.ciera.northwestern.edu',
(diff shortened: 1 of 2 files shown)


scottcoughlin2014:
@alurban I have done testing myself, I think the unit tests would catch any issues based on the `tab.empty` statement but probably not passing the `user` and `passwd` args. However I tested that myself and basically followed Duncan's suggestion in the reference issue

scottcoughlin2014:
@alurban do you think we could get this in `0.12.3`. I ask because the error you get if you do a `SQL` query that returns no results is super frustrating and I think will save people headaches. I should have done a PR a long time ago.

alurban:
@scottcoughlin2014, no worries, and yes, definitely we can get this in the next point release. That will likely be 0.13.0 based on discussions with @duncanmmacleod.

gwpy/table/io/gravityspy.py
```diff
@@ -92,10 +92,14 @@ def get_gravityspy_triggers(tablename, engine=None, **kwargs):
...
                        host='gravityspy.ciera.northwestern.edu',
-                       user=os.getenv('GRAVITYSPY_DATABASE_USER', None),
-                       passwd=os.getenv('GRAVITYSPY_DATABASE_PASSWD', None)):
+                       user=None,
+                       passwd=None):
     """Create string to pass to create_engine
     """
+    if (not user) or (not passwd):
+        user = os.getenv('GRAVITYSPY_DATABASE_USER', None)
+        passwd = os.getenv('GRAVITYSPY_DATABASE_PASSWD', None)
```
duncanmmacleod: @scottcoughlin2014, is it intention that `user` and `passwd` are _both_ required, otherwise are ignored? If so, can this be made clear in the docstring? Otherwise we should parse each separately.

gwpy/table/io/gravityspy.py
```diff
@@ -92,10 +92,14 @@ def get_gravityspy_triggers(tablename, engine=None, **kwargs):
 
 def get_connection_str(db='gravityspy',
                        host='gravityspy.ciera.northwestern.edu',
-                       user=os.getenv('GRAVITYSPY_DATABASE_USER', None),
-                       passwd=os.getenv('GRAVITYSPY_DATABASE_PASSWD', None)):
+                       user=None,
+                       passwd=None):
```
duncanmmacleod: @scottcoughlin2014, can you add a docstring for this method that notes how variables are resolved from `os.environ` if not given?

gwpy/table/io/sql.py
```diff
@@ -93,7 +93,8 @@ def fetch(engine, tablename, columns=None, selection=None, **kwargs):
     # Convert unicode columns to string
     types = tab.apply(lambda x: pd.api.types.infer_dtype(x.values))
 
-    for col in types[types == 'unicode'].index:
-        tab[col] = tab[col].astype(str)
+    if not tab.empty:
```
duncanmmacleod: why is this new clause required?

duncanmmacleod:
@scottcoughlin2014, @alurban, I have restarted the appveyor build.

gwpy/table/io/gravityspy.py
scottcoughlin2014: Ya I can do this

gwpy/table/io/sql.py
scottcoughlin2014: Because if the query is valid but the resulting table is empty then that loop fails as it never knows what `type` each column is because no columns where filled in.

gwpy/table/io/gravityspy.py
scottcoughlin2014: Yes both are needed, I will mkae clear

gwpy/table/io/gravityspy.py
scottcoughlin2014: @duncanmmacleod I have attempted to add more helpful doc strings let me know what you think

scottcoughlin2014:
@duncanmmacleod or @alurban  Let me know if the doc string where not satisfactory.

gwpy/table/io/gravityspy.py
```diff
@@ -92,10 +92,23 @@ def get_gravityspy_triggers(tablename, engine=None, **kwargs):
...
     """Create string to pass to create_engine
+
+    Note:
+        If not supplied, user and passwd are
+        resolved by scraping your environement for
+        ``GRAVITYSPY_DATABASE_USER`` and
+        ``GRAVITYSPY_DATABASE_PASSWD``.
+
+        *Both* user and passwd are required otherwise
+        it is ignored.
```
duncanmmacleod: @scottcoughlin2014, can you add `Parameters` and `Returns` sections in the docstring, see other functions for examples?

Also, the final sentence here, I think, is ambiguous. Perhaps something like (adding sphinx formatting):

```
    .. note::

       `user` and `passwd` should be given together, otherwise they will be
       ignored and values will be resolved from the ``GRAVITYSPY_DATABASE_USER``
       and ``GRAVITYSPY_DATABASE_PASSWD`` environment variables.

File: gwpy/table/io/gravityspy.py [hunk 5]
```diff
@@ -96,14 +96,31 @@ def get_connection_str(db='gravityspy',
...
+    Parameters
+    ----------
+    db : `str`, default: ``gravityspy``
+        The name of the SQL database your connecting to.
+
+    host : `str`, default: ``gravityspy.ciera.northwestern.edu``
+        The name of the server the database you are connecting to
+        lives on.
+
+    db : `str`, default: None
```
duncanmmacleod:
```diff
-    db : `str`, default: None
+    user : `str`, default: `None`
```

gwpy/table/io/gravityspy.py
```diff
@@ -96,14 +96,31 @@ def get_connection_str(db='gravityspy',
...
+        Your username for authentication to this database.
+
+    db : `str`, default: None
```
duncanmmacleod:
```diff
-    db : `str`, default: None
+    passwd : `str`, default: `None`
```

gwpy/table/io/gravityspy.py
```diff
@@ -96,14 +96,31 @@ def get_connection_str(db='gravityspy',
...
+    .. note::
+
+       `user` and `passwd` should be given together, otherwise they will be
+       ignored and values will be resolved from the
+       ``GRAVITYSPY_DATABASE_USER`` and ``GRAVITYSPY_DATABASE_PASSWD``
+       environment variables.
+
+    Returns
+    -------
+    conn_string : A SQLAlchemy engine compliant connection string
```
duncanmmacleod:
```
    conn_string : `str`
        A SQLAlchemy engine compliant connection string
```

duncanmmacleod:
@scottcoughlin2014, please rebase against `gwpy/gwpy@develop` to correct the appveyor failures.

duncanmmacleod:
@wip ready for review

PR #974, opened 2018-11-21
Patch a memory leak in TimeSeries.demodulate()

### LINKED ISSUE #973: Memory leaks in TimeSeries.demodulate()
The `TimeSeries.demodulate()` method currently calls `self.times` which leads to a memory leak for very long timeseries. This can be fixed by using indices instead.

alurban:
@duncanmmacleod, this PR patches up a memory leak I discovered while helping @siddharth101 track 50 Hz lines at Livingston. The issue is in `TimeSeries.demodulate()`, which originally contained a call to the `self.times` object, a massive memory hog for long timeseries. The fix I've implemented here is to instead range over sample indices, breaking them up into slices of length `stridesamp`.

The following changes are also applied:

* Change the reference time offset to `self.t0`, rather than GPS second 0
* Mix with a local oscillator inside the `for` loop, to limit its size in memory
* Replace use of `__metadata_finalize__()` with `__array_finalize__` in constructing the output in memory
* Minor refinement to the docstring
* Tweak the worked example in the docstring

This fixes #973.

### DIFF
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1315,8 +1315,8 @@ class TimeSeries(TimeSeriesBase):
             stride (seconds) between calculations, defaults to 1 second
 
         exp : `bool`, optional
-            return the demodulated magnitude and phase trends as one
-            `TimeSeries` object representing a complex exponential
+            return the magnitude and phase trends as one `TimeSeries` object
+            representing a complex exponential, default: False
 
         deg : `bool`, optional
             if `exp=False`, calculates the phase in degrees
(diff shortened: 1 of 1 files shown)


gwpy/timeseries/timeseries.py
```diff
@@ -1336,46 +1336,46 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
+        >>> ax = plot.gca()
+        >>> ax.set_ylabel('Strain Amplitude at 331.3 Hz')
         >>> plot.show()
         """
         stridesamp = int(stride * self.sample_rate.value)
         nsteps = int(self.size // stridesamp)
-        # mix with a complex oscillator and stride through the TimeSeries,
+        # stride through the TimeSeries and mix with a local oscillator,
         # taking the average over each stride
         out = numpy.zeros(nsteps, dtype=complex).view(type(self))
```
duncanmmacleod: Can you modify this to be

```python
        out = type(self)(numpy.zeros(nsteps, dtype=complex))
```

This will result in `out` owning its own data, rather than being a view of something else:

```python
>>> TimeSeries(numpy.zeros(10, dtype=complex)).flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
>>> numpy.zeros(10, dtype=complex).view(TimeSeries).flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

which makes modifying `out` easier downstream.

alurban: @duncanmmacleod aha, so that's the difference between `type(self)(...)` and `(...).view(type(self))`. Thanks for the explanation, this is updated in the latest push.

alurban:
Thanks @duncanmmacleod, I've made the requested change. Happy Thanksgiving!

PR #977, opened 2018-11-27
Updated default rcParams to not use LaTeX by default

duncanmmacleod:
This PR modifies the default rcParams (`gwpy.plot.rc`) to **not use TeX** by default (`GWPY_USETEX` will default to `False`). To accommodate this change, some of the other params have been tweaked to make the plots look better using non-tex formatting. There is also a new `GWPY_RCPARAMS` environment variable to disable updating _any_ parameters.

### DIFF
--- a/gwpy/table/table.py
+++ b/gwpy/table/table.py
@@ -546,6 +546,7 @@ class EventTable(Table):
         return self._plot('tile', self[x], self[y], self[w], self[h], **kwargs)
 
     def _plot(self, method, *args, **kwargs):
+        from matplotlib import rcParams
         from ..plot import Plot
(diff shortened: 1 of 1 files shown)


alurban:
@duncanmmacleod, thanks for putting this together, can you post an example made using these changes?

duncanmmacleod:
@alurban, see [here [LIGO.ORG]](https://ldas-jobs.ligo.caltech.edu/~duncan.macleod/gwpy/testing/977) for a documentation build from this update.

duncanmmacleod:
@areeda, can you take a look at this as well. This will probable have a number of implications for LDVW.

Issue #975, opened 2018-11-27
--title option to gwpy-plot raises TypeError


duncanmmacleod:
As reported by @borjasor, the `--title` option to `gwpy-plot` doesn't work as designed:

```pytb
$ gwpy-plot spectrogram --chan H1:PEM-CS_MIC_LVEA_INPUTOPTICS_DQ,raw --start 1227247141 --duration 1200 --out ldvw5168332937436378342.png --fmin 5 --fmax 1024 --imin 0.250 --imax 4.000 --norm --cmap Spectral_r --title TEST```
Traceback (most recent call last):
  File "/home/detchar/opt/summary-2.7/bin/gwpy-plot", line 105, in <module>
    prod.run()
  File "/home/detchar/opt/summary-2.7/lib/python2.7/site-packages/gwpy/cli/cliproduct.py", line 708, in run
    self.set_plot_properties()
  File "/home/detchar/opt/summary-2.7/lib/python2.7/site-packages/gwpy/cli/cliproduct.py", line 549, in set_plot_properties
    self.set_title(self.args.title)
  File "/home/detchar/opt/summary-2.7/lib/python2.7/site-packages/gwpy/cli/cliproduct.py", line 648, in set_title
    title = label_to_latex(title)
  File "/home/detchar/opt/summary-2.7/lib/python2.7/site-packages/gwpy/plot/tex.py", line 148, in label_to_latex
    for m in re_latex_control.finditer(text):
TypeError: expected string or buffer
```

I believe the issue is that the `--title option is being parsed as a list:

https://github.com/gwpy/gwpy/blob/96ecf1fd701c681427a2f9b64437388c7842c04c/gwpy/cli/cliproduct.py#L340-L341

but then passed as a single argument to the `ax.set_title` method:

https://github.com/gwpy/gwpy/blob/96ecf1fd701c681427a2f9b64437388c7842c04c/gwpy/cli/cliproduct.py#L641-L653

which ends up with a `re.Pattern` trying to match a `list` to an expression:

https://github.com/gwpy/gwpy/blob/96ecf1fd701c681427a2f9b64437388c7842c04c/gwpy/plot/tex.py#L112-L158

cc @areeda

PR #979, opened 2018-11-27
Latex enabled if necessay and title list handled properly

areeda:
Addresses issue #975 and reinstates latex when needed

### DIFF
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -704,7 +704,9 @@ class CliProduct(object):
             if isinstance(argsd[key], str):
                 hastex |= '$' in argsd[key]
         hastex |= self.args.mode == 'spectrum'
-        rc('text', usetex=hastex)
+        # ignore whether or not TeX is used resulting in weird labeling
+        # if TeX is diabled in matplotlib.rc or ia environmet
+#        rc('text', usetex=hastex)
 
         # grab the data
         self.get_data()
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -34,7 +34,7 @@ from functools import wraps
 
 from six import add_metaclass
 
-from matplotlib import rcParams
+from matplotlib import rcParams, rc
 try:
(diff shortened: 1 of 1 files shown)


gwpy/cli/cliproduct.py
```diff
@@ -639,18 +639,20 @@ def set_legend(self):
...
+
         if title is None:
-            title = self.get_title().rstrip(', ')
-        if self.usetex:
-            title = label_to_latex(title)
-        if title:
-            self.ax.set_title(title, fontsize=12)
-            self.log(3, ('Title is: %s' % title))
+            title = [self.get_title().rstrip(', ')]
+        for title_line in title:
```
duncanmmacleod: @areeda, if the user gives multiple arguments to `--title`, doesn't this mean that only the last given `title_line` will be preserved?

gwpy/cli/cliproduct.py
```diff
@@ -695,9 +697,14 @@ def run(self):
...
 
         self.log(3, 'Arguments:')
+        hastex = False
         argsd = vars(self.args)
         for key in sorted(argsd):
             self.log(3, '{0:>15s} = {1}'.format(key, argsd[key]))
+            if isinstance(argsd[key], str):
+                hastex |= '$' in argsd[key]
+        hastex |= self.args.mode == 'spectrum'
+        rc('text', usetex=hastex)
```
duncanmmacleod: @areeda, I'm not sure this is the right way to go, given that matplotlib has (some) support for mathtext without using latex. Can you experiment with always using functions like `label_to_latex` but _not_ using `text.usetex`?

gwpy/cli/cliproduct.py
duncanmmacleod: @areeda, ping?

areeda:
Withdrawn #977 now resolves issue.

duncanmmacleod:
@areeda, #977 does not fix the `title` issues that this PR address. Can you rebase this PR to only implement those fixes and reopen, or post a new PR?

PR #983, opened 2018-11-29
Allow users to save time and memory computing Q-scans

alurban:
This PR gives users the option to sample Q-scan spectrograms logarithmically in frequency, so that they can render Qscan plots more quickly (and using less memory) by invoking `Spectrogram.pcolormesh()` with a much smaller input array.

Other memory-saving changes in this PR include:
* Define an `ndarray` x-index proxy for interpolating Q-tiles, as this will use less memory than explicit calls to `series.times`
* Add a unit test for the `logf=True` use case
* Minor tweaks to unit tests affected by these changes

cc @duncanmmacleod, @areeda, @tjma12

### DIFF
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -929,7 +929,7 @@ class TestTimeSeries(_TestTimeSeriesBase):
         assert isinstance(qspecgram, Spectrogram)
         assert qspecgram.shape == (4000, 2403)
         assert qspecgram.q == 5.65685424949238
-        nptest.assert_almost_equal(qspecgram.value.max(), 156.0411964254892)
+        nptest.assert_almost_equal(qspecgram.value.max(), 156.43279233248313)
 
         # test whitening args
         asd = losc.asd(2, 1, method='scipy-welch')
@@ -955,6 +955,15 @@ class TestTimeSeries(_TestTimeSeriesBase):
         losc.q_transform(method='scipy-welch', norm=False)
         with pytest.raises(ValueError):
             losc.q_transform(method='scipy-welch', norm='blah')
+
+    def test_q_transform_logf(self, losc):
+        # test q-transform with log frequency spacing
+        qspecgram = losc.q_transform(method='scipy-welch', fftlength=2,
+                                     fres=500, logf=True)
+        assert isinstance(qspecgram, Spectrogram)
+        assert qspecgram.shape == (4000, 500)
+        assert qspecgram.q == 5.65685424949238
+        nptest.assert_almost_equal(qspecgram.value.max(), 156.43222488296405)
 
     def test_boolean_statetimeseries(self, array):
         comp = array >= 2 * array.unit
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1750,9 +1750,9 @@ class TimeSeries(TimeSeriesBase):
         return self.filter(*zpk, filtfilt=filtfilt)
 
     def q_transform(self, qrange=(4, 64), frange=(0, numpy.inf),
-                    gps=None, search=.5, tres=.001, fres=.5, norm='median',
-                    outseg=None, whiten=True, fduration=2, highpass=None,
-                    **asd_kw):
+                    gps=None, search=.5, tres=.001, fres=.5, logf=False,
+                    norm='median', outseg=None, whiten=True, fduration=2,
+                    highpass=None, **asd_kw):
         """Scan a `TimeSeries` using a multi-Q transform
 
         Parameters
@@ -1773,10 +1773,16 @@ class TimeSeries(TimeSeriesBase):
         tres : `float`, optional
             desired time resolution (seconds) of output `Spectrogram`
 
-        fres : `float`, `None`, optional
+        fres : `float`, `int`, `None`, optional
             desired frequency resolution (Hertz) of output `Spectrogram`,
             give `None` to skip this step and return the original resolution,
             e.g. if you're going to do your own interpolation
+
+        logf : `bool`, optional
+            boolean switch to enable (`True`) or disable (`False`) use of
+            log-sampled frequencies in the output `Spectrogram`,
+            if `True` then `fres` is interpreted as a number of frequency
+            samples, default: `False`
 
         norm : `bool`, `str`, optional
             whether to normalize the returned Q-transform output, or how,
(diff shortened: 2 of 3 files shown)


alurban:
@duncanmmacleod, here's an example plot made using `logf=True`, `fres=500`, and `Spectrogram.pcolormesh()`:



Compare with a Qscan of the same channel at the same time with the prior version of the code and `Spectrogram.imshow()`:

image

To my eye these look functionally identical.

gwpy/timeseries/timeseries.py
```diff
@@ -1920,32 +1931,30 @@ def q_transform(self, qrange=(4, 64), frange=(0, numpy.inf),
 
         # build regular Spectrogram from peak-Q data by interpolating each
         # (Q, frequency) `TimeSeries` to have the same time resolution
-        nx = int(abs(Segment(*outseg)) / tres)
+        nx = int(abs(outseg) / tres)
```
duncanmmacleod: @alurban , why is this change necessary?

gwpy/timeseries/timeseries.py
```diff
@@ -1920,32 +1931,30 @@ def q_transform(self, qrange=(4, 64), frange=(0, numpy.inf),
...
-                row.times.value, row.value)
-            out[:, i] = interp(out.times.value)
-
-        # then interpolate the spectrogram to increase the frequency resolution
-        # --- this is done because duncan doesn't like interpolated images
-        #     because they don't support log scaling
-        if fres is None:  # unless user tells us not to
+            xrow = numpy.linspace(0, 1, num=row.size)
+            interp = InterpolatedUnivariateSpline(xrow, row.value)
+            out[:, i] = interp(xout)
```
duncanmmacleod: @alurban, I don't understand what functional difference this change introduces? Can you explain this, please?

gwpy/timeseries/timeseries.py
```diff
@@ -1920,32 +1931,30 @@ def q_transform(self, qrange=(4, 64), frange=(0, numpy.inf),
...
         # interpolate rows
+        xout = numpy.linspace(0, 1, num=nx)
         for i, row in enumerate(norms):
-            interp = InterpolatedUnivariateSpline(
-                row.times.value, row.value)
-            out[:, i] = interp(out.times.value)
-
-        # then interpolate the spectrogram to increase the frequency resolution
-        # --- this is done because duncan doesn't like interpolated images
-        #     because they don't support log scaling
```
duncanmmacleod: @alurban, were these comments intentionally removed?

gwpy/timeseries/timeseries.py
alurban: @duncanmmacleod, this removes all references to a `Spectrogram.times` array and replaces them with a dummy variable, normalized between 0 and 1 so that the domain is consistent for interpolation. I find that this contributes to the reduction in memory overhead, as the production Omega scan code often reads in at least 64 seconds of data at a time per channel.

gwpy/timeseries/timeseries.py
alurban: @duncanmmacleod, yeah, these were intentionally removed. I was only trying to peacefully remove clutter, but if you want them put back I am happy to do that.

gwpy/timeseries/timeseries.py
alurban: @duncanmmacleod, the `outseg` object is already an instance of `~gwpy.segments.Segment`, which is enforced by the docstring, so the additional call to `Segment()` seemed redundant.

gwpy/timeseries/timeseries.py
alurban: I should clarify: doing this should have no (or very minor) impact on the output, but will reduce memory consumption for longer inputs.

alurban:
@duncanmmacleod, aha, thanks for catching that! In the most recent commit I switched to using `~numpy.logspace()`, added a note for Future Duncan and Future Alex, and added a unit test for the new `logf=True` use case to `test_q_transform()`.

gwpy/timeseries/timeseries.py
alurban: @duncanmmacleod, I had to go a different route to support using the `outseg` argument:

```python
# interpolate rows                                                      
xout = numpy.arange(outseg[0], outseg[1], tres)                         
for i, row in enumerate(norms):                                         
    xrow = numpy.arange(row.x0.value, (row.x0 + row.duration).value,    
                        row.dx.value)                                   
    interp = InterpolatedUnivariateSpline(xrow, row.value)              
    out[:, i] = interp(xout)
```

However, this still avoids an explicit call to `out.times` or `row.times`.

gwpy/timeseries/timeseries.py
duncanmmacleod: yes, please replace them.

gwpy/timeseries/timeseries.py
duncanmmacleod: Ok, thanks.

gwpy/timeseries/timeseries.py
duncanmmacleod: Technically, the docstring enforces nothing, but I get your point. However, having this here gives the user a little more freedom to just pass in a simple `tuple`, rather than a `Segment`. Can you replace it?

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -924,12 +924,19 @@ def test_notch(self, losc):
...
         assert isinstance(qspecgram, Spectrogram)
+        assert isinstance(qspecgram_logf, Spectrogram)
         assert qspecgram.shape == (4000, 2403)
+        assert qspecgram_logf.shape == (4000, 500)
         assert qspecgram.q == 5.65685424949238
-        nptest.assert_almost_equal(qspecgram.value.max(), 156.0411964254892)
+        assert qspecgram_logf.q == 5.65685424949238
+        nptest.assert_almost_equal(qspecgram.value.max(), 156.43279233248313)
+        nptest.assert_almost_equal(qspecgram_logf.value.max(),
+                                   156.43222488296405)
```
duncanmmacleod: @alurban, can you refactor this into two separate tests, something like `test_q_transform` and `test_q_transform_logf`? I think this will be cleaner to maintain.

gwpy/timeseries/timeseries.py
alurban: Thanks @duncanmmacleod, this is replaced in the latest commit.

gwpy/timeseries/timeseries.py
alurban: @duncanmmacleod, this is done in the latest commit.

gwpy/timeseries/tests/test_timeseries.py
alurban: @duncanmmacleod, that's a good idea, I did as you suggested and defined a separate unit test called `test_q_transform_logf`.

gwpy/timeseries/timeseries.py
```diff
@@ -1924,28 +1935,33 @@ def q_transform(self, qrange=(4, 64), frange=(0, numpy.inf),
...
-                          kind='cubic')
-        freqs2 = numpy.arange(planes.frange[0], planes.frange[1], fres)
-        new = Spectrogram(interp(out.times.value, freqs2 + fres/2.).T,
-                          x0=outseg[0], dx=tres,
-                          f0=planes.frange[0], df=fres)
+        interp = interp2d(xout, frequencies, out.value.T, kind='cubic')
+        if not logf:
+            outfreq = numpy.arange(planes.frange[0], planes.frange[1], fres)
+        else:
+            # using `~numpy.logspace` here to support numpy-1.1.7 for EPEL7,
```
duncanmmacleod: This should really say `numpy-1.7.1` but feel free to leave it, the comment serves its purpose.

alurban: @duncanmmacleod, nice catch! It was simple enough to change real fast, so I did that.

alurban:
thanks @duncanmmacleod! since it's assigned to you, I'll let you merge once build tests pass.

PR #984, opened 2018-12-03
Refactor TimeSeries.q_transform() to support eventgrams

### LINKED ISSUE #922: Refactor TimeSeries.q_transform() to support eventgrams
In the current iteration of `TimeSeries.q_transform()`, time-frequency tiles for several values of Q are computed and then ranged over, with the Q-plane containing the most significant tile interpolated over to produce a spectrogram, all under-the-hood. The intermediate step in this process (finding the Q-plane with the most significant tile) amounts to computing an "eventgram" for the input `TimeSeries`.

I propose refactoring the `q_transform()` method so that the eventgram is computed in a separate function (or method) and stored as an `EventTable`, with an option to pass it to `q_transform` as an argument. This would allow users to independently calculate eventgrams and Q-transform spectrograms, which is a requirement of the new Omega scan code: see [**here**](https://github.com/gwdetchar/gwdetchar/blob/master/bin/gwdetchar-omega#L239-L327).

Computing eventgrams is likely a substantial memory leak in the new Omega scan code, since it is effectively done four times per channel as currently written -- once independently and once during a call to `TimeSeries.q_transform()` for the whitened timeseries, and then both are repeated for the highpassed timeseries. The proposed refactoring would go a long way towards alleviating this.

cc @duncanmmacleod, @areeda, @scottcoughlin2014

### LINKED ISSUE #995: gwpy-plot crashes on long q-transform requests
The issue is caused by a fixed small time resolution value resulting in unreasonable intermediate image size.
Error is:
```  File "/Users/areeda/py_envs/gwpy/lib/python2.7/site-packages/gwpy/cli/qtransform.py", line 188, in get_spectrogram
    **self.qxfrm_args)
  File "/Users/areeda/py_envs/gwpy/lib/python2.7/site-packages/gwpy/timeseries/timeseries.py", line 1793, in q_transform
    new = Spectrogram(interp(out.times.value, freqs2 + fres/2.).T,
  File "/Users/areeda/py_envs/gwpy/lib/python2.7/site-packages/scipy/interpolate/interpolate.py", line 307, in __call__
    z = fitpack.bisplev(x, y, self.tck, dx, dy)
  File "/Users/areeda/py_envs/gwpy/lib/python2.7/site-packages/scipy/interpolate/_fitpack_impl.py", line 1045, in bisplev
    z, ier = _fitpack._bispev(tx, ty, c, kx, ky, x, y, dx, dy)
RuntimeError: Cannot produce output of size 300000x10327 (size too large)```

### PR DESCRIPTION
**Pull request by alurban**
This PR substantially refactors `TimeSeries.q_transform()` to support eventgrams, a feature ported over from [**`gwdetchar-omega`**](https://github.com/gwdetchar/gwdetchar/blob/master/bin/gwdetchar-omega#L166-L254). To simplify review, here is a summary of changes made:

**New Features**
* Introduce a new object class, `~gwpy.signal.qtransform.QGram`, to support eventgrams
* Introduce a new function, `~gwpy.signal.qtransform.q_scan()`, which identifies the time-frequency plane at fixed Q with the most significant tile within a given (optional) search window. This method returns an instance of `QGram` along with the estimated false alarm rate from white Gaussian noise.
* Include a method, `QGram.table()`, which converts a populated `QGram` object to an `EventTable` so that it can be rendered with matplotlib. Table columns are `'time'`, `'duration'`, `'frequency'`, `'bandwidth'`, and `'energy'`. The only argument is an SNR threshold, which selects for tiles with energy greater than or equal `snrthresh**2/2`.
* Introduce a `TimeSeries.q_gram` method to wrap around `QGram.table`.
* Add a `mismatch` argument for `q_scan()` and `TimeSeries.q_transform()`.
* Add a unit test for the `QGram` object under `gwpy/signal/tests/test_qtransform.py`.
* Add an example for the docs under `examples/signal/qscan.py`.

**Refactor Existing Features**
* Refactor spectrogram interpolation into `~gwpy.signal.qtransform.QGram.interpolate()`. This reduces the cyclomatic complexity of `TimeSeries.q_transform()` to a level acceptable to CodeClimate.

This fixes #922 and fixes #995.

cc @duncanmmacleod, @areeda, @scottcoughlin2014

### DIFF
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -34,12 +34,16 @@ from six.moves import xrange
 import numpy
 from numpy import fft as npfft
 
+from ..segments import Segment
 from ..timeseries import TimeSeries
 
 __author__ = 'Duncan Macleod <duncan.macleod@ligo.org>'
-__credits__ = 'Scott Coughlin <scott.coughlin@ligo.org>'
-__all__ = ['QTiling', 'QPlane', 'QTile']
+__credits__ = 'Scott Coughlin <scott.coughlin@ligo.org>, ' \
+              'Alex Urban <alexander.urban@ligo.org>'
+__all__ = ['QTiling', 'QPlane', 'QTile', 'QGram', 'q_scan']
 
+
+# -- object class definitions -------------------------------------------------
 
 class QObject(object):
     """Base class for Q-transform objects
@@ -89,12 +93,16 @@ class QTiling(QObject):
     ----------
     duration : `float`
         the duration of the data to be Q-transformed
+
     qrange : `tuple` of `float`
         `(low, high)` pair of Q extrema
+
     frange : `tuple` of `float`
         `(low, high)` pair of frequency extrema
+
     sampling : `float`
         sampling rate (in Hertz) of data to be Q-transformed
+
     mismatch : `float`
         maximum fractional mismatch between neighbouring tiles
     """
@@ -149,6 +157,44 @@ class QTiling(QObject):
             yield QPlane(q, self.frange, self.duration, self.sampling,
                          mismatch=self.mismatch)
 
+    def transform(self, fseries, **kwargs):
+        """Compute the time-frequency plane at fixed Q with the most
+        significant tile
+
+        Parameters
+        ----------
+        fseries : `~gwpy.timeseries.FrequencySeries`
+            the complex FFT of a time-series data set
+
+        **kwargs
+            other keyword arguments to pass to `QPlane.transform`
+
+        Returns
+        -------
+        out : `QGram`
+            signal energies over the time-frequency plane containing the most
+            significant tile
+
+        N : `int`
+            estimated number of statistically independent tiles
+
+        See Also
+        --------
+        QPlane.transform
+            compute the Q-transform over a single time-frequency plane
+        """
+        weight = 1 + numpy.log10(self.qrange[1]/self.qrange[0]) / numpy.sqrt(2)
+        nind, nplanes, peak, result = (0, 0, 0, None)
+        # identify the plane with the loudest tile
+        for plane in self:
+            nplanes += 1
+            nind += sum([1 + row.ntiles * row.deltam for row in plane])
+            result = plane.transform(fseries, **kwargs)
+            if result.peak['energy'] > peak:
+                out = result
+                peak = out.peak['energy']
+        return (out, nind * weight / nplanes)
+
 
 class QPlane(QBase):
     """Iterable representation of a Q-transform plane
@@ -160,12 +206,16 @@ class QPlane(QBase):
     ----------
     q : `float`
         the Q-value for this plane
+
     frange : `tuple` of `float`
         `(low, high)` range of frequencies for this plane
+
     duration : `float`
         the duration of the data to be Q-transformed
+
     sampling : `float`
         sampling rate (in Hertz) of data to be Q-transformed
+
     mismatch : `float`
         maximum fractional mismatch between neighbouring tiles
     """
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by alurban** [comment]
@duncanmmacleod, since #983 is merged, I've resolved WIP status here. This PR is now ready for review.

**Comment by duncanmmacleod** [comment]
@alurban, please rebase against `gwpy/gwpy@develop` to correct the appveyor failures.

**Comment by alurban** [comment]
@duncanmmacleod, this PR has been updated to pass build tests.

**Comment by duncanmmacleod** [comment]
@alurban, unfortunately, I'm not happy with this change - I can't 100% put my finger on why.

I do like the modularisation of the `TimeSeries.q_transform` method into the `gwpy.signal.qtransform` module, but i don't like the overloading of the `QPlane` object as a container for the _output_ of a Q-transform scan operation.

File: gwpy/timeseries/timeseries.py [hunk 1]
```diff
@@ -1749,11 +1749,122 @@ def notch(self, frequency, type='iir', filtfilt=True, **kwargs):
                                   type=type, **kwargs)
         return self.filter(*zpk, filtfilt=filtfilt)
 
+    def q_gram(self, qrange=(4, 64), frange=(0, numpy.inf), gps=None,
```
duncanmmacleod: What purpose does this method serve in the `TimeSeries` object? I think it might be better exposed simply via the `gwpy.signal.qtransform` module for expert users (like yourself) to use since the returned object doesn't really have much utility, except as a stepping-stone towards an `EventTable` or a `Spectrogram`.

gwpy/timeseries/timeseries.py
```diff
@@ -1805,13 +1920,17 @@ def q_transform(self, qrange=(4, 64), frange=(0, numpy.inf),
             highpass corner frequency (in Hz) of the FIR whitening filter,
             used only if `whiten` is not `False`, default: `None`
 
+        qgram : `~gwpy.signal.qtransform.QPlane`, optional
+            ready-made time-frequency plane (at fixed Q) over which to
+            interpolate, will override `whiten` if passed, default: `None`
+
```
duncanmmacleod: This keyword really doesn't belong here, since its inclusion removes all usage of the parent `TimeSeries` object in the method and reduces it to an alias of `qgram.interpolate`. If the user already has a `<qgram>` object, they can call the `interpolate` method themselves?

alurban:
@duncanmmacleod, I've implemented the idea you suggested to me on Slack. Under these new changes, there's now an object class called `QGram` and a utility called `~gwpy.signal.qtransform.q_scan()` that support eventgrams, which are wrapped around by `TimeSeries.q_transform()`. I've also updated gwdetchar/gwdetchar#206 to reflect this.

gwpy/signal/qtransform.py
```diff
@@ -149,6 +157,46 @@ def __iter__(self):
...
+            other keyword arguments to pass to `QPlane.transform`
+
+        Returns
+        -------
+        result : `QGram`
+            signal energies over the time-frequency plane containing the most
+            significant tile
+
+        N : `int`
+            estimated number of statistically significant tiles
```
duncanmmacleod: What's the utility of this number? Can this be computed externally (i.e. not included here)?

alurban: @duncanmmacleod, this number is needed to compute the false alarm rate (FAR) expected from white Gaussian noise, which is done in the [**`q_scan` utility**](https://github.com/alurban/gwpy/blob/eventgram/gwpy/signal/qtransform.py#L660). At least one downstream consumer, `gwdetchar-omega`, uses this metric to filter out quiet or insignificant channels, and other consumers (e.g., commissioners writing data quality checks) may also find it useful.

The number of statistically independent tiles is really a property of the search over a `QTiling`, i.e., it's calculated as a sum over all time-frequency planes in the tiling used in a search. So it needs to be made available as part of the `QTiling` object somehow, and this seemed to me like a natural way to do that.

alurban:
thanks @duncanmmacleod, I can definitely do that, do you want it in the docstring for `q_scan` or as a separate tutorial under `examples/`?

duncanmmacleod:
@alurban, I was meaning under `examples/`.

alurban:
@duncanmmacleod, I've added an example under `examples/signal/qscan.py` that computes a raw Q-scan around GW170817.

alurban:
@duncanmmacleod, I've updated this PR (and the example plot) to incorporate a new method, `TimeSeries.q_gram()`, which wraps around `QGram.table()`. This doesn't handle auto-whitening, but I personally am 100% fine with that, not least because the worked example warns the user to whiten beforehand.

gwpy/signal/qtransform.py
```diff
@@ -377,7 +435,227 @@ def transform(self, fseries, norm=True, epoch=None):
...
+        """
+        from ..table import EventTable
+        # get plane properties
+        freqs = self.plane.frequencies
+        bws = 2 * (freqs - self.plane.farray)
+        # collect table data as a recarray
+        names = ('time', 'frequency', 'duration', 'bandwidth', 'energy')
+        rec = numpy.recarray((0,), names=names, formats=['f8'] * len(names))
+        for f, bw, row in zip(freqs, bws, self.energies):
+            ind, = (row.value >= snrthresh ** 2 / 2).nonzero()
```
duncanmmacleod: @alurban, I'm concerned here that if `snrthresh` is given as an `int`, then `snrthresh ** 2 / 2` is entirely composed of integers and might not do what you think it does for python < 3: e.g.

```python
>>> 5 ** 2 / 2
12
>>> 5 ** 2 / 2.
12.5
```

To be safe, can you please add a float decimal point:

```suggestion
            ind, = (row.value >= snrthresh ** 2. / 2.).nonzero()
```

PR #940, opened 2018-10-19
Update frequencyseries.py

dainmiller:
Addresses #932

PR #986, opened 2018-12-06
Fix interpolation domain in FrequencySeries.interpolate()

### LINKED ISSUE #932: Bug in FrequencySeries.interpolate
The method `FrequencySeries.interpolate()` is currently always setting its output `f0` to 0 Hz. This causes problems if the original `FrequencySeries` begins above 0 Hz, as the interpolated `FrequencySeries` will start at 0 Hz but have the same bandwidth as the original, so its first several samples will be zero.

To fix this, the output `FrequencySeries` should (1) interpolate in the correct frequency range and (2) inherit the `f0` attribute from the original `FrequencySeries`.

alurban:
This PR fixes a bug in `FrequencySeries.interpolate()` so that the output of that method is interpolated properly only over the domain of the input. As a consequence, the `left` and `right` arguments have been removed, since they are redundant.

A more complete unit test for `FrequencySeries.interpolate()` is also added.

This fixes #932. (Note, the first attempt at doing this was in #940.)

### DIFF
--- a/gwpy/frequencyseries/frequencyseries.py
+++ b/gwpy/frequencyseries/frequencyseries.py
@@ -270,7 +270,7 @@ class FrequencySeries(Series):
         """
         return self.filter(zeros, poles, gain, analog=analog)
 
-    def interpolate(self, df, left=0, right=0):
+    def interpolate(self, df):
         """Interpolate this `FrequencySeries` to a new resolution.
 
         Parameters
@@ -279,12 +279,6 @@ class FrequencySeries(Series):
             desired frequency resolution of the interpolated `FrequencySeries`,
             in Hz
 
-        left : `float`
-            value to return for f < flow, default: 0
-
-        right : `float`
-            value to return for f > fhigh, default: 0
-
         Returns
         -------
         out : `FrequencySeries`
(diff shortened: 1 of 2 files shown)


gwpy/frequencyseries/frequencyseries.py
```diff
@@ -279,27 +279,23 @@ def interpolate(self, df, left=0, right=0):
...
-
         Returns
         -------
         out : `FrequencySeries`
             the interpolated version of the input `FrequencySeries`
 
         See Also
         --------
-        :func:`numpy.interp`
+        :func:`~numpy.interp`
```
duncanmmacleod: Can you change this to be:

```
        See Also
        ---------
        numpy.interp
```
(i.e. remove the ticks and tildes, please)?

alurban: @duncanmmacleod, this is done, I'm actually not sure why I put those there in the first place 😱

gwpy/frequencyseries/tests/test_frequencyseries.py
```diff
@@ -168,14 +168,16 @@ def test_interpolate(self):
...
+        fseries = FrequencySeries(numpy.ones(nsamp), f0=1, df=df, unit='')
 
         # create an interpolated FrequencySeries
         newf = fseries.interpolate(df/2.)
 
         # check that the interpolated series is what was expected
         assert newf.unit == fseries.unit
         assert newf.size == 2*(fseries.size - 1) + 1
+        assert newf.df.value == df / 2.
+        assert newf.f0.value == fseries.f0.value
```
duncanmmacleod: Can we just compare the quantities, to make sure the units are mangled as well:

```python
        assert new.df == fseries.df / 2.
        assert new.f0 == fseries.f0
```
?

PR #996, opened 2018-12-15
Deal with longer Q-transform requests, fix problem with Q-tranform vector size mismatch

areeda:
Addresses issue #995 
* tres (time resolution) now determined image size and search duration
* timeseries.q_transform modifiedto correct nx mismatch with xout length which became apparent only when tres does not divide a second into integer pieces
* Also add verbose messages (level 3) for q-transform parameters.


gwpy/cli/qtransform.py
```diff
@@ -112,12 +112,23 @@ def _finalize_arguments(self, args):
...
-        # ennsure we have enough data for filter settling
+        # ensure we have enough data for filter settling
         max_plot = max(args.plot)
         search = max(search, max_plot * 2 + 6)
+        args.search = search
         self.log(3, "Search window: {0:.0f} sec, max plot window {1:.0f}".
                  format(search, max_plot))
 
+        # make sure we don't create too big interpolations
+        xpix = 1200.
```
alurban: @areeda, it might reduce cognitive complexity to instead do the check like this:

```python
try:
    m = re.match('(\d+)x(\d+)', args.geometry)
    xpix = float(m.group(1))
except:
    xpix = 1200.
```
This way, there's only one decision point as opposed to two.

areeda: I really don't get 'cognitive complexity' but thank for fixing that.

areeda:
I resolved the bare except complaint by trapping only AttributeError
The other one is just silly and has been here forever.

gwpy/cli/qtransform.py
```diff
@@ -112,12 +112,24 @@ def _finalize_arguments(self, args):
...
+        # ensure we have enough data for filter settling
         max_plot = max(args.plot)
         search = max(search, max_plot * 2 + 6)
+        args.search = search
         self.log(3, "Search window: {0:.0f} sec, max plot window {1:.0f}".
                  format(search, max_plot))
 
+        # make sure we don't create too big interpolations
+        try:
+            m = re.match('(\\d+)x(\\d+)', args.geometry)
```
duncanmmacleod: @areeda, please move this statement outside of the `try/except` block, since we don't want to catch `AttributeError` here.

gwpy/timeseries/timeseries.py
```diff
@@ -1931,14 +1931,15 @@ def q_transform(self, qrange=(4, 64), frange=(0, numpy.inf),
 
         # build regular Spectrogram from peak-Q data by interpolating each
         # (Q, frequency) `TimeSeries` to have the same time resolution
-        nx = int(abs(Segment(*outseg)) / tres)
+        xout = numpy.arange(outseg[0], outseg[1], tres)
+        nx = xout.size
+
```
duncanmmacleod: What's the motivation for this change? Does this not give the same answer as before?

areeda:
I  don't like this solution.tres and search search length should be set for each scannot once per runn

areeda:
Reopening with a better fix. This one adjusts both the amount of data processed and time resloution based on the duration of the image produced.  

It's interesting to look at long duration q-scans.  I think we might miss short phenomena.  I've done 30 min long ones.

gwpy/cli/qtransform.py
areeda: This is now back to if statements which will triger a complexity notice but I think it's easier to understand

gwpy/timeseries/timeseries.py
areeda: Sorry, I thought I replied to this already.
It turns out that when tres is calculated and is not an integer multiple/divisor of the sample frequency that we can end up with round  off errors that result in a difference of 1.

areeda:
I should point out that during ER 13 there were multiple requests that would have been better handled by this fix for example:
https://ldvw.ligo.caltech.edu/ldvw/view?act=getImg&imgId=206905
https://ldvw.ligo.caltech.edu/ldvw/view?act=getImg&imgId=206292
https://ldvw.ligo.caltech.edu/ldvw/view?act=getImg&imgId=206280

gwpy/timeseries/timeseries.py
alurban: @areeda, I've also addressed this as part of [**#984**](https://github.com/gwpy/gwpy/blob/7a3ab3183d138e6c1442f11380ffe56d1363b630/gwpy/signal/qtransform.py#L526-L530) which will end up superseding this since that PR refactors spectrogram interpolation into `gwpy.signal.qtransform`.

duncanmmacleod:
@alurban, @areeda, this conflicts with #922 (merged), can you please resolve this amongst yourselves?

duncanmmacleod:
@areeda, @alurban, is there an update on this?

alurban:
@duncanmmacleod, last week I sent @areeda a slack message suggesting how best to resolve the merge conflict

areeda:
Cancel this pull request new one coming from a different clean branch.

PR #1001, opened 2018-12-18
Error on interactive prompts in non-interactive sessions for kinit()

### LINKED ISSUE #838: Enhancement: timeseries.fetch() option to fail on no TGT
I've had the problem come up that timeseries.fetch() will hang  when called in the background (by LDVW) with an expired TGT. This leaves zombie processes.

 I can think of 2 approaches that may be easy to implement.

1.  A new keyword argument that requests the failure rather than an interactive request for username and password.

2.  A timeout, can be long, for the user to enter that information.

### LINKED ISSUE #864: TimeSeries{,Dict} objects should include method for GPS start/stop times
It would be great if both the TimeSeries and TimeSeriesDict objects included methods that return the GPS start and stop of the segment.  A tuple of the form [start, stop) would be good.

duncanmmacleod:
This PR fixes #838 by patching `gwpy.io.kerberos.kinit` to raise an exception whenever it is about to try and prompt for a username or password in a session that won't allow a response from the user.

I also took an opportunity to clean up the test suite for the containing `gwpy.io.kerberos` module.

### DIFF
--- a/gwpy/io/kerberos.py
+++ b/gwpy/io/kerberos.py
@@ -125,12 +125,14 @@ def kinit(username=None, password=None, realm=None, exe=None, keytab=None,
             else:
                 keytab = None
 
-    # set verbose if prompting for input
-    if verbose is None and (
-        username is None or
-        (not keytab and password is None)
+    # refuse to prompt if we can't get an answer
+    if not sys.stdout.isatty() and (
+            username is None or
+            (not keytab and password is None)
     ):
-        verbose = True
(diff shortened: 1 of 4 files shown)


duncanmmacleod:
@areeda, can you take a look through the [changes to `gwpy.io.kerberos`](/gwpy/gwpy/pull/1001/files#diff-8a0c884b0093a1aaeaa938ae58ed1cfa) to make sure that this corrects any issues you were seeing as reported in #838?

PR #1007, opened 2019-01-03
Manually install openblas for scipy on macports

duncanmmacleod:
This PR attempts to fix the current Travis CI failures [[example](https://travis-ci.com/gwpy/gwpy/jobs/167964402)] by manually installing `OpenBLAS` with macports before building GWpy on that platform. This should allow scipy-1.2.0 to then build against those libraries, rather than the system ones that seem to be missing symbols.

### DIFF
--- a/ci/install-macports.sh
+++ b/ci/install-macports.sh
@@ -81,6 +81,9 @@ wvbpid=$!
 disown
 set -x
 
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
Progress towards an upstream fix can be followed here: https://trac.macports.org/ticket/57829

PR #1014, opened 2019-01-16
Added tconvert-like command-line interface for gwpy.time

duncanmmacleod:
This PR adds a command-line interface to `gwpy.time` to replicate `tconvert` functionality, used as follows:

```shell
$ python -m gwpy.time 1161887657
2016-10-30 18:34:00.000000 UTC
```

This also improves the functionality of `gwpy.time._str_to_datetime` by optionally using [`maya`](https://pypi.org/project/maya/) to parse datetime strings, its a little better than `dateutil`.

### DIFF
--- a/gwpy/time/_tconvert.py
+++ b/gwpy/time/_tconvert.py
@@ -23,6 +23,7 @@ Peter Shawhan.
 """
 
 import datetime
+import warnings
 from decimal import Decimal
 
 from six import string_types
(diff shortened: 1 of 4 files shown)


gwpy/time/_tconvert.py
```diff
@@ -233,12 +233,16 @@ def _str_to_datetime(datestr):
...
     except KeyError:  # any other string
-        try:
-            return dateparser.parse(datestr)
-        except (ValueError, TypeError) as exc:
-            exc.args = ("Cannot parse date string {0!r}: {1}".format(
-                datestr, exc.args[0]),)
-            raise
+        try:  # use maya
+            import maya
+            return maya.when(datestr).datetime()
```
alurban: @duncanmmacleod, this is a potentially awesome general-purpose feature, maybe we should move it upstream to `from_gps`?

duncanmmacleod: @alurban, I think I may have misspoke when we talked on Slack. This code is as 'upstream' as it gets, i was referring to the printing of timezone-aware strings as the output of `python -m gwpy.time`, however, since `gwpy.time` (the module) just returns `datetime.datetime` objects users can manually transition to other timezones if they like.

PR #1028, opened 2019-02-07
Added support for median average from scipy

duncanmmacleod:
This PR adds support for calling `scipy.signal.welch` with `average='median'`, which was introduced in scipy-1.2.0.


duncanmmacleod:
@alurban, this PR adds support for `method='median'`, which was previously supported but would map to `method='pycbc_median'` or `method='lal_median'`. My plan is to remove support for LAL and PyCBC FFT routines completely and just support scipy.

PR #1031, opened 2019-02-12
Use dqsegdb2 for segment queries

duncanmmacleod:
This PR modifies the `gwpy.segments` sub-package to use the [`dqsegdb2`](https://dqsegdb2.readthedocs.io) library for segment queries, instead of `dqsegdb`.

`dqsegdb2` is added as a first-class dependency, so segment queries should now be considered a standard part of GWpy's capability.


duncanmmacleod:
This build is failing for the same reason that #1032 failed to being with. Once that PR is merged we can rebase and the tests will run more completely.

PR #1038, opened 2019-02-12
Q-transform fix min search length

areeda:
I had the sign  of protract wrong. Plus align reading pad with calculation padding


duncanmmacleod:
Can we change that test to just use a regular expression, or similar, so that the actual numbers don't matter?

areeda:
I'll look at the test.

duncanmmacleod:
@areeda, have you been able to make any progress on this?

PR #1042, opened 2019-02-18
Improved EventTable I/O with decorators

duncanmmacleod:
This PR attempts to improve the `EventTable` I/O system, mainly to make the keyword arguments more consistent across formats.

All readers (including those inherited from astropy) now accept the `columns` and `selection` keywords, mainly thanks to the new `gwpy.table.io.utils.decorate_registered_reader` function.

This is technically backwards incompatible since it registers the 'hdf5' reader for the `EventTable`, meaning PyCBC HDF5 files cannot be distinguished from normal HDF5 files, which in turns means the `format` keyword now needs to be explicitly given when reading either PyCBC Live files.

### DIFF
--- a/gwpy/table/tests/test_table.py
+++ b/gwpy/table/tests/test_table.py
@@ -555,7 +555,8 @@ class TestEventTable(TestTable):
             utils.assert_table_equal(table, t2)
 
             # assert processed colums works
-            t2 = self.TABLE.read(fp,
+            t2 = self.TABLE.read(
+                fp,
                 format="hdf5.pycbc_live",
                 ifo="X1",
                 columns=["mchirp", "new_snr"],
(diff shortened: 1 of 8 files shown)


gwpy/table/tests/test_table.py
```diff
@@ -505,15 +555,27 @@ def test_read_pycbc_live(self):
             utils.assert_table_equal(table, t2)
 
             # assert processed colums works
-            t2 = self.TABLE.read(fp, ifo='X1', columns=['mchirp', 'new_snr'])
+            t2 = self.TABLE.read(fp,
```
alurban: @duncanmmacleod, PEP-8 is complaining about this line, maybe to pacify it you can make the following change?

```suggestion
            t2 = self.TABLE.read(
                fp,
```

PR #1046, opened 2019-02-19
Updated copyright years for all files under our license

duncanmmacleod:
This PR modifies essentially every file in the repo to update the years of copyright, which should match any year there has been a release: 2014-2019.

I also added a utility to automatically update the copyright statements, and documented that this should be run for each release, it should have no impact when run for a year that is already included in the copyright.

@alurban, @areeda, @scottcoughlin2014, you are listed as copyright holders for one or more files, so I need explicit approval from each of you.


areeda:
Files naming Areeda have been checked. Others were scanned. LGTM

PR #1047, opened 2019-02-20
Port LIGO_LW I/O to python-ligo-lw

duncanmmacleod:
This PR implements a backwards-compatible update to use the new `python-ligo-lw` library for `LIGO_LW` I/O, but with fallbacks to `glue.ligolw` when reading old files.

When reading there is now a try/except that should try and read with the new library, and if that fails, try and use the old library.

When writing:
- ...tables, the columns/datatypes in the table define the format version, so both libraries are tried (new first)
- ...segments, there is a new `ilwdchar_compat` keyword that currently defaults to `None`, which presents a warning that it will use `True` (old format), but will be switched to `False` for the 1.0.0 release (at which time the warning will be removed).

The upshot of this is that GWpy should now be able to seamlessly read LIGO_LW files written in either new or old formats.

### DIFF
--- a/docs/segments/io.rst
+++ b/docs/segments/io.rst
@@ -21,11 +21,11 @@ The :meth:`read` and :meth:`write` methods take different arguments and keywords
 ``LIGO_LW`` XML
 ===============
 
-**Additional dependencies:** :mod:`glue.ligolw`
+**Additional dependencies:** |python-ligo-lw|_
 
 The LIGO Scientific Collaboration uses a custom scheme of XML in which to
-store tabular data, called the ``LIGO_LW`` scheme.
-Complementing the scheme is a python library - :mod:`glue.ligolw` - which
+store tabular data, called ``LIGO_LW``.
+Complementing the scheme is a python library - |python-ligo-lw|_ - which
 allows users to read and write all of the different types of tabular data
 produced by gravitational-wave searches.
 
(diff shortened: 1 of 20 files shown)


duncanmmacleod:
This is probably going to have to wait until `python-ligo-lw=1.5.2` has been made available in the lscsoft repos for RPM/Deb.

duncanmmacleod:
@alurban, are you able to review this change, please?

docs/segments/io.rst
```diff
@@ -82,6 +82,18 @@ Extra attributes can be written to the tables via the ``attrs={}`` keyword, all
 
     >>> f.write('new-table.xml', append=True, overwrite=True, attrs={'process_id': 0})
 
+.. note::
+
+   The |python-ligo-lw| library reads and writes files using an updated
+   version of the format compared to :mod:`glue.ligolw` used to. GWpy
```
alurban: @duncanmmacleod, this is definitely pedantic, but I think maybe this passage can be re-worded?

```suggestion
   version of the ``LIGO_LW`` format compared to :mod:`glue.ligolw`. GWpy
```

docs/table/io.rst
```diff
@@ -151,6 +153,18 @@ To replace an existing table of the given type in an existing file, while preser
 
     >>> t.write('new-table.xml', format='ligolw', tablename='sngl_burst', append=True, overwrite=True)
 
+.. note::
+
+   The |python-ligo-lw| library reads and writes files using an updated
+   version of the format compared to :mod:`glue.ligolw` used to. GWpy
```
alurban: @duncanmmacleod, same comment here:

```suggestion
   version of the ``LIGO_LW`` format compared to :mod:`glue.ligolw`. GWpy
```

gwpy/io/ligolw.py
```diff
@@ -23,25 +23,89 @@
...
+        use_compat = kwargs.pop(
+            "ilwdchar_compat",
+            any(map(
+                _is_glue_ligolw_object,
+                list(args) + list(kwargs.values()),
+            )),
+        )
+        try:
+            if use_compat:
+                builtins.__import__ = _ligolw_compat_import
```
alurban: @duncanmmacleod, I don't have any real notes here, except to say that I've never seen anything like this before and I'm learning new things about what python does under-the-hood by reviewing your code, which is a lot of fun

duncanmmacleod: It's quite a neat solution to the problem of supporting two different upstream APIs with a consistent single downstream API.

PR #1049, opened 2019-02-20
Upgrade gwdatafind to first-class dependency

duncanmmacleod:
This PR upgrades `gwdatafind` to a first-class dependency. This was the case in practice already since `dqsegdb2` is a dependency and depends in turn on `gwdatafind`, but this allows us to clean up some imports.

I also took this chance to clean up the dependency lists as they exist across various files, now all alphabetically ordered, and rechecked.

### DIFF
--- a/gwpy/io/datafind.py
+++ b/gwpy/io/datafind.py
@@ -43,6 +43,8 @@ from functools import wraps
 
 from six.moves.urllib.parse import urlparse
 
+from gwdatafind import connect
+
 from ..segments import (Segment, SegmentList)
 from ..time import to_gps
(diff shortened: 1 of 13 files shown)


gwpy/io/datafind.py
```diff
@@ -309,7 +311,6 @@ def on_tape(*files):
 
 def _choose_connection(**datafind_kw):
     if os.getenv('LIGO_DATAFIND_SERVER') or datafind_kw.get('host'):
-        from gwdatafind import connect
         return connect(**datafind_kw)
```
alurban: @duncanmmacleod, methinks this line will have to change to the following?

```suggestion
         return gwdatafind.connect(**datafind_kw)
```

PR #1051, opened 2019-02-21
Make interpolated Q-transform spectrograms single precision

### LINKED ISSUE #1050: Defaults on q_transform method result in large-memory objects
If I use the `TimeSeries.q_transform()` method with the default settings, the resulting arrays are sometimes taking up several GB of memory. I was able to push this down by tuning the arguments, would it be possible to tweak the defaults so that the average user doesn't run into memory issues if they're making several Q-transforms?

alurban:
This pull request ensures that interpolated Q-transform spectrograms are populated with single-precision floats, to save memory. It also slightly expands unit tests for the Q-transform, and fixes one of the documentation examples to use `outseg` rather than `crop`.

This fixes #1050.

cc @duncanmmacleod, @tjma12

### DIFF
--- a/examples/timeseries/qscan.py
+++ b/examples/timeseries/qscan.py
@@ -42,12 +42,16 @@ from gwpy.timeseries import TimeSeries
 data = TimeSeries.fetch_open_data('H1', 1126259446, 1126259478)
 
 # Next, we generate the `~TimeSeries.q_transform` of these data:
-qspecgram = data.q_transform()
+qspecgram = data.q_transform(outseg=(1126259462.2, 1126259462.5))
 
-# Now, we can plot the resulting `~gwpy.spectrogram.Spectrogram`, focusing on a
-# specific window around the interesting time
+# .. note::
+#    We can save memory by focusing on a specific window around the
+#    interesting time. The ``outseg`` keyword argument returns a `Spectrogram`
+#    that is only as long as we need it to be.
 
-plot = qspecgram.crop(1126259462.2, 1126259462.5).plot(figsize=[8, 4])
+# Now, we can plot the resulting `~gwpy.spectrogram.Spectrogram`:
+
+plot = qspecgram.plot(figsize=[8, 4])
 ax = plot.gca()
 ax.set_xscale('seconds')
 ax.set_yscale('log')
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -518,7 +518,7 @@ class QGram(object):
 
         Notes
         -----
-        This method will return a `Spectrogram` of dtype ``float32``
+        This method will return a `Spectrogram` of dtype ``float32`` if
         ``norm`` is given, and ``float64`` otherwise.
 
         To optimize plot rendering with `~matplotlib.axes.Axes.pcolormesh`,
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -435,11 +435,12 @@ class QTile(QBase):
         if norm:
             norm = norm.lower() if isinstance(norm, string_types) else norm
             if norm in (True, 'median'):
-                return energy / energy.median()
+                narray = energy / energy.median()
             elif norm in ('mean',):
-                return energy / energy.mean()
+                narray = energy / energy.mean()
             else:
                 raise ValueError("Invalid normalisation %r" % norm)
+            return narray.astype("float32", casting="same_kind", copy=False)
         return energy
 
 
@@ -478,14 +479,15 @@ class QGram(object):
                 })
         return peak
 
-    def interpolate(self, tres=.001, fres="<default>",
-                    logf=False, outseg=None):
+    def interpolate(self, tres="<default>", fres="<default>", logf=False,
+                    outseg=None):
         """Interpolate this `QGram` over a regularly-gridded spectrogram
 
         Parameters
         ----------
         tres : `float`, optional
-            desired time resolution (seconds) of output `Spectrogram`
+            desired time resolution (seconds) of output `Spectrogram`,
+            default is `abs(outseg) / 1000.`
 
         fres : `float`, `int`, `None`, optional
             desired frequency resolution (Hertz) of output `Spectrogram`,
@@ -498,7 +500,8 @@ class QGram(object):
             log-sampled frequencies in the output `Spectrogram`
 
         outseg : `~gwpy.segments.Segment`, optional
-            GPS `[start, stop)` segment for output `Spectrogram`
+            GPS `[start, stop)` segment for output `Spectrogram`,
+            default is the full duration of the input
 
         Returns
         -------
@@ -515,6 +518,9 @@ class QGram(object):
 
         Notes
         -----
+        This method will return a `Spectrogram` of dtype ``float32``
+        ``norm`` is given, and ``float64`` otherwise.
+
         To optimize plot rendering with `~matplotlib.axes.Axes.pcolormesh`,
         the output `~gwpy.spectrogram.Spectrogram` can be given a log-sampled
         frequency axis by passing `logf=True` at runtime. The `fres` argument
@@ -532,6 +538,8 @@ class QGram(object):
         dtype = self.energies[0].dtype
         # build regular Spectrogram from peak-Q data by interpolating each
         # (Q, frequency) `TimeSeries` to have the same time resolution
+        if tres == "<default>":
+            tres = abs(Segment(outseg)) / 1000.
         xout = numpy.arange(*outseg, step=tres)
         nx = xout.size
         ny = frequencies.size
@@ -544,7 +552,8 @@ class QGram(object):
             xrow = numpy.arange(row.x0.value, (row.x0 + row.duration).value,
                                 row.dx.value)
             interp = InterpolatedUnivariateSpline(xrow, row.value)
-            out[:, i] = interp(xout)
+            out[:, i] = interp(xout).astype(dtype, casting="same_kind",
+                                            copy=False)
         if fres is None:
             return out
         # interpolate the spectrogram to increase its frequency resolution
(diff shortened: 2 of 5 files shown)


duncanmmacleod:
@alurban, have you been able to perform basic profiling to check thay this does what we want it to?

duncanmmacleod:
@alurban, I've now looked at the code (responded quickly on mobile earlier); I think we can actually do the type casting much further upstream, I'm looking into it now. Can you also add a sanity unit test that asserts the dtype is what we need it be?

duncanmmacleod:
Unforunately, scipy's interpolate is determined to return float64, which you probably discovered already, but here's my attempt to pass around float32 all the time **only when normalising the q_transform output**:

```patch
diff --git a/gwpy/signal/qtransform.py b/gwpy/signal/qtransform.py
index 6c8531de..e1f49234 100644
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -435,11 +435,12 @@ class QTile(QBase):
         if norm:
             norm = norm.lower() if isinstance(norm, string_types) else norm
             if norm in (True, 'median'):
-                return energy / energy.median()
+                narray = energy / energy.median()
             elif norm in ('mean',):
-                return energy / energy.mean()
+                narray = energy / energy.mean()
             else:
                 raise ValueError("Invalid normalisation %r" % norm)
+            return narray.astype("float32", casting="same_kind", copy=False)
         return energy


@@ -542,9 +543,10 @@ class QGram(object):
         # interpolate rows
         for i, row in enumerate(self.energies):
             xrow = numpy.arange(row.x0.value, (row.x0 + row.duration).value,
-                                row.dx.value)
+                                row.dx.value, dtype=dtype)
             interp = InterpolatedUnivariateSpline(xrow, row.value)
-            out[:, i] = interp(xout)
+            out[:, i] = interp(xout).astype(dtype, casting="same_kind",
+                                            copy=False)
         if fres is None:
             return out
         # interpolate the spectrogram to increase its frequency resolution
@@ -555,7 +557,8 @@ class QGram(object):
             if fres == "<default>":
                 fres = .5
             outfreq = numpy.arange(
-                self.plane.frange[0], self.plane.frange[1], fres)
+                self.plane.frange[0], self.plane.frange[1], fres,
+                dtype=dtype)
         else:
             if fres == "<default>":
                 fres = 500
@@ -563,9 +566,13 @@ class QGram(object):
             # but numpy-1.12.0 introduced the function `~numpy.geomspace`
             logfmin = numpy.log10(self.plane.frange[0])
             logfmax = numpy.log10(self.plane.frange[1])
-            outfreq = numpy.logspace(logfmin, logfmax, num=int(fres))
-        new = type(out)(interp(xout, outfreq).T,
-                        t0=outseg[0], dt=tres, frequencies=outfreq)
+            outfreq = numpy.logspace(logfmin, logfmax, num=int(fres),
+                                     dtype=dtype)
+        new = type(out)(
+            interp(xout, outfreq).T.astype(
+                dtype, casting="same_kind", copy=False),
+            t0=outseg[0], dt=tres, frequencies=outfreq,
+        )
         new.q = self.plane.q
         return new

diff --git a/gwpy/signal/tests/test_qtransform.py b/gwpy/signal/tests/test_qtransform.py
index 75ed4525..c6d670b7 100644
--- a/gwpy/signal/tests/test_qtransform.py
+++ b/gwpy/signal/tests/test_qtransform.py
@@ -60,6 +60,7 @@ def test_q_scan():
     # test spectrogram output
     assert ts_qspecgram.q == QSPECGRAM.q
     assert ts_qspecgram.shape == QSPECGRAM.shape
+    assert ts_qspecgram.dtype == numpy.dtype('float32')
     nptest.assert_allclose(ts_qspecgram.value, QSPECGRAM.value)

```

The test is trivial, but obviously should be extended to check that the return when _not_ normalising is `float64`.

alurban:
@duncanmmacleod, I've taken your suggestion and refined our unit tests to reflect the single-precision floats. I also added a unit test for the `norm=False` use case to check that it returns a double-precision spectrogram. Later tonight I'll do some profiling to make sure this actually addresses @tjma12's issue.

alurban:
@duncanmmacleod, the circleci build tests fail because they can't find various packages in the LSC software stack (perhaps because of today's authentication issues?), while one job in the Travis CI build is failing because `pycbc` is attempting to import `glue.segments` which has been replaced by `ligo.segments`. This is the same problem that affected #1052.

alurban:
@duncanmmacleod, I've compared a Q-scan of 200 seconds of 16 kHz data using the default arguments to `TimeSeries.q_transform()` with and without this PR's changes. This comparison is based on the following code snippet:

```python
# imports

# get 200 seconds of data
start = 1234665339
end = 1234665539
hoft = TimeSeries.get('L1:GDS-CALIB_STRAIN', start=start, end=end,
                      nproc=8, verbose=True)
qscan = hoft.q_transform()

# get size of scan in memory
print(qscan.nbytes / 1e9)
```

Before these changes, `qscan` has `dtype('float64')` and is 16.52 GB, while after these changes it has `dtype('float32')` and is 8.26 GB. So, we have saved exactly a factor of 2 in memory with the new default settings.

I can get this down to 400 MB by passing `logf=True`. **However**, if this is made the default, users wouldn't be able to call `qscan.plot()` or `qscan.imshow()` because of the log-sampled frequency axis (they would have to use `pcolormesh`). We might be able to hack together some axis settings that support `logf=True`, but I haven't tried it.

duncanmmacleod:
@alurban, I have worked around the pycbc failure in #1053, please rebase.

duncanmmacleod:
I think a factor of 2 in memory for no loss of utility is great, thanks for putting this together. The rest of the memory problems are, in principle, user error, so the thing we should do is beef up the documentation to promote proper usage of the keyword arguments. Something like this might be good:

```diff
diff --git a/gwpy/timeseries/timeseries.py b/gwpy/timeseries/timeseries.py
index cdc51c76..d35792dc 100644
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1853,6 +1853,12 @@ class TimeSeries(TimeSeriesBase):
         """Scan a `TimeSeries` using the multi-Q transform and return an
         interpolated high-resolution spectrogram

+        By default, this method returns a high-resolution spectrogram in
+        both time and frequency, which can result in a large memory
+        footprint.  If you know that you only need a subset of the output
+        for, say, a figure, consider using ``outseg`` and the other
+        keyword arguments to restrict the size of the returned data.
+
         Parameters
         ----------
         qrange : `tuple` of `float`, optional
```

We could also consider using a dynamic time resolution, where instead of the default `tres=0.01`, we have `tres=<data duration>/1000.`.
This will have the double impact of reducing memory for people who do really want a long-duration Q-transform, and also forcing those people who then `crop()` to a .5-second span to notice that they have done something wrong.
1000 is suggested only because it is roughly the number of pixels most people end up with along the X-axis of a matplotlib figure.

alurban:
@duncanmmacleod, I _really_ liked your `tres` suggestion, so I've set the default `tres` to be `abs(outseg) / 1000` (where `outseg` is the full duration of input data by default). Not only is this good because of the points you mentioned, but also, it _fixes_ the size in memory of interpolated spectrograms to be 41.3 MB irrespective of what the input is. Users who want to can use `logf=True` to get that down to just 2 MB, for a 500 x 1000 pixel grid.

I also updated the unit tests and docstring for `TimeSeries.q_transform()`.

alurban:
@duncanmmacleod, I have updated one of the docs examples to use the `outseg` kwarg instead of `crop`, this PR is now ready to go.

examples/timeseries/qscan.py
```diff
@@ -42,12 +42,17 @@
 data = TimeSeries.fetch_open_data('H1', 1126259446, 1126259478)
 
 # Next, we generate the `~TimeSeries.q_transform` of these data:
-qspecgram = data.q_transform()
+outseg = (1126259462.2, 1126259462.5)
+qspecgram = data.q_transform(outseg=outseg)
```
duncanmmacleod: @alurban, can you merge these into a single line?

gwpy/signal/qtransform.py
```diff
@@ -532,6 +533,8 @@ def interpolate(self, tres=.001, fres="<default>",
         dtype = self.energies[0].dtype
         # build regular Spectrogram from peak-Q data by interpolating each
         # (Q, frequency) `TimeSeries` to have the same time resolution
+        if tres == "<default>":
+            tres = abs(Segment(outseg)) / 1000
```
duncanmmacleod:
```suggestion
            tres = abs(Segment(outseg)) / 1000.
```

gwpy/timeseries/timeseries.py
```diff
@@ -1840,7 +1840,7 @@ def q_transform(self,
                     frange=qtransform.DEFAULT_FRANGE,
                     gps=None,
                     search=.5,
-                    tres=.001,
+                    tres="<default>",
```
duncanmmacleod: Please add documentation of the defaults in the docstring, similar to what is there for `fres`.

examples/timeseries/qscan.py
alurban: @duncanmmacleod, aye, this is done in the latest commit.

gwpy/signal/qtransform.py
alurban: @duncanmmacleod, this is done in the latest commit.

gwpy/timeseries/timeseries.py
alurban: @duncanmmacleod, this is done in the latest commit. I also updated the documentation of `outseg` to give its default (the full duration of the input `TimeSeries`).

alurban:
@duncanmmacleod, I've documented the `float32` magic, that was a good catch!

gwpy/signal/qtransform.py
```diff
@@ -515,6 +518,9 @@ def interpolate(self, tres=.001, fres="<default>",
 
         Notes
         -----
+        This method will return a `Spectrogram` of dtype ``float32``
```
duncanmmacleod:
```suggestion
        This method will return a `Spectrogram` of dtype ``float32`` if
```

alurban: thanks @duncanmmacleod, I've committed this change.

duncanmmacleod:
Thanks @alurban.

PR #1052, opened 2019-02-21
Added basic gating as TimeSeries method

tjma12:
This PR adds gating as a `TimeSeries` class method. The method whitens the time series, marks all points above a threshold, clusters them within a time window, and applies inverse Planck windows to zero out the data. The **unwhitened**, gated time series is returned.

An example can be found here (behind LIGO.org auth): https://ldas-jobs.ligo.caltech.edu/~thomas.massinger/notebook/demonstrate-gwpy-gating.html

Whitening arguments can be passed to the underlying whitening used to locate gating points.

### DIFF
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -58,45 +58,6 @@ def _fft_length_default(dt):
     return int(max(2, numpy.ceil(2048 * dt.decompose().value)))
 
 
-def _discover_gate_points(ts, thresh, window):
-    """Discovers a set of indices where gating windows should be placed
-       based on a given amplitude threshold. Points are clustered within
-       a time window.
-
-    Parameters
-    ----------
-    ts : `~gwpy.timeseries.TimeSeries`
-        input time series that will be compared to the provided threshold
-
-    thresh : `float`
-        amplitude threshold, if the data exceeds this value a gating window
-        will be placed
-
-    window : `float`
-        time duration over which gating points will be clustered
-
-    Returns
-    -------
-    gates : `numpy.ndarray`
-        An array of time series indices where gating windows should be centered
-    """
-    gates = []
-    thresh_idx = numpy.where(ts.value > thresh)[0]
-    thresh_amp = ts.value[thresh_idx]
-    thresh_time = ts.times.value[thresh_idx]
-    sort_idx = numpy.argsort(thresh_amp)[::-1]
-
-    while numpy.any(sort_idx):
-        tip = sort_idx[0]
-        gates.append(tip)
-        remove_idx = [i for i in sort_idx
-                      if (thresh_time[i] < thresh_time[tip] + window)
-                      and (thresh_time[i] > thresh_time[tip] - window)]
-        sort_idx = numpy.array([s for s in sort_idx if s not in remove_idx])
-
-    return thresh_idx[gates]
-
-
 # -- TimeSeries ---------------------------------------------------------------
 
 class TimeSeries(TimeSeriesBase):
@@ -1664,20 +1625,17 @@ class TimeSeries(TimeSeriesBase):
         >>> overlay.show()
         """
         # Find points to gate based on a threshold
-        if whiten:
-            gates = _discover_gate_points(self.whiten(**whiten_kwargs),
-                                          threshold, cluster_window)
-        else:
-            gates = _discover_gate_points(self, threshold, cluster_window)
+        data = self.whiten(**whiten_kwargs) if whiten else self
+        gates = signal.find_peaks(data.value, height=threshold,
+                       distance=cluster_window*data.sample_rate.value)[0]
         out = self.copy()
 
         # Iterate over list of indices to gate and apply each one
+        nzero = int(abs(tzero)*self.sample_rate.value)
+        npad = int(abs(tpad)*self.sample_rate.value)
+        half = nzero + npad
+        ntotal = 2*half
         for gate in gates:
-            nzero = int(abs(tzero)*self.sample_rate.value)
-            npad = int(abs(tpad)*self.sample_rate.value)
-            half = nzero + npad
-            ntotal = 2*half
-
             # Set the boundaries for windowed data in the original time series
             left_idx = max(0, gate - half)
             right_idx = min(gate + half, len(self.value) - 1)
@@ -1690,11 +1648,7 @@ class TimeSeries(TimeSeriesBase):
 
             window = 1 - planck(ntotal, nleft=npad, nright=npad)
             window = window[left_idx_window:right_idx_window]
-            window_timeseries = TimeSeries(numpy.ones_like(self.value),
-                                           dt=self.dt, t0=self.t0)
-            window_timeseries[left_idx:right_idx] = window
-
-            out *= window_timeseries
+            out[left_idx:right_idx] *= window
 
         return out
 
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1635,7 +1635,7 @@ class TimeSeries(TimeSeriesBase):
         data = self.whiten(**whiten_kwargs) if whiten else self
         window_samples = cluster_window * data.sample_rate.value
         gates = find_peaks(abs(data.value), height=threshold,
-                                  distance=window_samples)[0]
+                           distance=window_samples)[0]
         out = self.copy()
 
         # Iterate over list of indices to gate and apply each one
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1626,6 +1626,12 @@ class TimeSeries(TimeSeriesBase):
         >>> ax.legend()
         >>> overlay.show()
         """
+        try:
+             from scipy.signal import find_peaks
+        except ImportError:
+             raise ImportError("Must have scipy>=1.1.0 to utilize "
+                               "this method.")
+
         # Find points to gate based on a threshold
         data = self.whiten(**whiten_kwargs) if whiten else self
         window_samples = cluster_window * data.sample_rate.value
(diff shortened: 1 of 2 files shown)


alurban:
@duncanmmacleod, the build failure reported by Travis seems to be from `pycbc` trying to import `glue.segments`

gwpy/timeseries/timeseries.py
```diff
@@ -58,6 +58,45 @@ def _fft_length_default(dt):
     return int(max(2, numpy.ceil(2048 * dt.decompose().value)))
 
 
+def _discover_gate_points(ts, thresh, window):
```
duncanmmacleod: Is this method a generic 'points above threshold' detector? If so, can we move it into the `Series` class as an instance method and give it an appropriate name?

gwpy/timeseries/timeseries.py
```diff
@@ -1567,6 +1606,98 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+        >>> ax.set_xlim(1225542713, 1225542733)
+        >>> ax.legend()
+        >>> overlay.show()
+        """
+        # Find points to gate based on a threshold
+        if whiten:
+            gates = _discover_gate_points(self.whiten(**whiten_kwargs),
+                                          threshold, cluster_window)
+        else:
+            gates = _discover_gate_points(self, threshold, cluster_window)
```
duncanmmacleod: I would modify this to:

```python
        data = self.whiten(**whiten_kwargs) if whiten else self
        gates = _discover_gate_points(data, threshold, cluster_window)
```

gwpy/timeseries/timeseries.py
```diff
@@ -1567,6 +1606,98 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+        out = self.copy()
+
+        # Iterate over list of indices to gate and apply each one
+        for gate in gates:
+            nzero = int(abs(tzero)*self.sample_rate.value)
+            npad = int(abs(tpad)*self.sample_rate.value)
+            half = nzero + npad
+            ntotal = 2*half
```
duncanmmacleod: it looks like all of this is gate-independent, so could be moved outside the loop.

gwpy/timeseries/timeseries.py
```diff
@@ -1567,6 +1606,98 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+            left_idx_window = half - (gate - left_idx)
+            right_idx_window = half + (right_idx - gate)
+
+            window = 1 - planck(ntotal, nleft=npad, nright=npad)
+            window = window[left_idx_window:right_idx_window]
+            window_timeseries = TimeSeries(numpy.ones_like(self.value),
+                                           dt=self.dt, t0=self.t0)
+            window_timeseries[left_idx:right_idx] = window
+
+            out *= window_timeseries
```
duncanmmacleod: Can you instead just do this and not have to create another timeseries?:

```python
            out[left_idx:right_idx] *= window
```

duncanmmacleod:
@tjma12, I have worked around the pycbc failure in #1053, please rebase.

gwpy/timeseries/timeseries.py
tjma12: It's a "points above threshold" detector with the automatic clustering that you'd want for gating. If I move it into the `Series` class, would you like me to make the clustering optional?

tjma12:
> @tjma12, I have worked around the pycbc failure in #1053, please rebase.

Rebased and squashed commits, I'll implement the other fixes once the build succeeds

gwpy/timeseries/timeseries.py
tjma12: Will replace this with `scipy.signal.find_peaks`

gwpy/timeseries/timeseries.py
```diff
@@ -1567,6 +1567,92 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
         out = in_.convolve(tdw, window=window)
         return out * numpy.sqrt(2 * in_.dt.decompose().value)
 
+    def gate(self, tzero=1.0, tpad=0.5, whiten=True,
+             threshold=50., cluster_window=0.5, **whiten_kwargs):
+        """Discovers a set of indices where gating windows should be placed
```
duncanmmacleod: I don't think this docstring is correct. Perhaps something like

```python
        """Mask clustered peaks above threshold using planck windows
 
        Parameters
        ------------
        ...
```

gwpy/timeseries/timeseries.py
```diff
@@ -1567,6 +1567,92 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+            a copy of the original `TimeSeries` that has had gating windows
+            applied
+
+        Examples
+        --------
+
+        Read data into a `TimeSeries`
+        >>> from gwpy.timeseries import TimeSeries
+        >>> data = TimeSeries.get('H1:GDS-CALIB_STRAIN', 1225542623,
+                                   1225542823)
```
duncanmmacleod: Can you modify this example to use `fetch_open_data`?

gwpy/timeseries/timeseries.py
```diff
@@ -1567,6 +1567,92 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+                              fftlength=4, overlap=2, method='median')
+
+        Plot the original data and the gated data
+        >>> overlay = data.plot(dpi=150, label='Ungated',
+                                color='dodgerblue', zorder=3)
+        >>> ax = overlay.gca()
+        >>> ax.plot(gated, label='Gated', color='orange', zorder=2)
+        >>> ax.set_xlim(1225542713, 1225542733)
+        >>> ax.legend()
+        >>> overlay.show()
```
duncanmmacleod: Can you post in this ticket an example of the figure you get?

gwpy/timeseries/timeseries.py
```diff
@@ -1567,6 +1567,92 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+        """
+        # Find points to gate based on a threshold
+        data = self.whiten(**whiten_kwargs) if whiten else self
+        window_samples = cluster_window*data.sample_rate.value
```
duncanmmacleod:
```suggestion
        window_samples = cluster_window * data.sample_rate.value
```

gwpy/timeseries/timeseries.py
```diff
@@ -1567,6 +1567,92 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+        gates = signal.find_peaks(data.value, height=threshold,
+                                  distance=window_samples)[0]
+        out = self.copy()
+
+        # Iterate over list of indices to gate and apply each one
+        nzero = int(abs(tzero)*self.sample_rate.value)
```
duncanmmacleod:
```suggestion
        nzero = int(abs(tzero) * self.sample_rate.value)
```

gwpy/timeseries/timeseries.py
```diff
@@ -1567,6 +1567,92 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+        npad = int(abs(tpad)*self.sample_rate.value)
```
duncanmmacleod:
```suggestion
        npad = int(abs(tpad) * self.sample_rate.value)
```

gwpy/timeseries/timeseries.py
```diff
@@ -1567,6 +1567,92 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+        half = nzero + npad
+        ntotal = 2*half
```
duncanmmacleod:
```suggestion
        ntotal = 2 * half
```

duncanmmacleod:
@tjma12, what's the status of this PR? It looks from @alurban's comment that you are going to put in one or more unit tests, which would be great.

tjma12:
@duncanmmacleod I plan to address your requested changes and write up a unit test today

gwpy/timeseries/timeseries.py
tjma12: Done, switched this to use a large glitch from O1 open data

gwpy/timeseries/timeseries.py
tjma12: See attached figure that is produced by updated docstring

gating_example

tjma12:
I wrapped the data in an `abs()` when running the peak finding since whitened data can glitch negatively. Unit test is next.

duncanmmacleod:
@tjma12, it turns out that scipy.signal.find_peaks` is new in scipy-1.1.0, I don't think we should change it, but can you add some importerror protectionn so that users get told what is wrong when they try to run `TimeSeries.gate` and it doesn't work.

You can auto-skip the unit test using a decorator

```python
from ...testing.utils import skip_minimum_version

...

        @skip_minimum_version("scipy", "1.1.0")
        def test_gate(self, ...)
```

alurban:
@tjma12, what @duncanmmacleod said, I was going to suggest a direct import like this:

```python
try:
    from scipy.signal import find_peaks
except ImportError:
    raise ImportError('find_peaks is only available in scipy >= 1.1.0')
# Find points to gate based on a threshold
data = self.whiten(**whiten_kwargs) if whiten else self
window_samples = cluster_window * data.sample_rate.value
gates = find_peaks(abs(data.value), height=threshold,
                   distance=window_samples)[0]
```

gwpy/timeseries/timeseries.py
```diff
@@ -1626,6 +1626,12 @@ def gate(self, tzero=1.0, tpad=0.5, whiten=True,
         >>> ax.legend()
         >>> overlay.show()
         """
+        try:
+             from scipy.signal import find_peaks
+        except ImportError:
+             raise ImportError("Must have scipy>=1.1.0 to utilize "
```
duncanmmacleod: @tjma12, can you implement it a little different, to preserve the traceback:

```python
        except ImportError as exc:
            exc.args = ("scipy.signal.find_peaks is only available in scipy >= 1.1.0",)
            raise
````

tjma12: Both have been implemented

PR #1057, opened 2019-02-26
This allows downloaded images based on new table column naming scheme

scottcoughlin2014:
Also I correct some egregious use of camel case. This is not critical for O3 release as old column names will exist in new table and vice versa. But it is better and more up to date.

Scotty

### DIFF
--- a/gwpy/table/gravityspy.py
+++ b/gwpy/table/gravityspy.py
@@ -68,9 +68,9 @@ class GravitySpyTable(EventTable):
             Specify exactly which durations you want to download
             default is to download all the avaialble GSpy durations.
 
-        kwargs: Optional TrainingSet and LabelledSamples args
+        kwargs: Optional training_set and labelled_samples args
             that will download images in a special way
-            ./"Label"/"SampleType"/"image"
+            ./"ml_label"/"sample_type"/"image"
 
         Returns
         -------
@@ -80,16 +80,16 @@ class GravitySpyTable(EventTable):
         import os
         # back to pandas
         try:
-            imagesDB = self.to_pandas()
+            images_db = self.to_pandas()
         except ImportError as exc:
             exc.args = ('pandas is required to download triggers',)
             raise
 
         # Remove any broken links
-        imagesDB = imagesDB.loc[imagesDB.imgUrl1 != '?']
+        images_db = images_db.loc[images_db.url1 != '']
 
-        TrainingSet = kwargs.pop('TrainingSet', 0)
-        LabelledSamples = kwargs.pop('LabelledSamples', 0)
+        training_set = kwargs.pop('training_set', 0)
+        labelled_samples = kwargs.pop('labelled_samples', 0)
         download_location = kwargs.pop('download_path',
                                        os.path.join('download'))
         duration_values = np.array([0.5, 1.0, 2.0, 4.0])
@@ -102,10 +102,10 @@ class GravitySpyTable(EventTable):
         duration_values = duration_values[duration_idx]
         duration_values = np.array([duration_values]).astype(str)
 
-        # LabelledSamples are only available when requesting the
-        if LabelledSamples:
-            if 'SampleType' not in imagesDB.columns:
-                raise ValueError('You have requested Labelled Samples '
+        # labelled_samples are only available when requesting the
+        if labelled_samples:
+            if 'sample_type' not in images_db.columns:
+                raise ValueError('You have requested ml_labelled Samples '
                                  'for a Table which does not have '
                                  'this column. Did you fetch a '
                                  'trainingset* table?')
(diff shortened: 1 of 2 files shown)


gwpy/table/gravityspy.py
```diff
@@ -211,18 +213,38 @@ def search(cls, uniqueID, howmany=10):
...
         base = 'https://gravityspytools.ciera.northwestern.edu' + \
             '/search/similarity_search_restful_API'
 
+        map_era_to_url = {
+            'ALL': "event_time BETWEEN 1126400000 AND 1229176818",
+            'O1': "event_time BETWEEN 1126400000 AND 1137250000",
+            'ER10': "event_time BETWEEN 1161907217 AND 1164499217",
+            'O2a': "event_time BETWEEN 1164499217 AND 1219276818",
+            'ER13': "event_time BETWEEN 1228838418 AND 1229176818",
+        }
```
duncanmmacleod: It would be nice to replace this with an API call that returns the eras, or somehow have the API itself accept an era in the query.

gwpy/table/gravityspy.py
```diff
@@ -211,18 +213,38 @@ def search(cls, uniqueID, howmany=10):
...
+
+        map_ifos_to_url = {
+            'H1': "\'H1\'",
+            'H1L1': "\'H1\', \'L1\'",
+            'H1L1V1': "\'H1\', \'L1\', \'V1\'",
+            'L1': "\'L1\'",
+            'L1V1': "\'L1\', \'V1\'",
+            'V1': "\'V1\'",
+        }
```
duncanmmacleod: This can be removed, see below.

gwpy/table/gravityspy.py
```diff
@@ -211,18 +213,38 @@ def search(cls, uniqueID, howmany=10):
...
+
         parts = {
             'howmany': howmany,
-            'imageid': uniqueID,
+            'imageid': gravityspy_id,
+            'era': map_era_to_url[era],
+            'ifo': map_ifos_to_url[ifos],
```
duncanmmacleod:
```suggestion
            'ifo': "{}".format(", ".join(
                map(repr, [ifos[i:i+2] for i in range(0, len(ifos), 2)]),
            )),
```

gwpy/table/gravityspy.py
```diff
@@ -231,5 +253,5 @@ def search(cls, uniqueID, howmany=10):
                 return GravitySpyTable(json.load(f))
         except HTTPError as exc:
             if exc.code == 500:
-                exc.msg = exc.msg + ', please confirm the uniqueID is valid'
+                exc.msg = exc.msg + ', confirm the gravityspy_id is valid'
```
duncanmmacleod:
```suggestion
                exc.msg += ', confirm the gravityspy_id is valid'
```

gwpy/table/gravityspy.py
scottcoughlin2014: I agree with the fact the API should either return a dict of available eras or it should be able to take in the string. I will work on this but may take a minute. Other requested changes sjust were pushed

duncanmmacleod:
There is a failure that is unrelated to this request, see gwpy/gwpy#1064. Once that is merged this PR can be rebased.

duncanmmacleod:
@scottcoughlin2014, the test query is reliably timing out, can you implement this patch to add a `remote_timeout` option to the `search` method:

```patch
diff --git a/gwpy/table/gravityspy.py b/gwpy/table/gravityspy.py
index db1f3f81..2d4e8095 100644
--- a/gwpy/table/gravityspy.py
+++ b/gwpy/table/gravityspy.py
@@ -20,6 +20,7 @@
 """

 from ..utils import mp as mp_utils
+from ..utils.env import bool_env
 from .table import EventTable
 import numpy as np

@@ -191,7 +192,7 @@ class GravitySpyTable(EventTable):

     @classmethod
     def search(cls, gravityspy_id, howmany=10,
-               era='ALL', ifos='H1L1'):
+               era='ALL', ifos='H1L1', remote_timeout=20):
         """perform restful API version of search available here:
         https://gravityspytools.ciera.northwestern.edu/search/

@@ -242,7 +243,7 @@ class GravitySpyTable(EventTable):
         url = '{}/?{}'.format(base, search)

         try:
-            with get_readable_fileobj(url) as f:
+            with get_readable_fileobj(url, remote_timeout=remote_timeout) as f:
                 return GravitySpyTable(json.load(f))
         except HTTPError as exc:
             if exc.code == 500:
```

duncanmmacleod:
Also, please rebase to fix the test failure that has nothing to do with this change.

duncanmmacleod:
@scottcoughlin2014, please rebase again, apologies for the inconvenience.

PR #1059, opened 2019-02-26
Moved all Linux CI jobs over to CircleCI

duncanmmacleod:
This PR moves all linux-based CI jobs onto circle-ci; the use of docker images on that platform should speed things up.


duncanmmacleod:
I'll fix the reduced coverage in a subsequent PR.

PR #1067, opened 2019-02-28
Swap the defaults for applying ADC bias and slope calibration

alurban:
This PR swaps the defaults for applying ADC bias and slope calibration to a dynamic choice based on the channel name. For H1 and L1 channels, bias and slope are ignored, else they are applied.

cc @duncanmmacleod

### DIFF
--- a/gwpy/timeseries/core.py
+++ b/gwpy/timeseries/core.py
@@ -72,6 +72,8 @@ _UFUNC_STRING = {
 }
 
 
+# -- utilities ----------------------------------------------------------------
+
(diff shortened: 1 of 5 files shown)


duncanmmacleod:
@alurban (with permission) I have updated this to use a dynamic scaling that sets the default to `False` for LIGO channels _only_ when using the `.read`, `.fetch`, `.find`, and `.get` methods.

alurban:
@duncanmmacleod, your changes look good to me, thanks for handling this

PR #1070, opened 2019-03-05
Added gwpy.io.gwf.data_segments function

duncanmmacleod:
This PR adds `gwpy.io.gwf.data_segments()` (copied from gwpy/pyomicron#88) to determine available data segments for a given channel in a list of GWF paths.

This is useful when trying to read data from a list of paths, some of which don't actually contain the channel you want.

### DIFF
--- a/gwpy/io/gwf.py
+++ b/gwpy/io/gwf.py
@@ -19,11 +19,13 @@
 """I/O utilities for GWF files using the lalframe or frameCPP APIs
 """
 
-import six
+import warnings
 
+import six
 from six.moves.urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


alurban:
@duncanmmacleod, do we want to wrap [**this line**](https://github.com/gwpy/gwpy/blob/develop/gwpy/timeseries/tests/test_timeseries.py#L351) with a try-except so that we can skip the test if it fails because of an upstream issue with GWOSC?

alurban:
@duncanmmacleod, at any rate, the build is now passing, so it's probably a moot point.

PR #1085, opened 2019-03-28
gwpy.io: return unique creds from parse_keytab

duncanmmacleod:
This PR fixes an issue in `gwpy.io.kerberos` whereby keytab files with multiple copies of the same kvno/principal would be returned, causing `kinit()` to fall over. `parse_keytab()` has been modified to return an ordered, unique list of credentials, and also to return a `list` of `tuple`, rather than a `list` of `list`.

cc @scottcoughlin2014

### DIFF
--- a/gwpy/io/tests/test_kerberos.py
+++ b/gwpy/io/tests/test_kerberos.py
@@ -35,6 +35,7 @@ KLIST = b"""Keytab name: FILE:/test.keytab
 KVNO Principal
 ---- -------------------------------
    1 albert.einstein@LIGO.ORG
+   2 ronald.drever@LIGO.ORG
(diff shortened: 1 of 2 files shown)


gwpy/io/tests/test_kerberos.py
```diff
@@ -35,6 +35,7 @@
 KVNO Principal
 ---- -------------------------------
    1 albert.einstein@LIGO.ORG
+   2 ronald.drever@LIGO.ORG
```
alurban: @duncanmmacleod, is this intentionally duplicated?

duncanmmacleod: @alurban, yes, to check that the `ronald.drever` principal only gets returned once.

PR #1088, opened 2019-04-02
Fixed GPS scaling issue again

duncanmmacleod:
This PR fixes issues with dynamic scaling when plotting GPS scales (again).

| Before | After |
| ------ | ----- |
| before | after |


duncanmmacleod:
@alurban, thanks for the review, I have added a unit test, in case you wanted to look at that.

PR #1102, opened 2019-04-25
Updated custom legend formatting

duncanmmacleod:
This PR changes the way legends are customised by default, to make things easier to modify for the user.

The old way was to use some extra keyword arguments, namely `linewidth`, to specify a new default linewidth in the legend. However, those keywords aren't recognised by default matplotlib figures, so calling `ax.legend(linewidth=4)` isn't cross-compatible between gwpy and non-gwpy plots. There's also no way to disable the custom linewidth and use the linewidth of the original handle.

The new way is to define a custom legend handler, which operates on `Line2D` objects, and sets the default linewidth. The interface is now that gwpy's `Axes.legend` method inserts the new handler into the existing `handler_map` keyword of matplotlib's `Axes.legend` method, which is entirely cross-compatible with non-gwpy plots. This gwpy handler can be disabled simply be passing `handler_map=None` to `ax.legend().

This PR also includes a docs page that describes the legend customisation, how to disable it, or how to modify it with your own custom legend handler.

### DIFF
--- a/docs/plot/legend.rst
+++ b/docs/plot/legend.rst
@@ -1,0 +1,62 @@
+.. currentmodule:: gwpy.plot
+
+.. _gwpy-plot-legend:
+
+######################
+Custom legends in GWpy
+######################
+
+GWpy overrides the default :class:`~matplotlib.axes.Axes` class with one that
+uses a different default legend handler for line plots.
+This means that, by default, a lines in a legend will be thicker than on a
+standard matplotlib figure:
+
+.. plot::
+   :include-source:
+   :context: reset
+
+   >>> import gwpy  # <- import anything from gwpy
+   >>> from matplotlib import pyplot
+   >>> fig = pyplot.figure()
+   >>> ax = fig.gca()
+   >>> ax.plot(range(10), label='My data')
+   >>> ax.legend()
+   >>> fig.show()
+
+This can be disabled by passing an empty ``handler_map`` to the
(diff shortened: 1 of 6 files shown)


docs/plot/legend.rst
```diff
@@ -0,0 +1,62 @@
...
+
+.. _gwpy-plot-legend:
+
+######################
+Custom legends in GWpy
+######################
+
+GWpy overrides the default :class:`~matplotlib.axes.Axes` class with one that
+uses a different default legend handler for line plots.
+This means that, by default, a lines in a legend will be thicker than on a
```
alurban: @duncanmmacleod, I think there's a typo on this line:

```suggestion
This means that, by default, lines in a legend will be thicker than on a
```

docs/plot/legend.rst
```diff
@@ -0,0 +1,62 @@
...
+   :context: close-figs
+
+   >>> fig = pyplot.figure()
+   >>> ax = fig.gca()
+   >>> ax.plot(range(10), label='My data')
+   >>> ax.legend(handler_map=None)
+   >>> fig.show()
+
+Similarly, you can implement your own custom legend handler and overwrite
+things yourself, below is a simple example, but for more details see
```
alurban: @duncanmmacleod, this probably comes down to style, but I think the last two clauses ought to be a separate sentence:

```suggestion
things yourself. Below is a simple example, but for more details see
```

PR #1107, opened 2019-05-08
Improved reading tables from GWOSC

duncanmmacleod:
This PR improves the parsing of open-data catalog event tables from GWOSC, mainly by detecting and replacing the `'NA'` that GWOSC uses to mark missing data. The logic here is a bit more complicated, because we have to determine the underlying data type of the column without the missing values, in order to replace the `'NA'` with the appropriate null value (`0` for `int`, `nan` for float).

### DIFF
--- a/gwpy/table/io/losc.py
+++ b/gwpy/table/io/losc.py
@@ -80,6 +80,25 @@ def _mask_replace(value, dtype):
     return value
 
 
+def _mask_column(col):
+    """Find and replace missing data in a column
+
+    Returns the new data, and the mask as `list`
+    """
+    # find masked indices
+    mask = [v in MISSING_DATA for v in col]
+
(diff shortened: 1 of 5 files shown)


gwpy/table/io/losc.py
```diff
@@ -32,11 +36,25 @@
     read_with_selection,
 )
 
-# handle custom GWOSC units
+#: custom GWOSC unit mapping
```
alurban: @duncanmmacleod, is this colon here on purpose?

duncanmmacleod: Yes, this is how one would, in principle add a docstring to a module variable for sphinx to parse.

alurban:
Ah, that’s fancy, this is good to now for the future.

PR #1108, opened 2019-05-08
Implemented __array_ufunc__ for gwpy.types.Array

duncanmmacleod:
This PR implements a top-level `Array.__array_ufunc__` method that returns scalars as `Quantity` objects from ufuncs. This will be required to support astropy-3.2, but should be backwards compatible.

### DIFF
--- a/gwpy/timeseries/statevector.py
+++ b/gwpy/timeseries/statevector.py
@@ -34,7 +34,7 @@ from six.moves import range
 
 import numpy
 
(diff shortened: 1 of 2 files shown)


duncanmmacleod:
> Just to double-check, this does _not_ require astropy >= 2.0.0, right?

Yes, the fact that the build tests pass guarantees that, because (at least) `el7` only provides astropy-1.2.x (or thereabouts).

PR #1111, opened 2019-05-21
Raise IndexError early when reading an empty list of files

duncanmmacleod:
This PR modifies `gwpy.io.mp.read_multi` (the function that underpins `EventTable.read`, `TimeSeries.read`, etc) to raise an `IndexError` early if the input `source` argument is parsed as an empty list.

Currently the behaviour in this condition is unpredictable, and dependent on the other keyword arguments, which doesn't make a great UX.

I also added an independent test suite for that function.

### DIFF
--- a/gwpy/io/tests/test_mp.py
+++ b/gwpy/io/tests/test_mp.py
@@ -53,7 +53,7 @@ def tmp2():
 def test_read_multi_single(tmp1):
     """Check that serial processing works with `mp.read_multi`
     """
-    tab = io_mp.read_multi(vstack, Table, tmp1, verbose=True)
+    tab = io_mp.read_multi(vstack, Table, tmp1.name, verbose=True)
     assert tab.colnames == ["a", "b", "c"]
     assert list(tab[0]) == [1, 2, 3]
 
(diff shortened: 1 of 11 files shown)


gwpy/io/tests/test_mp.py
```diff
@@ -0,0 +1,115 @@
...
+    with pytest.raises(IndexError) as exc:
+        io_mp.read_multi(1, int, [])
+    assert str(exc.value) == "cannot read int from empty source list"
+
+
+def test_read_multi_not_a_file():
+    """Check that a strange input gets passed along properly
+    so that errors can be raise by the reader.
+    """
+    with pytest.raises(TypeError):
```
alurban: @duncanmmacleod, in python < 3.6, this use case actually raises a `ValueError`, which I suspect is due to a pinned lower version of astropy. So, perhaps something like this is necessary?

```suggestion
    exc = TypeError if sys.version_info < (3, 6) else ValueError
    with pytest.raises(exc):
```

duncanmmacleod: I have added a check against the astropy version in 0d9f5b2a2e1230a7585eac528b7b3fc63944be23.

duncanmmacleod:
> There are a few more in Appveyor that I haven't yet been able to track down, and others related to plot epochs that probably aren't related to this PR.

Yes, see #1112 for the matplotlib-3.1 epoch fixes.

PR #1115, opened 2019-05-30
Introduce a SegmentList.to_table method

alurban:
This PR introduces a new method, `SegmentList.to_table`, to cast `SegmentList` objects as type `~gwpy.table.Table`. This is provided mainly to have a way to write `SegmentList` objects to standard CSV tables, which has proven useful in delivering data products to LIGO management; see [**this tool**](https://git.ligo.org/detchar/ligo-summary-pages/blob/master/scripts/gw_epoch_time_accounting) for the LIGO summary pages (requires `LIGO.ORG` credentials).

cc @duncanmmacleod

### DIFF
--- a/gwpy/segments/segments.py
+++ b/gwpy/segments/segments.py
@@ -144,6 +144,25 @@ class SegmentList(segmentlist):
         return self
     coalesce.__doc__ = segmentlist.coalesce.__doc__
 
+    def to_table(self):
+        """Convert this `SegmentList` to a `~gwpy.table.Table`
+
+        The resulting `Table` has four columns: `index`, `start`, `end`, and
+        `duration`, corresponding to the zero-counted list index, GPS start
+        and end times, and total duration in seconds, respectively.
+
+        This method exists mainly to provide a way to write `SegmentList`
+        objects in comma-separated value (CSV) format, via the
+        :meth:`~gwpy.table.Table.write` method.
+        """
+        from ..table import Table
+        return Table([
+            list(range(len(self))),
+            [seg[0] for seg in self],
+            [seg[1] for seg in self],
+            [abs(seg) for seg in self],
+        ], names=('index', 'start', 'end', 'duration'))
+
     # -- i/o ------------------------------------
 
     @classmethod
--- a/gwpy/segments/segments.py
+++ b/gwpy/segments/segments.py
@@ -23,6 +23,7 @@ configuration.
 """
 
 from astropy.io import registry as io_registry
+from astropy.table import Table
 
 from ligo.segments import (segment, segmentlist, segmentlistdict)
 
@@ -145,7 +146,7 @@ class SegmentList(segmentlist):
     coalesce.__doc__ = segmentlist.coalesce.__doc__
 
     def to_table(self):
-        """Convert this `SegmentList` to a `~gwpy.table.Table`
+        """Convert this `SegmentList` to a `~astropy.table.Table`
 
         The resulting `Table` has four columns: `index`, `start`, `end`, and
         `duration`, corresponding to the zero-counted list index, GPS start
@@ -153,9 +154,8 @@ class SegmentList(segmentlist):
 
         This method exists mainly to provide a way to write `SegmentList`
         objects in comma-separated value (CSV) format, via the
-        :meth:`~gwpy.table.Table.write` method.
+        :meth:`~astropy.table.Table.write` method.
         """
-        from ..table import Table
         return Table([
             list(range(len(self))),
             [seg[0] for seg in self],
--- a/gwpy/segments/segments.py
+++ b/gwpy/segments/segments.py
@@ -156,12 +156,10 @@ class SegmentList(segmentlist):
         objects in comma-separated value (CSV) format, via the
         :meth:`~astropy.table.Table.write` method.
         """
-        return Table([
-            list(range(len(self))),
-            [seg[0] for seg in self],
-            [seg[1] for seg in self],
-            [abs(seg) for seg in self],
-        ], names=('index', 'start', 'end', 'duration'))
+        return Table(
+            rows=[(i, s[0], s[1], abs(s)) for i, s in enumerate(self)],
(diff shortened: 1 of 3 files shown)


gwpy/segments/tests/test_segments.py
```diff
@@ -82,6 +84,15 @@ def test_coalesce(self):
...
 
+    def test_to_table(self, segmentlist):
+        segtable = segmentlist.to_table()
+        assert isinstance(segtable, Table)
+        assert_array_equal(segtable['index'], list(range(len(segmentlist))))
+        assert_array_equal(segtable['start'], [seg[0] for seg in segmentlist])
+        assert_array_equal(segtable['end'], [seg[1] for seg in segmentlist])
+        assert_array_equal(
+            segtable['duration'], [abs(seg) for seg in segmentlist])
+
```
duncanmmacleod: Can you just instead use `assert_table_equal` and build the expected table on the fly? Something like:

```python
    def test_to_table(self, segmentlist):
        segtable = segmentlist.to_table()
        assert_table_equal(
            segtable,
            Table(
                rows=[
                    (0, 1, 2, 1),
                    (1, 3, 4, 1),
                    (2, 4, 6, 2),
                    (3, 8, 10, 2),
                ],
                names=('index', 'start', 'end', 'duration'),
            ),
        )
```
(untested)

gwpy/segments/segments.py
```diff
@@ -144,6 +144,25 @@ def coalesce(self):
         return self
     coalesce.__doc__ = segmentlist.coalesce.__doc__
 
+    def to_table(self):
+        """Convert this `SegmentList` to a `~gwpy.table.Table`
```
duncanmmacleod: this class is really just `astropy.table.Table`
```suggestion
        """Convert this `SegmentList` to a `~astropy.table.Table`
```

gwpy/segments/segments.py
```diff
@@ -144,6 +144,25 @@ def coalesce(self):
...
+
+        The resulting `Table` has four columns: `index`, `start`, `end`, and
+        `duration`, corresponding to the zero-counted list index, GPS start
+        and end times, and total duration in seconds, respectively.
+
+        This method exists mainly to provide a way to write `SegmentList`
+        objects in comma-separated value (CSV) format, via the
+        :meth:`~gwpy.table.Table.write` method.
+        """
+        from ..table import Table
```
duncanmmacleod:
```suggestion
        from astropy.table import Table
```

gwpy/segments/tests/test_segments.py
```diff
@@ -25,7 +25,9 @@
 
 import h5py
 
-from ...testing.utils import (assert_segmentlist_equal, TemporaryFilename)
+from ...testing.utils import (assert_array_equal, assert_segmentlist_equal,
```
duncanmmacleod:
```suggestion
from ...testing.utils import (assert_segmentlist_equal, assert_table_equal,
```

gwpy/segments/tests/test_segments.py
```diff
@@ -25,7 +25,9 @@
...
+                              TemporaryFilename)
+from ...table import Table
```
duncanmmacleod:
```suggestion
from astropy.table import Table
```
(with appropriate repositioning)

gwpy/segments/tests/test_segments.py
alurban: @duncanmmacleod, your proposed unit test passes (and runs about 6.5 times faster), so I've implemented it in the latest commit.

alurban:
Thanks for your review @duncanmmacleod, I've addressed your comments and moved the method-level import of `gwpy.table.Table` out into a _module_-level import of `astropy.table.Table`, since it no longer causes a circular import.

gwpy/segments/segments.py
```diff
@@ -144,6 +145,24 @@ def coalesce(self):
...
+        This method exists mainly to provide a way to write `SegmentList`
+        objects in comma-separated value (CSV) format, via the
+        :meth:`~astropy.table.Table.write` method.
+        """
+        return Table([
+            list(range(len(self))),
+            [seg[0] for seg in self],
+            [seg[1] for seg in self],
+            [abs(seg) for seg in self],
+        ], names=('index', 'start', 'end', 'duration'))
```
duncanmmacleod: Looking at this again, does it make more sense to only unpack the segmentlist once:

```python
        return Table(
            list(zip(*((i, s[0], s[1], abs(s)) for i, s in enumerate(a)))),
            names=('index', 'start', 'end', 'duration'),
        )

Comment (alurban) [reply gwpy/segments/segments.py]: @duncanmmacleod, this speeds up the unit test by another factor of 2 (and takes up fewer lines), so I went ahead and pushed it with the last commit.

**Comment by duncanmmacleod** [comment]
@alurban, the circleci-based conda builds are failing, the only thing I can thing of to try is to refresh the circleci cache:

```patch
diff --git a/.circleci/config.yml b/.circleci/config.yml
index 2a9973e0..84765bbf 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,7 +82,7 @@ aliases:
         - checkout
         - *attach_workspace
         - restore_cache:
-            key: v4-gwpy-{{ .Environment.CIRCLE_JOB }}
+            key: v5-gwpy-{{ .Environment.CIRCLE_JOB }}
         - run: *set_python_environment
         - run:
             name: Install
@@ -90,7 +90,7 @@ aliases:
         - run: *run-tests
         - run: *codecov
         - save_cache:
-            key: v4-gwpy-{{ .Environment.CIRCLE_JOB }}
+            key: v5-gwpy-{{ .Environment.CIRCLE_JOB }}
             paths:
               - "/opt/conda/envs/gwpyci"
         - *store_test_results
```

alurban:
@duncanmmacleod, I'm giving it a shot, here goes nothing...

alurban:
@duncanmmacleod, this gambit was successful, thanks for the suggestion

PR #1117, opened 2019-05-31
WIP: Attempt to fix the codecov path issues

duncanmmacleod:
This PR attempts to fix issues with the codecov reports, whereby the paths in the codecov reports don't match those in the repo, so you can't read the reports properly ([example](https://codecov.io/gh/gwpy/gwpy/tree/068b09fd0aaa0cfee7bcf459d79a16a11616daa8)).

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -11,3 +11,5 @@ coverage:
       default:
         target: 90
 
+fixes:
+  - "::gwpy/"


duncanmmacleod:
This can't really be tested until the [timeout issues](https://twitter.com/duncanmmacleod/status/1134463833337159680) are resolved, which will probably come when codecov fixes itself, or we reduce the number of builds by abandoning system packages and python2.

Issue #1119, opened 2019-06-08
Problem reading segments from HDF5 group


areeda:
I'm working on a single HDF5 file with segments and triggers in a single file.
When reading he segetts wih
```python
    infile = '/data/t/testtrig1.h5'
    inf = h5py.File(infile, 'r')
    segpth - '/segments/L1:Manual:1'
    seg = DataQualityFlag.read(inf, path=segpth, verbose=True, format='hdf5')
```
Produces error (Python 3):
```pytb
Traceback (most recent call last):
  File "/Users/areeda/ligo/python/Learn/hdf5/bin/sgFind.py", line 72, in <module>
    seg = DataQualityFlag.read(inf, path=segpth, verbose=True, format='hdf5')
  File "/Users/areeda/miniconda2/envs/ligo-py36/lib/python3.6/site-packages/gwpy/segments/flag.py", line 635, in read
    return io_read_multi(combiner, cls, source, *args, **kwargs)
  File "/Users/areeda/miniconda2/envs/ligo-py36/lib/python3.6/site-packages/gwpy/io/mp.py", line 96, in read_multi
    nproc, _read_single_file, files, verbose=verbose, unit='files')
  File "/Users/areeda/miniconda2/envs/ligo-py36/lib/python3.6/site-packages/gwpy/utils/mp.py", line 132, in multiprocess_with_queues
    return list(map(_inner, inputs))
  File "/Users/areeda/miniconda2/envs/ligo-py36/lib/python3.6/site-packages/gwpy/utils/mp.py", line 119, in _inner
    return func(x)
  File "/Users/areeda/miniconda2/envs/ligo-py36/lib/python3.6/site-packages/gwpy/io/mp.py", line 81, in _read_single_file
    return fobj, io_read(cls, fobj, *args, **kwargs)
  File "/Users/areeda/miniconda2/envs/ligo-py36/lib/python3.6/site-packages/astropy/io/registry.py", line 517, in read
    data = reader(*args, **kwargs)
  File "/Users/areeda/miniconda2/envs/ligo-py36/lib/python3.6/site-packages/gwpy/io/hdf5.py", line 60, in decorated_func
    return func(fobj, *args, **kwargs)
  File "/Users/areeda/miniconda2/envs/ligo-py36/lib/python3.6/site-packages/gwpy/segments/io/hdf5.py", line 151, in read_hdf5_flag
    return DataQualityFlag(active=active, known=known, **dict(dataset.attrs))
[...]

PR #1120, opened 2019-06-08
fix string/byte array problem reading hdf5

areeda:
Addresses issue #1119 by casting name to str in __init__


duncanmmacleod:
@areeda, I believe that this should have been fixed in #1104, its just that this hasn't been released yet. I will prioritise a new release.

areeda:
Duncan is right.  This was fixed on maser. I was complaining about an  older version.  Sorry.

PR #1124, opened 2019-06-22
Introduce a range_spectrogram method as a figure-of-merit

alurban:
This PR introduces two new functions in `gwpy.astro`:

* `range_timeseries`
* `range_spectrogram`

Unit tests and documentation examples for these new functions are also included.

The `range_spectrogram` utility works by first calculating a standard PSD spectrogram (or accepting one as an input), then stepping through its time bins, passing each PSD measurement to either the `astro.burst_range_spectrum` or `astro.inspiral_range_psd` function depending on the use case. The `range_timeseries` utility follows a similar procedure, but returns a `TimeSeries` record of the output of `astro.burst_range` or `astro.inspiral_range`.

**Notes:**
* This PR is mainly to address requests from Advanced LIGO commissioners, who want to analyze spectral contributions to the binary neutron star inspiral range on the LIGO summary pages
* Both methods default to a 1.4-1.4 binary neutron star (BNS) inspiral range, but support either inspiral (of arbitrary masses) or burst (of arbitrary energy) range calculations

cc @duncanmmacleod

### DIFF
--- a/examples/miscellaneous/index.rst
+++ b/examples/miscellaneous/index.rst
@@ -9,4 +9,5 @@ Miscellaneous examples
    :numbered:
 
    range-timeseries
+   range-spectrogram
    open-data-spectrogram
--- a/gwpy/astro/range.py
+++ b/gwpy/astro/range.py
@@ -28,6 +28,9 @@ from scipy import integrate
 
 from astropy import (units, constants)
 
+from ..timeseries import TimeSeries
+from ..spectrogram import Spectrogram
+
 __author__ = 'Duncan Macleod <duncan.macleod@ligo.org>'
 __credits__ = 'Alex Urban <alexander.urban@ligo.org>'
 
@@ -288,7 +291,7 @@ def burst_range(psd, snr=8, energy=1e-2, fmin=100, fmax=500):
     return r.to('Mpc')
 
 
-def range_timeseries(hoft, stride, fftlength=None, overlap=None,
+def range_timeseries(hoft, stride=None, fftlength=None, overlap=None,
                      window='hann', method=DEFAULT_FFT_METHOD, nproc=1,
                      **rangekwargs):
     """Measure timeseries trends of astrophysical detector range (Mpc)
@@ -296,11 +299,12 @@ def range_timeseries(hoft, stride, fftlength=None, overlap=None,
 
     Parameters
     ----------
-    hoft : `~gwpy.timeseries.TimeSeries`
+    hoft : `~gwpy.timeseries.TimeSeries` or `~gwpy.spectrogram.Spectrogram`
         record of gravitational-wave strain output from a detector
 
-    stride : `float`
-        desired step size (seconds) of range timeseries
+    stride : `float`, optional
+        desired step size (seconds) of range timeseries, required if
+        `hoft` is an instance of `TimeSeries`
 
     fftlength : `float`, optional
         number of seconds in a single FFT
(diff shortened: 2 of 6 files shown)


alurban:
@duncanmmacleod, here is an example percentile spectrum made using the new method:

L1-OBSERVING_C02_BNS_RANGE_INTEGRAND-1186963218-86400

and here is the integral over the 50th percentile curve, showing the total cumulative contribution to BNS range (squared):

L1-OBSERVING_C02_CUMULATIVE_SNR-1186963218-86400

alurban:
@duncanmmacleod, below are the plots made by the examples included with this PR:

range-test

range-spectrogram-test

jeffrey-kissel:
> @duncanmmacleod, here is an example percentile spectrum made using L1 data from 2017-08-17 using the new method:
> 
> L1-OBSERVING_C02_BNS_RANGE_INTEGRAND-1186963218-86400
> 
> and here is the integral over the 50th percentile curve, showing the total cumulative contribution to BNS range (squared):
> 
> L1-OBSERVING_C02_CUMULATIVE_SNR-1186963218-86400

Testing out how to use this system. Hoping to comment directly on ... the "commit?" ... about the range integrand plot...
Hoping for logarithmic frequency axis on the percentile integrand, take the square-root so it’s in Mpc / rtHz, and limit from 5 or 10 Hz to 500 Hz.

jeffrey-kissel:
> L1-OBSERVING_C02_BNS_RANGE_SPECTRUM-1186963218-86400

Hey Alex -- sorry if I wasn't more clear -- I'm looking for a semilogx plot (leave the y axis linear). However, honestly, since people will ask you for it both ways (linear and log y axis) -- why not make both?

alurban:
@jeffrey-kissel, apologies for the misunderstanding, I've updated these example plots to be semilogx to visualize what the eventual final product will look like. Once the plots are supported by the summary pages, they will be infinitely customizable as all other plots are, i.e. we can freely set their axis limits, axis scaling, titles, trace colors, etc.

gwpy/astro/range.py
```diff
@@ -283,3 +287,184 @@ def burst_range(psd, snr=8, energy=1e-2, fmin=100, fmax=500):
     # normalize and return
     r = units.Quantity(result / (fmax - fmin), unit=integrand.unit) ** (1/3.)
     return r.to('Mpc')
+
+
+def range_trends(hoft, stride, fftlength=None, overlap=None,
```
duncanmmacleod: I think I would rather have this function be called `range_timeseries`.

examples/miscellaneous/index.rst
```diff
@@ -8,5 +8,6 @@ Miscellaneous examples
    :maxdepth: 1
    :numbered:
 
-   range-timeseries
+   range
```
duncanmmacleod: I don't think this should have been changed

```suggestion
   range-timeseries
```

gwpy/astro/range.py
alurban: @duncanmmacleod, makes sense to me, this is done in commit d8dcdd0ee5bb2cc233a40ea89357183b40f24e5f including updates to all references to this function.

examples/miscellaneous/index.rst
alurban: @duncanmmacleod, good catch, this typo was a holdover from a previous attempt to implement all this that I just forgot to revert back. The module name is updated in commit d8dcdd0ee5bb2cc233a40ea89357183b40f24e5f.

PR #1131, opened 2019-07-10
A few compatibility fixes for pytest-5

duncanmmacleod:
This PR fixes a few issues related to the release of pytest-5.0.0, which is python3 only, and exposes a couple of bugs in some test code to do with error message handling.


duncanmmacleod:
@gwpy/maintainers, the circleCI build is still failing, but only because of issues with numpy-1.17.0rc1 which are fixed in #1132.

PR #1132, opened 2019-07-10
Update gwpy.table test code for numpy-1.17

duncanmmacleod:
This PR fixes up some test failures when running with `numpy-1.17.0rc1`, see astropy/astropy#8976.


duncanmmacleod:
@gwpy/maintainers, in the end, to fix a number of build failures, I have removed any and all support for python3.4, which has already been dropped by the upstream maintainers anyway.

PR #1139, opened 2019-07-22
Updates to custom LogFormatters for plots

### LINKED ISSUE #1129: Log-axis formatting occasionally malformed
When using matplotlib-3.1+ and the latest head of gwpy `master`, I see formatting issues on log-scaled axes:

V1-LOCKED_56FD0E_PYCBC_LIVE_END_TIME_TEMPLATE_DURATION_NEW_SNR_TRIGGERS-1242518418-3600

duncanmmacleod:
This PR reworks the custom log formatters defined in `gwpy.plot.log` to improve the display of ticks across a number of different ranges. The end result is actually simpler code (less custom objects), which is nice.

Fixes #1129.

### DIFF
--- a/docs/install/index.rst
+++ b/docs/install/index.rst
@@ -55,7 +55,7 @@ GWpy has the following strict requirements:
    |h5py|_             ``>=1.3``
    |ligo-segments|_    ``>=1.0.0``
    |ligotimegps|_      ``>=1.2.1``
-   |matplotlib|_       ``!=2.1.0,!=2.1.1,>=1.2.0``
+   |matplotlib|_       ``>=3.1.0``
    |numpy|_            ``>=1.12.0``
    |dateutil|_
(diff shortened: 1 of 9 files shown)


duncanmmacleod:
Gah. This is not compatible with matplotlib<2.0.0, so we can't deploy it for 1.0.0. Let's leave this open and hopefully we can use it for 2.0.0 (or earlier if I can get our EL7 machines force-updated to a newer matplotlib).

alurban:
@duncanmmacleod, I'm alright with that plan, if it's not compatible with our current EL7 setup then we're better off waiting until a version that explicitly abandons support for it.

duncanmmacleod:
This should be rebased and revived when #1225 is merged.

duncanmmacleod:
@alurban, no, I don't have example images - the changes have minimal impact on the figures, but clean up the relevant code and improve the docs.

PR #1145, opened 2019-08-22
Added fixed length args to TimeSeries.taper

tjma12:
Adds `duration` and `nsamples` options to `TimeSeries.taper()` that allow user to specify a time duration or number of samples to taper rather than using the automatic tapering window. If neither are provided, the method defaults to the automatic tapering. If both `duration` and `nsamples` are provided, a `ValueError` is thrown to avoid mismatched taper arguments.

I also added a couple of explicit type casts to `int` that were causing issues when the specified number of taper samples was more than half of the duration of the data.

### DIFF
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -856,6 +856,22 @@ class TestTimeSeries(_TestTimeSeriesBase):
         assert tapered.size == data.size
         utils.assert_allclose(data.value, numpy.cos(10*numpy.pi*t))
 
+        # run the same tests for a user-specified taper duration
+        dtapered = data.taper(duration=0.1)
+        assert dtapered[0].value == 0
+        assert dtapered[-1].value == 0
+        assert dtapered.unit == data.unit
+        assert dtapered.size == data.size
+        utils.assert_allclose(data.value, numpy.cos(10*numpy.pi*t))
+
+        # run the same tests for a user-specified number of samples to taper
+        stapered = data.taper(nsamples=10)
+        assert stapered[0].value == 0
+        assert stapered[-1].value == 0
+        assert stapered.unit == data.unit
+        assert stapered.size == data.size
+        utils.assert_allclose(data.value, numpy.cos(10*numpy.pi*t))
+
     def test_inject(self):
         # create a timeseries out of an array of zeros
         duration, sample_rate = 1, 4096
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1421,12 +1421,12 @@ class TimeSeries(TimeSeriesBase):
             `'right'`, or `'leftright'`
 
         duration : `float`, optional
-            the duration of time to taper, cannot be specified if `nsamples`
-            is provided as an argument
+            the duration of time to taper, will override `nsamples`
+            if both are provided as arguments
 
         nsamples : `int`, optional
-            the number of samples to taper, cannot be specified if `duration`
-            is provided as an argument
+            the number of samples to taper, will be overridden by `duration`
+            if both are provided as arguments
 
         Returns
         -------
@@ -1474,10 +1474,7 @@ class TimeSeries(TimeSeriesBase):
         if side not in ('left', 'right', 'leftright'):
             raise ValueError("side must be one of 'left', 'right', "
                              "or 'leftright'")
-        if duration and nsamples:
-            raise ValueError("only one of 'duration' or 'nsamples' may be "
-                             "provided as arguments")
-        elif duration:
+        if duration:
             nsamples = int(duration * self.sample_rate.to("Hz").value)
         out = self.copy()
         # if a duration or number of samples is not specified, automatically
@@ -1488,11 +1485,10 @@ class TimeSeries(TimeSeriesBase):
             mini, = signal.argrelmin(out.value)
             maxi, = signal.argrelmax(out.value)
         if 'left' in side:
-            nleft = nsamples if nsamples else max(mini[0], maxi[0])
+            nleft = nsamples or max(mini[0], maxi[0])
             nleft = min(nleft, int(self.size/2))
         if 'right' in side:
-            nright = (nsamples if nsamples else
-                      out.size - min(mini[-1], maxi[-1]))
+            nright = nsamples or out.size - min(mini[-1], maxi[-1])
             nright = min(nright, int(self.size/2))
         out *= planck(out.size, nleft=nleft, nright=nright)
         return out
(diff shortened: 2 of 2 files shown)


duncanmmacleod:
@tjma12, thanks for this, can you please add/update the relevant automated test of this method to check that the new options are handled properly in myriad circumstances?

gwpy/timeseries/timeseries.py
```diff
@@ -1463,18 +1474,31 @@ def taper(self, side='leftright'):
         if side not in ('left', 'right', 'leftright'):
             raise ValueError("side must be one of 'left', 'right', "
                              "or 'leftright'")
+        if duration and nsamples:
+            raise ValueError("only one of 'duration' or 'nsamples' may be "
+                             "provided as arguments")
```
duncanmmacleod: I think I would handle all of argument parsing here, something like:

```python
        elif duration:
            nsamples = int(duration * self.sample_rate.to("Hz").value)
```

then you only need to consider `if nsamples` from that point forward.

tjma12:
@duncanmmacleod @alurban I was able to reduce the complexity a bit, but codeclimate is still unhappy. I'll follow your lead if you have suggestions on refactoring the method or if you want to ignore codeclimate in this case. I don't think it likes functions that have case switches constructed as if/else statements.

gwpy/timeseries/timeseries.py
tjma12: I'll give that a try, maybe it'll help with codeclimate

tjma12: This helped a lot with the cyclomatic complexity, thanks for the suggestion. The cognitive complexity is still showing up as high.

tjma12:
> @tjma12, thanks for this, can you please add/update the relevant automated test of this method to check that the new options are handled properly in myriad circumstances?

Done, I replicated the tests used for the auto-tapering but with user-specified arguments instead

tjma12:
I would personally lean towards leaving the functionality in since it's a minimal amount of code for added functionality. @duncanmmacleod can be the tie-breaker?

duncanmmacleod:
@tjma12, I believe I have corrected the build failures in #1151, please rebase this PR against the upstream `master` branch.

tjma12:
@duncanmmacleod @alurban Any thoughts on the codeclimate issue? I can try to refactor if necessary.

gwpy/timeseries/timeseries.py
```diff
@@ -1463,18 +1474,26 @@ def taper(self, side='leftright'):
...
+        elif duration:
```
alurban: @tjma12, instead of this `elif` block, it might appease Codeclimate if you use two pythonic shorthands:

```python
    duration = duration or 0
    nsamples = nsamples or int(duration * self.sample_rate.to("Hz").value)
```

The first line sets `duration` to 0 if it isn't passed, the second does the same for `nsamples`, and this plays nicely with the `if not nsamples` block on line 1487.

gwpy/timeseries/timeseries.py
```diff
@@ -1463,18 +1474,26 @@ def taper(self, side='leftright'):
...
         nleft, nright = 0, 0
-        mini, = signal.argrelmin(out.value)
-        maxi, = signal.argrelmax(out.value)
+        if not nsamples:
+            mini, = signal.argrelmin(out.value)
+            maxi, = signal.argrelmax(out.value)
         if 'left' in side:
-            nleft = max(mini[0], maxi[0])
-            nleft = min(nleft, self.size/2)
+            nleft = nsamples if nsamples else max(mini[0], maxi[0])
```
alurban: @tjma12, this can be simplified somewhat, which may appease Codeclimate:

```suggestion
            nleft = nsamples or max(mini[0], maxi[0])
```

gwpy/timeseries/timeseries.py
```diff
@@ -1463,18 +1474,26 @@ def taper(self, side='leftright'):
...
+            nleft = min(nleft, int(self.size/2))
         if 'right' in side:
-            nright = out.size - min(mini[-1], maxi[-1])
-            nright = min(nright, self.size/2)
+            nright = (nsamples if nsamples else
```
alurban: @tjma12, this can be simplified somewhat, which may appease Codeclimate:

```suggestion
            nright = (nsamples or
```

gwpy/timeseries/timeseries.py
```diff
@@ -1463,18 +1474,26 @@ def taper(self, side='leftright'):
         if side not in ('left', 'right', 'leftright'):
             raise ValueError("side must be one of 'left', 'right', "
                              "or 'leftright'")
+        if duration and nsamples:
```
alurban: @tjma12, if the other suggestions don't appease Codeclimate, then consider removing this `ValueError` exception and just warning the user in the docstring that if both arguments are passed then `nsamples` will override `duration`, which is true under the changes I suggested.

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -856,6 +856,22 @@ def test_taper(self):
...
+        utils.assert_allclose(data.value, numpy.cos(10*numpy.pi*t))
+
+        # run the same tests for a user-specified number of samples to taper
+        stapered = data.taper(nsamples=10)
+        assert stapered[0].value == 0
+        assert stapered[-1].value == 0
+        assert stapered.unit == data.unit
+        assert stapered.size == data.size
+        utils.assert_allclose(data.value, numpy.cos(10*numpy.pi*t))
+
```
alurban: @tjma12, if you decide to keep the use case exception in `TimeSeries.taper()`, could you include a separate unit test for it here? Something like:

```python
        with pytest.raises(ValueError) as exc:
            data.taper(duration=0.1, nsamples=10)
        assert str(exc.value) == ("only one of 'duration' or 'nsamples' "
                                  "may be provided as arguments")
```

gwpy/timeseries/timeseries.py
duncanmmacleod: I actually think the code as it exists now is easier to read, regardless of what codeclimate will say. I am also not convinced that this _will_ appease codeclimate, since there are still conditionals.

gwpy/timeseries/timeseries.py
duncanmmacleod: I agree, this can be achieved in the documentation.

tjma12: Removed this check in new commit

gwpy/timeseries/tests/test_timeseries.py
tjma12: Removed this check in new commit

gwpy/timeseries/timeseries.py
tjma12: I left the check on `duration` in place and noted in the docstring that it will override `nsamples` if both are provided

PR #1147, opened 2019-08-23
Add heterodyne method to TimeSeries class

### LINKED ISSUE #1146: Add heterodyne method to TimeSeries
I'd like to add a heterodyne method to the TimeSeries class. This would be very similar to the current demodulate method, but you would pass it a phase time series rather than a frequency, and it would output a complex time series rather than magnitudes and phases. I could then use this easily for the known pulsar searches (I'm trying to Python-ify the known pulsar search code.) I'll add a PR for this when I get the time.

mattpitkin:
This adds a method similar to the TimeSeries demodulate method, however, this takes in the phase evolution within which to demodulate/heterodyne the array. By default this returns a double-sided complex output, rather than the single-sided value.

This PR also adds testing of the method.

Closes #1146 

cc @duncanmmacleod

### DIFF
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -861,7 +861,7 @@ class TestTimeSeries(_TestTimeSeriesBase):
             data.heterodyne(phases[0:len(phases) // 2])
 
         # test with exp=True
-        het = data.heterodyne(phases, stride=stride, exp=True)
+        het = data.heterodyne(phases, stride=stride)
         assert het.unit == data.unit
         assert het.size == duration // stride
         utils.assert_allclose(numpy.abs(het.value), 0.5*amp, rtol=1e-4)
@@ -869,25 +869,22 @@ class TestTimeSeries(_TestTimeSeriesBase):
 
         # test with singlesided=True
         het = data.heterodyne(
-            phases, stride=stride, exp=True, singlesided=True
+            phases, stride=stride, singlesided=True
         )
         assert het.unit == data.unit
         assert het.size == duration // stride
         utils.assert_allclose(numpy.abs(het.value), amp, rtol=1e-4)
         utils.assert_allclose(numpy.angle(het.value), phase, rtol=2e-4)
 
-        # test with exp=False, deg=True
-        mag, ph = data.heterodyne(phases, stride=stride)
-        assert mag.unit == data.unit
-        assert mag.size == ph.size
-        assert ph.unit == 'deg'
-        utils.assert_allclose(mag.value, 0.5*amp, rtol=1e-4)
-        utils.assert_allclose(ph.value, numpy.rad2deg(phase), rtol=2e-4)
-
-        # test with exp=False, deg=False
-        mag, ph = data.heterodyne(phases, stride=stride, deg=False)
-        assert ph.unit == 'rad'
-        utils.assert_allclose(ph.value, phase, rtol=2e-4)
+        # test against demodulate for a fixed frequency
+        phasesfixed = 2*numpy.pi*(f*t)
+        data = TimeSeries(amp * numpy.cos(phasesfixed + phase),
+                          unit='', times=t)
+        het = data.heterodyne(
+            phases, stride=stride, singlesided=True
+        )
+        demod = data.demodulate(f, stride=stride, exp=True)
+        utils.assert_array_equal(het, demod)
 
     def test_taper(self):
         # create a flat timeseries, then taper it
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -868,7 +868,9 @@ class TestTimeSeries(_TestTimeSeriesBase):
         utils.assert_allclose(numpy.angle(het.value), phase, rtol=2e-4)
 
         # test with singlesided=True
-        het = data.heterodyne(phases, stride=stride, exp=True, singlesided=True)
+        het = data.heterodyne(
+            phases, stride=stride, exp=True, singlesided=True
+        )
         assert het.unit == data.unit
         assert het.size == duration // stride
         utils.assert_allclose(numpy.abs(het.value), amp, rtol=1e-4)
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -876,15 +876,15 @@ class TestTimeSeries(_TestTimeSeriesBase):
         utils.assert_allclose(numpy.abs(het.value), amp, rtol=1e-4)
         utils.assert_allclose(numpy.angle(het.value), phase, rtol=2e-4)
 
-        # test against demodulate for a fixed frequency
+        # test against demodulation for a fixed frequency
         phasesfixed = 2*numpy.pi*(f*t)
         data = TimeSeries(amp * numpy.cos(phasesfixed + phase),
                           unit='', times=t)
         het = data.heterodyne(
-            phases, stride=stride, singlesided=True
+            phasesfixed, stride=stride, singlesided=True
         )
         demod = data.demodulate(f, stride=stride, exp=True)
-        utils.assert_array_equal(het, demod)
+        utils.assert_allclose(het.value, demod.value, rtol=1e-9)
 
     def test_taper(self):
         # create a flat timeseries, then taper it
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -860,7 +860,7 @@ class TestTimeSeries(_TestTimeSeriesBase):
         with pytest.raises(ValueError):
             data.heterodyne(phases[0:len(phases) // 2])
 
-        # test with exp=True
+        # test with default settings
         het = data.heterodyne(phases, stride=stride)
         assert het.unit == data.unit
         assert het.size == duration // stride
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -842,6 +842,51 @@ class TestTimeSeries(_TestTimeSeriesBase):
         assert ph.unit == 'rad'
         utils.assert_allclose(ph.value, phase, rtol=1e-5)
 
+    def test_heterodyne(self):
+        # create a timeseries that is simply one loud sinusoidal oscillation,
+        # with a frequency and frequency derivative, then heterodyne using the
+        # phase evolution to recover the amplitude and phase
+        amp, phase, f, fdot = 1., numpy.pi/4, 30, 1e-4
+        duration, sample_rate, stride = 600, 4096, 60
+        t = numpy.linspace(0, duration, duration*sample_rate)
+        phases = 2*numpy.pi*(f*t + 0.5*fdot*t**2)
+        data = TimeSeries(amp * numpy.cos(phases + phase),
+                          unit='', times=t)
+
+        # test exceptions
+        with pytest.raises(TypeError):
+            data.heterodyne(1.0)
+
+        with pytest.raises(ValueError):
+            data.heterodyne(phases[0:len(phases) // 2])
+
+        # test with exp=True
+        het = data.heterodyne(phases, stride=stride, exp=True)
+        assert het.unit == data.unit
+        assert het.size == duration // stride
+        utils.assert_allclose(numpy.abs(het.value), 0.5*amp, rtol=1e-4)
+        utils.assert_allclose(numpy.angle(het.value), phase, rtol=2e-4)
+
+        # test with singlesided=True
+        het = data.heterodyne(phases, stride=stride, exp=True, singlesided=True)
+        assert het.unit == data.unit
+        assert het.size == duration // stride
+        utils.assert_allclose(numpy.abs(het.value), amp, rtol=1e-4)
+        utils.assert_allclose(numpy.angle(het.value), phase, rtol=2e-4)
+
+        # test with exp=False, deg=True
+        mag, ph = data.heterodyne(phases, stride=stride)
+        assert mag.unit == data.unit
+        assert mag.size == ph.size
+        assert ph.unit == 'deg'
+        utils.assert_allclose(mag.value, 0.5*amp, rtol=1e-4)
+        utils.assert_allclose(ph.value, numpy.rad2deg(phase), rtol=2e-4)
+
+        # test with exp=False, deg=False
+        mag, ph = data.heterodyne(phases, stride=stride, deg=False)
+        assert ph.unit == 'rad'
+        utils.assert_allclose(ph.value, phase, rtol=2e-4)
+
     def test_taper(self):
         # create a flat timeseries, then taper it
         t = numpy.linspace(0, 1, 2048)
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -876,16 +876,6 @@ class TestTimeSeries(_TestTimeSeriesBase):
         utils.assert_allclose(numpy.abs(het.value), amp, rtol=1e-4)
         utils.assert_allclose(numpy.angle(het.value), phase, rtol=2e-4)
 
-        # test against demodulation for a fixed frequency
-        phasesfixed = 2*numpy.pi*(f*t)
-        data = TimeSeries(amp * numpy.cos(phasesfixed + phase),
-                          unit='', times=t)
-        het = data.heterodyne(
-            phasesfixed, stride=stride, singlesided=True
-        )
-        demod = data.demodulate(f, stride=stride, exp=True)
-        utils.assert_allclose(het.value, demod.value, rtol=1e-9)
-
     def test_taper(self):
         # create a flat timeseries, then taper it
         t = numpy.linspace(0, 1, 2048)
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1411,6 +1411,80 @@ class TimeSeries(TimeSeriesBase):
         phase.override_unit('deg' if deg else 'rad')
         return (mag, phase)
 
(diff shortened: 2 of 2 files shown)


mattpitkin:
There seems to be some problem with the Python 3.6 build, but I don't think its specific to this PR. Maybe a Travis issue, or something in the build environment?

gwpy/timeseries/timeseries.py
```diff
@@ -1411,6 +1411,78 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
         phase.override_unit('deg' if deg else 'rad')
         return (mag, phase)
 
+    def heterodyne(self, phase, stride=1, exp=False, deg=True,
+                   singlesided=False):
+        """Compute the average magnitude and phase of this `TimeSeries`
```
alurban: @mattpitkin, if I'm reading your original feature request correctly, you want this method to always return a complex number rather than a `(mag, phase)` tuple, correct? So I think you can clean up the use case and docstring:

```python
    def heterodyne(self, phase, stride=1, singlesided=False):
        """Compute the average magnitude and phase of this `TimeSeries`
        once per stride after heterodyning with a given phase series

        Parameters
        ----------
        phase : `array_like`
            an array of phase measurements (radians) with which to heterodyne
            the signal

        stride : `float`, optional
            stride (seconds) between calculations, defaults to 1 second

        singlesided : `bool`, optional
            Boolean switch to return single-sided output (i.e., to multiply by 2
            so that the signal is distributed across positive frequencies only),
            default: False

        Returns
        -------
        out : `TimeSeries`
            magnitude and phase trends, represented as
            `mag * exp(1j*phase)` with `dt=stride`

        Notes
        -----
        This is similar to the :meth:`~gwpy.timeseries.TimeSeries.demodulate`
        method, but is more general in that it accepts a varying phase evolution,
        rather than a fixed frequency.

        Unlike :meth:`~gwpy.timeseries.TimeSeries.demodulate`, the complex
        output is double-sided by default, so is not multiplied by 2.

        See also
        --------
        TimeSeries.demodulate
            for a method to heterodyne at a fixed frequency
        """
        try:
            phaselen = len(phase)
        except Exception as exc:
            raise TypeError("Phase series is not array_like: {}".format(exc))
        if phaselen != len(self):
            raise ValueError(
                "Phase array must be the same length as the TimeSeries"
            )
        stridesamp = int(stride * self.sample_rate.value)
        nsteps = int(self.size // stridesamp)
        # stride through the TimeSeries and heterodyne
        out = type(self)(numpy.zeros(nsteps, dtype=complex))
        out.__array_finalize__(self)
        out.sample_rate = 1 / float(stride)
        phasearray = numpy.asarray(phase)  # make sure phase is a numpy array
        for step in range(nsteps):
            istart = int(stridesamp * step)
            iend = istart + stridesamp
            idx = numpy.arange(istart, iend)
            mixed = numpy.exp(-1j * phasearray[idx]) * self.value[idx]
            out.value[step] = 2 * mixed.mean() if singlesided else mixed.mean()
        return out
```

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -842,6 +842,53 @@ def test_demodulate(self):
         assert ph.unit == 'rad'
         utils.assert_allclose(ph.value, phase, rtol=1e-5)
 
+    def test_heterodyne(self):
```
alurban: This looks good, though if you use my other suggestion, the unit test can be cleaned up and simplified.

gwpy/timeseries/timeseries.py
mattpitkin: Thanks, your changes look good. I'd only ever be using it to output the complex time series rather than phases and magnitudes, and if anyone wants those quantities they are easy to generate.

gwpy/timeseries/tests/test_timeseries.py
mattpitkin: I'll implement your suggestion then clean up the test appropriately.

mattpitkin:
@alurban I'll look into adding a worked example. It shouldn't be too hard.

gwpy/timeseries/timeseries.py
mattpitkin: I've made the changes based on your suggestion

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -842,6 +842,50 @@ def test_demodulate(self):
...
+        # test with singlesided=True
+        het = data.heterodyne(
+            phases, stride=stride, singlesided=True
+        )
+        assert het.unit == data.unit
+        assert het.size == duration // stride
+        utils.assert_allclose(numpy.abs(het.value), amp, rtol=1e-4)
+        utils.assert_allclose(numpy.angle(het.value), phase, rtol=2e-4)
+
+        # test against demodulation for a fixed frequency
```
alurban: @mattpitkin, I think this block (lines 879 - 888) can safely be removed, since they're redundant with existing unit tests for `TimeSeries.demodulate`.

mattpitkin: You suggestions look good. I've made the changes.

mattpitkin:
@alurban I've now added an example. Its quite involved and uses a CW hardware injection in S6 data. It also requires lalsuite being installed - is this ok, as I'm guessing this might break the docs build if it tries to run the example?

One line, which contains a long URL, breaks the PEP8 line length, but I've added a '# noqa` to stop it raising it as a problem. Is this ok?

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,184 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
+        This is similar to the :meth:`~gwpy.timeseries.TimeSeries.demodulate`
+        method, but is more general in that it accepts a varying phase
+        evolution, rather than a fixed frequency.
+
+        Unlike :meth:`~gwpy.timeseries.TimeSeries.demodulate`, the complex
+        output is double-sided by default, so is not multiplied by 2.
+
+        Examples
+        --------
+        Heterodyning can be useful in examining quasi-monochromatic signals
```
alurban: @mattpitkin, thanks for putting together this example, however as you pointed out earlier there are a couple of issues:

* The example is rather complicated, and
* The example uses routines from LAL that are beyond the scope of gwpy's documentation

Is it possible to split the difference and think of either a simpler use case that doesn't depend on LAL? For example, maybe try constructing a toy model for fdot/f of the 60 Hz U.S. power grid line, then heterodyne some O2 data with that model? Alternatively, you could just inject a pulsar signal (using `TimeSeries.inject`) into O2 data and then recover it with `TimeSeries.heterodyne`?

mattpitkin: @alurban I've simplified the example to inject a sinusoid with a frequency that evolves with time. I then extract the signal's amplitude using `heterodyne` and compare it to demodulation at a fixed frequency.

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,143 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
+        >>> amp = 1.5e-22  # signal amplitude
+        >>> phase0 = 0.4  # signal phase at the phase epoch
+        >>> import numpy
+        >>> times = data.times.value - fepoch
+        >>> phase = 2 * numpy.pi * (f0 * times + 0.5 * fdot * times**2)
+        >>> signal = TimeSeries(amp * numpy.cos(phase + phase0),
+        >>>                     sample_rate=data.sample_rate, t0=data.t0)
+        >>> data = data.inject(signal)
+
+        We now bandpass the data around the signal frequency, and then
```
alurban: I propose a slight edit:

```python
        To recover the signal, we can bandpass the injected data around the
        signal frequency, then heterodyne using our phase model with a stride
        of 60 seconds:
```

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,143 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
+        heterodyne the `TimeSeries` using this phase evolution with a stride of
+        one minute:
+
+        >>> filtdata = data.bandpass(f0 - 0.5, f0 + 0.5)
+        >>> het = filtdata.heterodyne(phase, stride=60, singlesided=True)
+
+        Let's also demodulate the signal at a fixed freqeuncy for comparison
```
alurban: @mattpitkin, comparing to straight demodulation is slightly expensive for the docs and outside the scope of this example, so I propose we just get rid of `demod` and any references to it below.

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,143 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
+        We can then plot signal amplitude over time (cropping the first two
+        minutes to remove the filter response):
+
+        >>> combined = TimeSeriesDict()
+        >>> combined['Heterodyned'] = het.crop(het.x0.value + 180).abs()
+        >>> combined['Demodulated'] = demod.crop(demod.x0.value + 180).abs()
+        >>> plot = combined.plot()
+        >>> ax = plot.gca()
+        >>> ax.legend()
+        >>> ax.set_ylabel("Signal Strain Amplitude")
```
alurban: For consistency with other gwpy examples:

```suggestion
        >>> ax.set_ylabel("Strain amplitude")
```

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,143 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
+
+        >>> demod = filtdata.demodulate(f0, stride=60, exp=True)
+
+        We can then plot signal amplitude over time (cropping the first two
+        minutes to remove the filter response):
+
+        >>> combined = TimeSeriesDict()
+        >>> combined['Heterodyned'] = het.crop(het.x0.value + 180).abs()
+        >>> combined['Demodulated'] = demod.crop(demod.x0.value + 180).abs()
+        >>> plot = combined.plot()
```
alurban: @mattpitkin, based on the comment above, the previous three lines involving `combined` can be deleted, and this line replaced with:

```python
        >>> plot = het.plot()
```

The `ax.legend()` line below can also be deleted.

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,143 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
+        the Earth's rotational and orbital motion.
+
+        We can show an example of demodulation by creating a simulated signal
+        with a phase evolution described by a second order Taylor expansion in
+        phase, i.e., it is defined by a frequency and first frequency
+        derivative. We can download some LIGO Livingston data from GWOSC and
+        inject such a signal and attempt to recover the amplitude.
+
+        >>> from gwpy.timeseries import TimeSeries, TimeSeriesDict
+        >>> data = TimeSeries.fetch_open_data('L1', 1131350417, 1131357617)
```
alurban: @mattpitkin, is there a reason so much data needs to be read in? Can you just read in the 180 seconds you need?

```suggestion
        >>> data = TimeSeries.fetch_open_data('L1', 1131350417, 1131350587)
```

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,143 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
         >>> plot.show()
+
+        See also
+        --------
+        TimeSeries.heterodyne
+            for the underlying heterodyne detection method
         """
+        # stride through the TimeSeries and heterodyne at a single frequency
+        phase = (2 * numpy.pi * f *
+                 self.dt.decompose().value *
```
alurban: @mattpitkin, after thinking about it, the following change is needed to prevent a memory leak due to double-precision floats:

```suggestion
                 self.dt.astype('float32').decompose().value *
```

gwpy/timeseries/timeseries.py
mattpitkin: It doesn't need to be quite so much (I just copied from the demodulation example), but it does need to be enough to show the difference between the heterodyning and demodulation, which requires more than 180 seconds.

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,143 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
+        This is similar to the :meth:`~gwpy.timeseries.TimeSeries.demodulate`
+        method, but is more general in that it accepts a varying phase
+        evolution, rather than a fixed frequency.
+
+        Unlike :meth:`~gwpy.timeseries.TimeSeries.demodulate`, the complex
+        output is double-sided by default, so is not multiplied by 2.
+
+        Examples
+        --------
+        Heterodyning can be useful in examining quasi-monochromatic signals
```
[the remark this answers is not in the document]
mattpitkin: Thanks. I've made this change.

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,143 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
+
+        We now need to set the signal parameters, generate the expected
+        phase evolution, and create the signal:
+
+        >>> f0 = 123.456789  # signal frequency (Hz)
+        >>> fdot = -9.87654321e-7  # signal frequency derivative (Hz/s)
+        >>> fpeoch = 1131350417  # phase epoch
+        >>> amp = 1.5e-22  # signal amplitude
+        >>> phase0 = 0.4  # signal phase at the phase epoch
+        >>> import numpy
```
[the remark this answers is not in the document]
mattpitkin: Done.

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,143 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
+        Heterodyning can be useful in examining quasi-monochromatic signals
+        with a known phase evolution. Examples of such signals are the
+        `continuous-wave hardware injections
+        <https://www.gw-openscience.org/s6hwcw/>`_ in many of the LIGO
+        observing runs, which are designed to mimic the expected signals
+        from rapidly rotating neutron stars. For these sources the signal's
+        frequency slowly decreases and is also Doppler modulated due to
+        the Earth's rotational and orbital motion.
+
+        We can show an example of demodulation by creating a simulated signal
```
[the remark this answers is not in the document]
mattpitkin: Done.

gwpy/timeseries/timeseries.py
mattpitkin: I was working backwards through these, so have now seen you comment to remove the demodulation comparison.

mattpitkin: I've not changed it just 180 seconds, but have halved the amount of data.

gwpy/timeseries/timeseries.py
mattpitkin: Done.

gwpy/timeseries/timeseries.py
mattpitkin: Done.

gwpy/timeseries/timeseries.py
mattpitkin: Ok, done.

gwpy/timeseries/timeseries.py
mattpitkin: Done.

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -842,6 +842,40 @@ def test_demodulate(self):
         assert ph.unit == 'rad'
```
alurban: @mattpitkin, because of the switch to 32-bit floats, there is some loss in precision when calculating the phase of a demodulated signal. I think the loss is acceptable, but the unit tests for `TimeSeries.demodulate` will have to be updated.

With apologies for the extra hassle, can you make the following changes in `test_timeseries.py::TestTimeSeries::test_demodulate`?

* Line 830: change `rtol` to `1e-2`
* Line 838: change `rtol` to `1e-2`
* Line 843: change `rtol` to `1e-2`

Note, if you have the `pytest` package installed then you can run this unit test on the command-line to check that it behaves the way you want:

```bash
$ python -m pytest gwpy/timeseries/tests/test_timeseries.py::TestTimeSeries::test_demodulate
```

gwpy/timeseries/timeseries.py
```diff
@@ -1388,28 +1388,132 @@ def demodulate(self, f, stride=1, exp=False, deg=True):
...
         >>> plot.show()
+
+        See also
+        --------
+        TimeSeries.heterodyne
+            for the underlying heterodyne detection method
         """
+        # stride through the TimeSeries and heterodyne at a single frequency
+        phase = (2 * numpy.pi * f *
+                 self.dt.astype('float32').decompose().value *
```
alurban: @mattpitkin, I was wrong before, apologies about that, can you remove the the conversion to `float32`?

```suggestion
                 self.dt.decompose().value *
```

**Context:** casting to `float32` causes a loss of precision, which breaks the unit tests for `TimeSeries.demodulate`. More to the point, the loss of precision is unacceptable since it leads to a systematic error in phase at the level of ~1 degree, which is on the same order as the current calibration uncertainty for advanced LIGO.

I was able to profile this method before and after your changes; before the change, demodulating 12 hours of 4 kHz data takes about 4.5 GB, while the same operation after your changes takes about 6.5 GB. I believe this is an acceptable trade-off to get the precision necessary to demodulate advanced LIGO's calibration lines, which is a mission-critical downstream use of this method.

alurban:
Thanks @mattpitkin, I think this PR is ready.

@duncanmmacleod, Matt and I have iterated a few times and I'm more than happy with the proposed changes, so I now pass the baton to you for final approval and merging.

duncanmmacleod:
@mattpitkin, I believe I have corrected the build failures in #1151, please rebase this PR against the upstream `master` branch.

mattpitkin:
@duncanmmacleod sure, it's been rebased with master now.

duncanmmacleod:
Thanks @mattpitkin

PR #1148, opened 2019-08-23
corrected LIGOTimeGPS attribute call

tjma12:
In `segments/flag.py`,  there are cases of `LIGOTimeGPS` objects attempting to access the `LIGOTimeGPS.seconds` attribute, which doesn't exist, the correct attribute should be `.gpsSeconds`. These are invoked in cases where the end time of a flag is set to `+inf`, such as when populating a `DataQualityDict` from a veto definer. I was able to successfully populate a `DataQualityDict` with `+inf` end times locally after this fix.

### DIFF
--- a/gwpy/segments/flag.py
+++ b/gwpy/segments/flag.py
@@ -494,7 +494,11 @@ class DataQualityFlag(object):
         for start, end in qsegs:
             # handle infinities
             if float(end) == +inf:
+<<<<<<< HEAD
                 end = to_gps('now').gpsSeconds
+=======
+                end = int(to_gps('now'))
+>>>>>>> 4987e35509a2121b3dca6886ccbfe63bb58cf41a
 
             # query
             try:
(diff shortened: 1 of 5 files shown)


gwpy/segments/flag.py
```diff
@@ -494,7 +494,7 @@ def query_dqsegdb(cls, flag, *args, **kwargs):
         for start, end in qsegs:
             # handle infinities
             if float(end) == +inf:
-                end = to_gps('now').seconds
+                end = to_gps('now').gpsSeconds
```
duncanmmacleod: Can you implement this change to make the operation object agnostic?

```suggestion
                end = int(to_gps('now'))
```

gwpy/segments/flag.py
```diff
@@ -1152,7 +1152,7 @@ def query_segdb(cls, flags, *args, **kwargs):
                 vers = dqflag.version
             for gpsstart, gpsend in qsegs:
                 if float(gpsend) == +inf:
-                    gpsend = to_gps('now').seconds
+                    gpsend = to_gps('now').gpsSeconds
```
duncanmmacleod:
```suggestion
                    gpsend = int(to_gps('now'))
```

tjma12:
@duncanmmacleod Thanks for the suggestion, that makes sense to me, updated.

duncanmmacleod:
@tjma12, I believe I have corrected the build failures in #1151, please rebase this PR against the upstream `master` branch.

PR #1152, opened 2019-09-12
Add exception for catching in IO

mattpitkin:
I'm attempting to use `TimeSeriesBase` class as the basis for a data object in my new [CWInPy](https://github.com/cwinpy/cwinpy) code. I've written a new IO reader registry for the data format. However, an allowed data format for reading is an acsii txt file that has four columns. In these cases when the `io.is_cache` function checks these sort of files it causes a `RuntimeError` exception. However, this is not an exception that is caught, so this `is_cache` function doesn't just exit as `False`, but instead fails with an exception. This PR just adds `RuntimeError` as an exception that means that `is_cache` does return `False` in this case.

### DIFF
--- a/gwpy/io/cache.py
+++ b/gwpy/io/cache.py
@@ -279,7 +279,13 @@ def is_cache(cache):
     if isinstance(cache, string_types + FILE_LIKE):
         try:
             return bool(len(read_cache(cache)))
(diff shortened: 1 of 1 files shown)


mattpitkin:
Actually, it looks like something else must have been causing the error and this PR is not needed (at least for now), so I'll close this.

PR #1153, opened 2019-09-14
add optional segment ribbons to gwpy-plot when x-axis is time

areeda:
There are two new command line options 
 * `--std-seg` which adds a predefined list of DQ segment ribbons see: https://ldvw.ligo.caltech.edu/ldvw/view?act=getImg&imgId=254114
* `--seg <segment name:version> [...]` add one or more segments by full name. See: https://ldvw.ligo.caltech.edu/ldvw/view?act=getImg&imgId=254112

Next version of ldvw implements the `--std-seg` option.  A subsequent version will have the ability to select which segments are displayed.

### DIFF
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -732,7 +732,8 @@ class CliProduct(object):
             self.plot_num += 1
 
     def add_segs(self, args):
-        """ If requested add DQ segments"""
+        """ If requested add DQ segments
+        """
         std_segments = [
                 '{ifo}:DMT-GRD_ISC_LOCK_NOMINAL:1',
                 '{ifo}:DMT-DC_READOUT_LOCKED:1',
@@ -764,14 +765,9 @@ class CliProduct(object):
 
             for segment in segments:
                 seg_name = segment.replace('{ifo}', ifo)
-                try:
-                    seg_data = DataQualityFlag. \
-                        query_dqsegdb(seg_name, start, end,
-                                      url='https://segments.ligo.org')
-                except:             # noqa: E722
-                    seg_data = DataQualityFlag. \
-                        query_dqsegdb(seg_name, start, end,
-                                      url='https://segments-backup.ligo.org/')
+                seg_data = DataQualityFlag.query_dqsegdb(
+                        seg_name, start, end)
+
                 self.plot.add_segments_bar(seg_data, label=seg_name)
 
 
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -48,6 +48,7 @@ from ..plot.gps import (GPS_SCALES, GPSTransform)
 from ..plot.tex import label_to_latex
 from ..segments import DataQualityFlag
 
+
 __author__ = 'Joseph Areeda <joseph.areeda@ligo.org>'
 
 BAD_UNITS = {'*', 'Counts.', }
@@ -752,6 +753,7 @@ class CliProduct(object):
             ifo = m.group(1) if m else '?:'
 
             start = None
+            end = 0
             for ts in self.timeseries:
                 if start:
                     start = min(ts.t0, start)
@@ -946,7 +948,7 @@ class TimeDomainProduct(CliProduct):
             args.xmin = min(starts)
         if args.epoch is None:
             args.epoch = args.xmin
-        elif (args.epoch < 1e8):
+        elif args.epoch < 1e8:
             args.epoch += min(starts)
 
         if args.xmax is None:
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -46,7 +46,7 @@ from ..time import to_gps
 from ..timeseries import TimeSeriesDict
 from ..plot.gps import (GPS_SCALES, GPSTransform)
 from ..plot.tex import label_to_latex
-from ..segments import DataQualityFlag   # noqa: E402
+from ..segments import DataQualityFlag
 
 __author__ = 'Joseph Areeda <joseph.areeda@ligo.org>'
 
@@ -732,8 +732,7 @@ class CliProduct(object):
 
     def add_segs(self, args):
         """ If requested add DQ segments"""
-        std_segments = \
-            [
+        std_segments = [
                 '{ifo}:DMT-GRD_ISC_LOCK_NOMINAL:1',
                 '{ifo}:DMT-DC_READOUT_LOCKED:1',
                 '{ifo}:DMT-CALIBRATED:1',
@@ -750,9 +749,8 @@ class CliProduct(object):
 
             chan = args.chan[0][0]
             m = re.match('^([A-Za-z][0-9]):', chan)
-            ifo = '?:'
-            if m:
-                ifo = m.group(1)
+            ifo = m.group(1) if m else '?:'
+
             start = None
             for ts in self.timeseries:
                 if start:
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -185,6 +185,10 @@ class CliProduct(object):
         #: channels to load
         self.chan_list = unique(c for clist in args.chan for c in clist)
 
+        # use reduced as an alias for rds in channel name
+        for idx in range(len(self.chan_list)):
+            self.chan_list[idx] = self.chan_list[idx].replace('reduced', 'rds')
+
         # total number of datasets that _should_ be acquired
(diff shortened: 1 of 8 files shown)


duncanmmacleod:
@areeda, I believe I have corrected the build failures in #1151, please rebase this PR against the upstream `master` branch.

gwpy/cli/cliproduct.py
```diff
@@ -46,6 +46,7 @@
 from ..timeseries import TimeSeriesDict
 from ..plot.gps import (GPS_SCALES, GPSTransform)
 from ..plot.tex import label_to_latex
+from ..segments import DataQualityFlag   # noqa: E402
```
duncanmmacleod: I don't think the `noqa` is needed here.

```suggestion
from ..segments import DataQualityFlag
```

gwpy/cli/cliproduct.py
```diff
@@ -724,6 +730,50 @@ def run(self):
                 show_error = False
             self.plot_num += 1
 
+    def add_segs(self, args):
+        """ If requested add DQ segments"""
```
duncanmmacleod:
```suggestion
        """If requested add DQ segments
        """
```

gwpy/cli/cliproduct.py
```diff
@@ -724,6 +730,50 @@ def run(self):
...
+        std_segments = \
+            [
+                '{ifo}:DMT-GRD_ISC_LOCK_NOMINAL:1',
+                '{ifo}:DMT-DC_READOUT_LOCKED:1',
+                '{ifo}:DMT-CALIBRATED:1',
+                '{ifo}:DMT-ANALYSIS_READY:1'
+            ]
```
duncanmmacleod: Can you please reformat this as
```python
        std_segments = [
            '{ifo}:DMT-GRD_ISC_LOCK_NOMINAL:1',
            '{ifo}:DMT-DC_READOUT_LOCKED:1',
            '{ifo}:DMT-CALIBRATED:1',
            '{ifo}:DMT-ANALYSIS_READY:1'
        ]

File: gwpy/cli/cliproduct.py [hunk 4]
```diff
@@ -724,6 +730,50 @@ def run(self):
...
+                    start = ts.t0
+                    end = start + ts.duration
+
+            for segment in segments:
+                seg_name = segment.replace('{ifo}', ifo)
+                try:
+                    seg_data = DataQualityFlag. \
+                        query_dqsegdb(seg_name, start, end,
+                                      url='https://segments.ligo.org')
+                except:             # noqa: E722
```
duncanmmacleod: Can you please itemise the exceptions that we don't care about here, rather than leaving a bare `except`?

gwpy/cli/cliproduct.py
```diff
@@ -724,6 +730,50 @@ def run(self):
...
+            if args.seg:
+                for seg in args.seg:
+                    # NB: args.seg may be list of lists
+                    segments += seg
+
+            chan = args.chan[0][0]
+            m = re.match('^([A-Za-z][0-9]):', chan)
+            ifo = '?:'
+            if m:
+                ifo = m.group(1)
```
duncanmmacleod: this may be simplified as
```python
            ifo = m.group(1) if m else '?:'
```
but if you think that is less readable I'm not bothered.

gwpy/cli/cliproduct.py
areeda: The point is we don't care why. If segments fail we try segments-backup.
Is there a way to get a complete list of possible exceptions?

gwpy/cli/cliproduct.py
areeda: if the match fails m=None and m.<anything> throws an exception. N'est-ce pas?

gwpy/cli/cliproduct.py
areeda: ok

gwpy/cli/cliproduct.py
areeda: I thing the early set matplotlib back end is why but I'll confirm.  It was pasted.

areeda: You're right.  fixed.

gwpy/cli/cliproduct.py
areeda: If segmemts-backup also fails we report that error.
I'd prefer to leave it unless we can get a list of exceptions.

gwpy/cli/cliproduct.py
```diff
@@ -724,6 +731,49 @@ def run(self):
...
+                else:
+                    start = ts.t0
+                    end = start + ts.duration
+
+            for segment in segments:
+                seg_name = segment.replace('{ifo}', ifo)
+                try:
+                    seg_data = DataQualityFlag. \
+                        query_dqsegdb(seg_name, start, end,
+                                      url='https://segments.ligo.org')
```
duncanmmacleod: Can you please restructure this as

```python
                    seg_data = DataQualityFlag.query_dqsegdb(
                        seg_name,
                        start,
                        end,
                    )
```
and do not include a hardcoded `url` argument, the method itself has logic to auto-select the default server which allows the user to optimise their selection across the board.

gwpy/cli/cliproduct.py
duncanmmacleod: I presume we only care about failed connections, and not missing flags etc. in that case we should probably use `except URLError` or similar.

gwpy/cli/cliproduct.py
areeda: done

gwpy/cli/cliproduct.py
areeda: We can do that but isn't it possible that backup succeeds when segments fail for other than connection problems?  If the error is unrecoverable won't we get the same error from backup?  In my mind it only costs one more connection on the rare error.

gwpy/cli/cliproduct.py
areeda: that does solve the bare except problem

gwpy/cli/cliproduct.py
duncanmmacleod: Ping @areeda, I don't see this change.

duncanmmacleod:
@areeda, I can't help but notice that this PR contains commits that were implemented in other PRs. I presume this happened when you rebased (merged) the upstream master branch. Just a reminder that the best way to rebase is (IMO):

```
# presuming your remote for https://github.com/gwpy/gwpy.git is called 'upstream'
git fetch upstream
git pull --rebase upstream master
git push --force
```
<sub>[standard disclaimer, don't `push --force` without checking things.]</sub>

which should result in a clean update with no merge commits.

gwpy/cli/cliproduct.py
areeda: a) I don't understand why I don't see email on my PRs including this ping while I seem to get them for everybody else's??

b) This change shows up at https://github.com/areeda/gwpy/blob/segment-add/gwpy/cli/cliproduct.py#L737 I wonder why you didn't see it??

gwpy/cli/cliproduct.py
areeda: This is resolved by your suggestion to remove segdb urls from this level

gwpy/cli/cliproduct.py
duncanmmacleod: @areeda, the link you posted includes identical code to what is currently on the branch, what I want is this:

```diff
diff --git a/gwpy/cli/cliproduct.py b/gwpy/cli/cliproduct.py
index cbbd350e..f237e7fa 100644
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -735,11 +735,11 @@ class CliProduct(object):
         """ If requested add DQ segments
         """
         std_segments = [
-                '{ifo}:DMT-GRD_ISC_LOCK_NOMINAL:1',
-                '{ifo}:DMT-DC_READOUT_LOCKED:1',
-                '{ifo}:DMT-CALIBRATED:1',
-                '{ifo}:DMT-ANALYSIS_READY:1'
-            ]
+            '{ifo}:DMT-GRD_ISC_LOCK_NOMINAL:1',
+            '{ifo}:DMT-DC_READOUT_LOCKED:1',
+            '{ifo}:DMT-CALIBRATED:1',
+            '{ifo}:DMT-ANALYSIS_READY:1'
+        ]
         segments = list()
         if hasattr(args, 'std_seg'):
             if args.std_seg:
```

areeda: I see you want indentation changed.  How's this one?

BTW I got an email with your comment did you change something?

PR #1154, opened 2019-09-16
Raise exception on a corner case in Q-transforms

### LINKED ISSUE #1144: qtransform 'out' variable referenced before assignment
The [QTiling.transform()](https://github.com/gwpy/gwpy/blob/ac2d494d2045ac0afdbfda12486087a7e305db5d/gwpy/signal/qtransform.py#L166) assigns a variable 'out' in an 'if' statement. This can lead to the following error:

`UnboundLocalError: local variable 'out' referenced before assignment`

Example stacktrace:
```
 qax.imshow(data.q_transform())

  File "/home/avi.vajpeyi/.conda/envs/timeshifts/lib/python3.7/site-packages/gwpy/timeseries/timeseries.py", line 2108, in     q_transform
    mismatch=mismatch, search=search)

  File "/home/avi.vajpeyi/.conda/envs/timeshifts/lib/python3.7/site-packages/gwpy/signal/qtransform.py", line 686, in q_scan
    frange=frange).transform(data, **kwargs)

  File "/home/avi.vajpeyi/.conda/envs/timeshifts/lib/python3.7/site-packages/gwpy/signal/qtransform.py", line 202, in transform
    return (out, nind * weight / nplanes)

UnboundLocalError: local variable 'out' referenced before assignment

```
Maybe `out` should be initialised to some value? Or is it the case that if `out` is not set, then an exception should be raised (because `out` should always be set in the if statement)?

alurban:
This fixes #1144.

cc @duncanmmacleod

### DIFF
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -189,6 +189,8 @@ class QTiling(QObject):
         QPlane.transform
             compute the Q-transform over a single time-frequency plane
         """
+        if not numpy.isfinite(fseries).all():
+            raise ValueError('Input signal contains non-numerical values')
         weight = 1 + numpy.log10(self.qrange[1]/self.qrange[0]) / numpy.sqrt(2)
         nind, nplanes, peak, result = (0, 0, 0, None)
         # identify the plane with the loudest tile
@@ -199,12 +201,7 @@ class QTiling(QObject):
             if result.peak['energy'] > peak:
                 out = result
                 peak = out.peak['energy']
-        try:
-            return (out, nind * weight / nplanes)
-        except UnboundLocalError:
-            raise ValueError(
-                'No Q-planes were found, likely because the input timeseries '
-                'record is too short, please load more data and try again')
+        return (out, nind * weight / nplanes)
 
 
 class QPlane(QBase):
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -199,7 +199,12 @@ class QTiling(QObject):
             if result.peak['energy'] > peak:
                 out = result
                 peak = out.peak['energy']
-        return (out, nind * weight / nplanes)
+        try:
+            return (out, nind * weight / nplanes)
+        except UnboundLocalError:
+            raise ValueError(
+                'No Q-planes were found, likely because the input timeseries '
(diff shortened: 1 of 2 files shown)


duncanmmacleod:
@alurban, I believe I have corrected the build failures in #1151, please rebase this PR against the upstream `master` branch.

alurban:
Thanks @duncanmmacleod, I've rebased this against `upstream/master`.

duncanmmacleod:
@alurban, the test failure is real, `numpy.full_like` is new in numpy-1.9.0 (I think), so we can't use it if we want to support python2.7 on epel7.

alurban:
@duncanmmacleod, I noticed that just before you did, the PR has been updated to not depend on this utility.

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -1052,6 +1052,13 @@ def test_q_transform_logf(self, losc):
         assert qspecgram.q == 5.65685424949238
         nptest.assert_almost_equal(qspecgram.value.max(), 155.93774, decimal=5)
 
+    def test_q_transform_nan(self, losc):
+        data = TimeSeries([numpy.nan for y in losc.value])
+        data.__array_finalize__(losc)
```
duncanmmacleod: I would just create a new `TimeSeries` of a fixed size for this test, rather than pulling in any real data

```python
        data = TimeSeries(numpy.empty(<N>) * numpy.nan, ...)
```

alurban: @duncanmmacleod, good call, this is done in the latest commit.

gwpy/signal/qtransform.py
```diff
@@ -189,6 +189,8 @@ def transform(self, fseries, **kwargs):
         QPlane.transform
             compute the Q-transform over a single time-frequency plane
         """
+        if not numpy.isfinite(fseries).all():
```
duncanmmacleod: I think it is more intuitive to have a positive statement here:
```suggestion
        if numpy.isinf(fseries).any():
```

alurban: @duncanmmacleod, I would agree, however `numpy.isinf` doesn't account for NaNs, which is the most common corner case this block is trying to catch.

PR #1156, opened 2019-09-25
Require that Q-planes have strictly increasing frequency

### LINKED ISSUE #1155: Occasional duplicate frequencies on Q-planes
Occasionally, when computing Q-transforms of data streams down to very low frequency (~0.01 Hz), the floor division in `QPlane._iter_frequencies` results in duplicate frequencies on some fixed-Q planes. A fix for this is on its way imminently.

This issue was originally reported by @dethodav after running `gwdetchar-omega` with the following configuration:

```ini
[PEM]
name = Physical Environment Monitoring
q-range = 3.3166,50.0
frequency-range = 0.03,inf
frametype = L1_R
duration = 512
fftlength = 8
max-mismatch = 0.35
snr-threshold = 5.5
always-plot = True
plot-time-durations = 1,4,16
channels = L1:PEM-EX_SEIS_VEA_FLOOR_QUAD_SUM_DQ
```

Since this PEM channel is low-sample-rate, the very low frequency behavior is meaningful, and needs to be analyzable by omega scans.

alurban:
This PR enforces monotonic-increasing frequency in all `QPlane` objects by stripping off duplicate frequencies, while preserving the order. This resolves an issue where duplicate frequencies cause `scipy.interpolate.interp2d` to return all NaNs, for a demonstration see [**this omega scan**](https://ldas-jobs.ligo.caltech.edu/~alexander.urban/wdq/L1_1253318420.0/) (requires `LIGO.ORG` credentials).

An extra unit test checking monotonicity is included.

This fixes #1155.

cc @duncanmmacleod, @dethodav

### DIFF
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -27,7 +27,6 @@ from __future__ import division
 
 import warnings
 from math import (log, ceil, pi, isinf, exp)
-from more_itertools import unique_everseen
 
 from six import string_types
 from six.moves import xrange
@@ -243,8 +242,9 @@ class QPlane(QBase):
 
         Yields a `QTile` at each frequency
         """
+        from ..cli.cliproduct import unique
         # for each frequency, yield a QTile
-        for freq in unique_everseen(self._iter_frequencies()):
+        for freq in unique(self._iter_frequencies()):
             yield QTile(self.q, freq, self.duration, self.sampling,
                         mismatch=self.mismatch)
 
@@ -269,7 +269,8 @@ class QPlane(QBase):
 
         :type: `numpy.ndarray`
         """
-        return numpy.array(list(unique_everseen(self._iter_frequencies())))
+        from ..cli.cliproduct import unique
+        return numpy.array(list(unique(self._iter_frequencies())))
 
     @property
     def farray(self):
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -27,6 +27,7 @@ from __future__ import division
 
 import warnings
 from math import (log, ceil, pi, isinf, exp)
+from more_itertools import unique_everseen
 
 from six import string_types
 from six.moves import xrange
@@ -243,7 +244,7 @@ class QPlane(QBase):
         Yields a `QTile` at each frequency
         """
         # for each frequency, yield a QTile
-        for freq in self._iter_frequencies():
+        for freq in unique_everseen(self._iter_frequencies()):
             yield QTile(self.q, freq, self.duration, self.sampling,
                         mismatch=self.mismatch)
 
@@ -268,7 +269,7 @@ class QPlane(QBase):
 
         :type: `numpy.ndarray`
         """
-        return numpy.array(list(self._iter_frequencies()))
+        return numpy.array(list(unique_everseen(self._iter_frequencies())))
 
     @property
     def farray(self):
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -34,6 +34,7 @@ from six.moves import xrange
 import numpy
 from numpy import fft as npfft
 
+from ..utils import unique
 from ..segments import Segment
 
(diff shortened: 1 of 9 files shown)


gwpy/signal/qtransform.py
```diff
@@ -27,6 +27,7 @@
 
 import warnings
 from math import (log, ceil, pi, isinf, exp)
+from more_itertools import unique_everseen
```
duncanmmacleod: `more_itertools` is not a standard library package, and is currently not declared as a dependency of gwpy, however...

gwpy/signal/qtransform.py
```diff
@@ -243,7 +244,7 @@ def __iter__(self):
         Yields a `QTile` at each frequency
         """
         # for each frequency, yield a QTile
-        for freq in self._iter_frequencies():
+        for freq in unique_everseen(self._iter_frequencies()):
```
duncanmmacleod: The same result can be achieved using something like this with no extra dependencies:

https://github.com/gwpy/gwpy/blob/v0.15.0/gwpy/cli/cliproduct.py#L94-L102

alurban: @duncanmmacleod, good call, I've patched this to import `unique` from `gwpy.cli.cliproduct` (which I wasn't aware we had til you pointed it out). Note there is one small catch: a module-level import would be circular, so the import must be done from inside the two class methods that use this function.

gwpy/signal/tests/test_qtransform.py
```diff
@@ -53,6 +53,12 @@ def test_far():
     assert FAR < 1 / (1.37e10 * 365 * 86400)
 
 
+def test_monotonicity():
+    # test that Q-plane frequencies are monotonic increasing
```
duncanmmacleod:
```suggestion
    # test that Q-plane frequencies are strictly increasing
```

gwpy/signal/tests/test_qtransform.py
```diff
@@ -53,6 +53,12 @@ def test_far():
...
+    freq = QGRAM.plane.frequencies
+    assert all(x < y for x, y in zip(freq, freq[1:]))
```
duncanmmacleod:
```suggestion
    assert (freq[1:] > freq[:-1]).all()
```

alurban:
@duncanmmacleod, I've addressed your second round of comments, all your suggestions were sensible to me.

To answer your question: initially I did try to force `QPlane._iter_frequencies` to return unique elements, unfortunately there's no good way to guarantee that this happens. The issue is floor division, no matter what you do, when the frequency spacing is small enough there will always be duplicate frequencies because floor division will return the same integer for multiple use cases. And because this method `yield`s one element at a time, there's no way to strip duplicates during the method call itself, so I decided to do it after-the-fact by feeding it to `gwpy.utils.misc.unique`.

duncanmmacleod:
How about something like this:

```patch
diff --git a/gwpy/signal/qtransform.py b/gwpy/signal/qtransform.py
index ee78f416..5af40ff6 100644
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -255,10 +255,15 @@ class QPlane(QBase):
         fstep = fcum_mismatch / nfreq
         fstepmin = 1 / self.duration
         # for each frequency, yield a QTile
+        last = None
         for i in xrange(nfreq):
-            yield (minf *
-                   exp(2 / (2 + self.q**2)**(1/2.) * (i + .5) * fstep) //
-                   fstepmin * fstepmin)
+            this =  (
+                minf * exp(2 / (2 + self.q**2)**(1/2.) * (i + .5) * fstep) //
+                fstepmin * fstepmin
+            )
+            if this != last:  # only keep uniq elements
+                yield this
+            last = this

     @property
     def frequencies(self):
```

alurban:
@duncanmmacleod, I've given your suggestion a try and it worked, so I've stuck the `unique` utility back where I found it and went with your less-invasive suggestion.

PR #1158, opened 2019-09-27
Introduce a round_to_power utility

alurban:
This PR introduces a new utility, `gwpy.utils.misc.round_to_power`, which accepts any positive scalar (`float`, `int`, `~astropy.units.Quantity`, etc.) and rounds (in log space) to the nearest integer power of a given base. The output has the same type as the base, and users can force the function to round up or down if they desire.

The new utility replaces and generalizes an existing one in `gwpy.signal.qtransform`.

cc @duncanmmacleod

### DIFF
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -289,7 +289,8 @@ class QPlane(QBase):
     def whitening_duration(self):
         """The recommended data duration required for whitening
         """
-        return round_to_power(self.q / (2 * self.frange[0]))
+        return round_to_power(self.q / (2 * self.frange[0]),
+                              base=2, which=None)
 
     def transform(self, fseries, norm=True, epoch=None, search=None):
         """Calculate the energy `TimeSeries` for the given `fseries`
@@ -355,7 +356,8 @@ class QTile(QBase):
         :type: `int`
         """
         tcum_mismatch = self.duration * 2 * pi * self.frequency / self.q
-        return round_to_power(tcum_mismatch / self.deltam, which='upper')
+        return round_to_power(tcum_mismatch / self.deltam,
+                              base=2, which='upper')
 
     @property
     def windowsize(self):
--- a/gwpy/signal/qtransform.py
+++ b/gwpy/signal/qtransform.py
@@ -34,6 +34,7 @@ from six.moves import xrange
 import numpy
 from numpy import fft as npfft
 
+from ..utils import round_to_power
 from ..segments import Segment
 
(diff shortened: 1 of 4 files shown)


gwpy/signal/qtransform.py
```diff
@@ -288,7 +289,7 @@ def farray(self):
     def whitening_duration(self):
         """The recommended data duration required for whitening
         """
-        return 2 ** (round(log(self.q / (2 * self.frange[0]), 2)))
+        return round_to_power(self.q / (2 * self.frange[0]))
```
duncanmmacleod: I think passing the keywords might make the usage a little clearer in this case.

```suggestion
        return round_to_power(self.q / (2 * self.frange[0]), base=2, which=None)
```

gwpy/signal/qtransform.py
```diff
@@ -354,7 +355,7 @@ def ntiles(self):
         :type: `int`
         """
         tcum_mismatch = self.duration * 2 * pi * self.frequency / self.q
-        return next_power_of_two(tcum_mismatch / self.deltam)
+        return round_to_power(tcum_mismatch / self.deltam, which='upper')
```
duncanmmacleod:
```suggestion
        return round_to_power(tcum_mismatch / self.deltam, base=2, which='upper')
```

gwpy/signal/qtransform.py
alurban: @duncanmmacleod, this is done in the latest commit

gwpy/signal/qtransform.py
alurban: Done in the latest commit

gwpy/utils/tests/test_misc.py
```diff
@@ -47,6 +47,25 @@ def test_null_context():
...
+    assert utils_misc.round_to_power(2) == 2
+    assert utils_misc.round_to_power(9, base=10) == 10
+    assert utils_misc.round_to_power(5, which='lower') == 4
+    assert utils_misc.round_to_power(5, which='upper') == 8
+    # test output
+    base = 10.
+    rounded = utils_misc.round_to_power(9, base=base)
+    assert base == rounded
+    assert type(base) == type(rounded)
+    with pytest.raises(ValueError) as exc:
```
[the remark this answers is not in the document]
alurban: Done in the latest commit

alurban:
Thanks for reviewing @duncanmmacleod, I've implemented your suggested changes in the latest commit.

PR #1159, opened 2019-10-01
Improved support for pad in TimeSeries.read and TimeSeries.find

duncanmmacleod:
This PR improves the support for padding gaps in data read using `.find()` or `.read()`. The main changes are:

- borrowed logic from the nds2 reader to ensure that all series are returned with the requested boundaries, padding the start or end as required
- allow gaps in datafind queries if `pad` is given

This should allow users to read, for example, months of trend data directly from GWF even if gaps are present.

### DIFF
--- a/gwpy/timeseries/io/core.py
+++ b/gwpy/timeseries/io/core.py
@@ -90,13 +90,11 @@ def _pad_series(ts, pad, start=None, end=None):
     """
     span = ts.span
     if start is None:
-        pada = 0
-    else:
-        pada = max(int((span[0] - start) * ts.sample_rate.value), 0)
+        start = span[0]
     if end is None:
-        padb = 0
-    else:
-        padb = max(int((end - span[1]) * ts.sample_rate.value), 0)
+        end = span[1]
+    pada = max(int((span[0] - start) * ts.sample_rate.value), 0)
+    padb = max(int((end - span[1]) * ts.sample_rate.value), 0)
     if pada or padb:
         return ts.pad((pada, padb), mode='constant', constant_values=(pad,))
     return ts
--- a/gwpy/timeseries/io/core.py
+++ b/gwpy/timeseries/io/core.py
@@ -49,6 +49,7 @@ def read(cls, source, *args, **kwargs):
     # read
     return io_read_multi(joiner, cls, source, *args, **kwargs)
 
+
 def _join_factory(cls, gap, pad, start, end):
     """Build a joiner for the given cls, and the given padding options
     """
--- a/gwpy/timeseries/io/core.py
+++ b/gwpy/timeseries/io/core.py
@@ -39,13 +39,17 @@ def read(cls, source, *args, **kwargs):
     # get join arguments
     pad = kwargs.pop('pad', None)
     gap = kwargs.pop('gap', 'raise' if pad is None else 'pad')
-    joiner = _join_factory(cls, gap, pad)
-
+    joiner = _join_factory(
+        cls,
+        gap,
+        pad,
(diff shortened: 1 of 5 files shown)


gwpy/timeseries/io/core.py
```diff
@@ -39,13 +39,20 @@ def read(cls, source, *args, **kwargs):
...
-    joiner = _join_factory(cls, gap, pad)
+    joiner = _join_factory(
+        cls,
+        gap,
+        pad,
+        kwargs.get("start", None),
+        kwargs.get("end", None),
+    )
+    # read and join
+    out = io_read_multi(joiner, cls, source, *args, **kwargs)
```
alurban: @duncanmmacleod, why not simply `return` on this line? i.e.:

```python
    return io_read_multi(joiner, cls, source, *args, **kwargs)
```

gwpy/timeseries/io/nds2.py
```diff
@@ -34,6 +34,7 @@
 from ...utils import gprint
 from ...utils.progress import progress_bar
 from .. import (TimeSeries)
+from .core import _pad_series
```
alurban: @duncanmmacleod, if the `_pad_series` utility is now being imported from another module, should it no longer be private?

gwpy/timeseries/io/core.py
```diff
@@ -56,11 +63,42 @@ def _join(data):
...
     return _join
+
+
+def _pad_series(ts, pad, start=None, end=None):
+    """Pad a timeseries to match the specified [start, end) limits
+
+    To cover a gap in data returned from a data source
+    """
+    span = ts.span
+    if start is None:
```
alurban: @duncanmmacleod, is it worth trying to define `pada` and `padb` on one line each?

```python
    pada = (0 if start is None else
            max(int((span[0] - start) * ts.sample_rate.value), 0))
    padb = (0 if end is None else
            max(int((end - span[1]) * ts.sample_rate.value), 0))
```

Personally, I just think the latter looks a little cleaner.

gwpy/timeseries/io/hdf5.py
```diff
@@ -42,6 +42,10 @@ def read_hdf5_timeseries(h5f, path=None, start=None, end=None, **kwargs):
     series = read_hdf5_array(h5f, path=path, **kwargs)
     # crop if needed
     if start is not None or end is not None:
```
alurban: @duncanmmacleod, is it worth trying to make it so these `if` statements aren't nested? Something like

```python
    if start is not None:
        start = max(start, series.span[0])
    if end is not None:
        end = min(end, series.span[1])
    if start or end:
        return series.crop(start, end)
    return series
```

gwpy/timeseries/io/nds2.py
duncanmmacleod: IMO, no, but only because it's being imported by another module in the same package. The underscore denotes that this is private in the scope of the package, not the module.

gwpy/timeseries/io/core.py
duncanmmacleod: I have simplified the logic in a different way, please re-review.

gwpy/timeseries/io/nds2.py
alurban: Ah, for some reason I'd always thought that was rule was at the module level, but this makes sense, thank you.

gwpy/timeseries/io/core.py
alurban: Looks good, we're really cookin' with gas now.

PR #1161, opened 2019-10-10
Fix to handle type changes when applying ADC scaling

duncanmmacleod:
This PR fixes a bug in the application of ADC scale factors that prevented type casting between int arrays and floats.

cc @pmeyers279, @basswinkels

### DIFF
--- a/ci/install-el.sh
+++ b/ci/install-el.sh
@@ -40,6 +40,10 @@ pushd build
 
 yum -y -q update
 
+# correct issue with missing tzdata files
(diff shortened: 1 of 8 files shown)


duncanmmacleod:
@gwpy/maintainers, this PR has expanded in scope somewhat since I authored it first, the changes are now:

- fix a bug in applying ADC scale factors related to type casting
- add support for writing `ADC`/`Sim` data types to GWF instead of just `Proc` (both framecpp and lalframe APIs) and related test suite additions

PR #1165, opened 2019-10-24
Add basic clustering method to EventTable

tjma12:
This adds a basic clustering method to `EventTable` objects. The clustering algorithm uses a pooling method to identify points that are all separated in time by less than `window`. The maximum value point of each cluster is calculated and replaces the cluster. 

An example (behind LIGO credentials) of clustering PyCBC Live triggers clustered with a 0.1 second window and maximizing over SNR.
Before: https://ldas-jobs.ligo.caltech.edu/~thomas.massinger/plots/test_clustering_unclustered_SNR.pdf
After: https://ldas-jobs.ligo.caltech.edu/~thomas.massinger/plots/test_clustering_clustered_SNR.pdf

Unit tests all pass when run from the command line.

### DIFF
--- a/gwpy/table/table.py
+++ b/gwpy/table/table.py
@@ -755,12 +755,12 @@ class EventTable(Table):
         # and divide the resulting array into clusters of adjacent points
         clusterpoints = numpy.where(numpy.diff(times) <= window)[0]
         sublists = numpy.split(clusterpoints,
-                            numpy.where(numpy.diff(clusterpoints) > 1)[0]+1)
+                               numpy.where(numpy.diff(clusterpoints) > 1)[0]+1)
 
         # Add end-points to each cluster and find the index of the maximum
         # point in each list
         padded_sublists = [numpy.append(s, numpy.array([s[-1]+1]))
-                          for s in sublists]
+                           for s in sublists]
         maxidx = [s[numpy.argmax(param[s])] for s in padded_sublists]
 
         # Construct a mask that removes all points within clusters and
--- a/gwpy/table/table.py
+++ b/gwpy/table/table.py
@@ -706,3 +706,64 @@ class EventTable(Table):
         >>> filter(my_table, ('time', in_segmentlist, segs))
         """
         return filter_table(self, *column_filters)
+
+    def cluster(self, timecolumn=None, clusterparam=None, window=None):
+        """Cluster this `EventTable` in time, maximizing over a specified
+        column in the table.
+
+        The clustering algorithm uses a pooling method to identify groups
+        of points that are all separated in time by less than `window`.
+
+        Each cluster of nearby points is replaced by the point in that cluster
+        with the maximum value of `clusterparam`.
+
+        Parameters
+        ----------
+        timecolumn : `str`
+            name of the column to use as the time axis when clustering
+
+        clusterparam : `str`
+            name of the column to maximize over in each time window
+
+        window : `float`
+            time window to use when clustering data points
+
+        Returns
+        -------
+        table : `EventTable`
+            a new table that has had the clustering algorithm applied via
+            slicing of the original
+
+        Examples
+        --------
+        To cluster an `EventTable` (``table``) whose `timecolumn` is
+        `end_time`, `window` is `0.1`, and maximize over `snr`:
+
+        >>> table.cluster(timecolumn='end_time', clusterparam='snr',
+                          window=0.1)
+
+        """
+        # Generate time and clusterparam vectors that are time-ordered
+        timeidx = numpy.argsort(self.get_column(timecolumn))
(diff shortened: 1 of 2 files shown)


gwpy/table/table.py
```diff
@@ -706,3 +706,67 @@ def filter(self, *column_filters):
...
+        `end_time`, `window` is `0.1`, and maximize over `snr`:
+
+        >>> table.cluster(timecolumn='end_time', clusterparam='snr',
+                          window=0.1)
+        """
+        if window <= 0.0:
+            raise ValueError('Window must be a positive value')
+
+        # Generate time and clusterparam vectors that are time-ordered
+        timeidx = numpy.argsort(self.get_column(timecolumn))
```
duncanmmacleod:
```suggestion
        timeidx = numpy.argsort(self[timecolumn])
```

gwpy/table/table.py
```diff
@@ -706,3 +706,67 @@ def filter(self, *column_filters):
...
+        times = self.get_column(timecolumn)[timeidx]
```
duncanmmacleod:
```suggestion
        times = self[timecolumn][timeidx]
```

gwpy/table/table.py
```diff
@@ -706,3 +706,67 @@ def filter(self, *column_filters):
...
+        param = self.get_column(clusterparam)[timeidx]
```
duncanmmacleod:
```suggestion
        param = self[clusterparam][timeidx]
```

gwpy/table/tests/test_table.py
```diff
@@ -505,6 +505,30 @@ def test_hist(self, table):
...
+        # check that a large cluster window returns at least one data point
+        t3 = t.cluster(timecolumn='time', clusterparam='amplitude', window=10)
+        assert len(t3) == 1
+        assert all(t3.get_column('amplitude') == [11])
+        assert all(t3.get_column('time') == [0.0])
+
+        # check that a non-positive window throws an appropriate ValueError
+        with pytest.raises(ValueError) as exc:
+            t.cluster(timecolumn='time', clusterparam='amplitude', window=0)
+        assert str(exc.value) == 'Window must be a positive value'
```
duncanmmacleod: Can you split this test into three separate tests, since they are all fairly independent? You might want to set up `clustertable` as a `fixture` so that it can be reused by each of the separate tests.

tjma12: Done, all tests pass from the command line

alurban:
You’re a true American hero, @tjma12

tjma12:
@duncanmmacleod @alurban I think the CI failures are unrelated to the code changes in my PR

gwpy/table/table.py
```diff
@@ -706,3 +706,67 @@ def filter(self, *column_filters):
...
+                           for s in sublists]
+        maxidx = [s[numpy.argmax(param[s])] for s in padded_sublists]
+
+        # Construct a mask that removes all points within clusters and
+        # replaces them with the maximum point from each cluster
+        mask = numpy.ones_like(times, dtype=bool)
+        mask[numpy.concatenate(padded_sublists)] = False
+        mask[maxidx] = True
+
+        return self[timeidx[mask]]
```
duncanmmacleod: Is this method actually `time` specific? Could we rename the arguments to just `column` and `rank`, or similar, to make it dimensionally agnostic?

tjma12: Fair point! I suppose there's no reason it has to be time-specific. I'll change a few variable names.

gwpy/table/tests/test_table.py
```diff
@@ -119,6 +119,13 @@ def create(cls, n, names, dtypes=None):
     def table(cls):
         return cls.create(100, ['time', 'snr', 'frequency'])
 
+    @classmethod
+    @pytest.fixture()
+    def clustertable(cls):
+        return cls.TABLE(data=[[11, 1, 1, 10, 1, 1, 9],
+                         [0.0, 1.9, 1.95, 2.0, 2.05, 2.1, 4.0]],
```
duncanmmacleod: Can you please line up the indentation here:
```suggestion
                               [0.0, 1.9, 1.95, 2.0, 2.05, 2.1, 4.0]],
```

or re-indent the whole thing:

```python
       return cls.TABLE(
            data=[
                [11, 1, 1, 10, 1, 1, 9],
                [0.0, 1.9, 1.95, 2.0, 2.05, 2.1, 4.0],
            ],
            names=['amplitude', 'time'],
        )

tjma12: Sure, pushed

tjma12:
@duncanmmacleod @alurban Looks like the build failure is due to checks which rely on open data access: https://travis-ci.com/gwpy/gwpy/jobs/250172290#L5232

PR #1192, opened 2019-11-27
Refactor gwpy-plot as an entry_point

duncanmmacleod:
This PR refactors the `gwpy-plot` command-line script as an [`entry_point`](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation) and separates it into its own binary package for Debian and RPM - this should have no impact on any functionality, but improves the cross-platform distribution of this script.

cc @areeda


areeda:
Sorry for the delay changes to the non albert.einstein accounts have made it different installing gwpy. I still don't have a good way but I'm  working on it.

I tested it on my development system and it looks good to me.

PR #1195, opened 2019-12-17
Patch pycbc empty_hdf5_files() to ignore gates dataset

duncanmmacleod:
This PR patches `gwpy.table.io.pycbc.patch_empty_files` to ignore the `'gates'` dataset that is now included in `pycbc_live` files. I also added a unit test for that function.

### DIFF
--- a/gwpy/table/io/pycbc.py
+++ b/gwpy/table/io/pycbc.py
@@ -234,11 +234,17 @@ def empty_hdf5_file(h5f, ifo=None):
     # the decorator opens the HDF5 file for us, so h5f is guaranteed to
     # be an h5py.Group object
     h5f = h5f.file
+    # if the root group is empty, the file is empty
     if list(h5f) == []:
         return True
-    if ifo is not None and (ifo not in h5f or list(h5f[ifo]) == ['psd']):
-        return True
-    return False
+
+    # for each group (or the IFO group given by the user),
+    # check whether there is any useful content
+    groups = h5f.keys() if ifo is None else [ifo]
+    for group in groups:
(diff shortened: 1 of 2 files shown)


gwpy/table/io/pycbc.py
```diff
@@ -234,11 +234,21 @@ def empty_hdf5_file(h5f, ifo=None):
...
-    return False
+
+    # for each group (or the IFO group given by the user),
+    # check whether there is any useful content
+    if ifo is None:
+        groups = h5f.keys()
+    else:
+        groups = [ifo]
+    for group in groups:
+        print(set(h5f.get(group, [])))
```
alurban: @duncanmmacleod, these changes look great, unit tests are passing, and hotfixing them fixes the issues I was seeing on the production summary pages, so thank you for putting this together.

That having been said, I think this stray print statement needs to be removed, after which I'm happy to approve this.

gwpy/table/io/pycbc.py
```diff
@@ -234,11 +234,21 @@ def empty_hdf5_file(h5f, ifo=None):
...
+    # if the root group is empty, the file is empty
     if list(h5f) == []:
         return True
-    if ifo is not None and (ifo not in h5f or list(h5f[ifo]) == ['psd']):
-        return True
-    return False
+
+    # for each group (or the IFO group given by the user),
+    # check whether there is any useful content
+    if ifo is None:
```
alurban: @duncanmmacleod, it may also appease codeclimate to replace this with a one-liner:

```python
    groups = [ifo] if ifo else h5f.keys()
```

PR #1201, opened 2020-01-16
Omicron clustering algorithm in EventTable.cluster() method

dcorre:
In addition to the current `EventTable.cluster()` algorithm, one can set `index`='omicron' to perform the clustering similarly as omicron.

Examples
        --------

**As before**, to cluster an `EventTable` (``table``) whose `index` is
 `end_time`, `window` is `0.1`, and maximize over `snr`:

        >>> table.cluster('end_time', 'snr', 0.1)

**New**, To cluster an `EventTable` (``table``) with the omicron clustering
algorithm, with a `window` of `0.1`, and maximize over `snr`:

        >>> table.cluster('omicron', 'snr', 0.1)


dcorre:
Hi @duncanmmacleod , after some time I can work again on the clustering and pyomicron things. As we discussed, I simply added a conditional check within table.cluster() to choose between the current clustering method and the one implemented in omicron. However, according to codeclimate it is a bit too complex due to the break in the linear flow of the code. 

So I'm wondering whether we can ignore the codeclimate check? Otherwise I guess I will need to write 2 distincts methods.

Regarding the test failures, I am trying to understand where the problem with  test_fetch_hacr() comes from. Any ideas?

duncanmmacleod:
@dcorre, please rebase this branch against the upstream`master` branch, which should have fix the unrelated test failures.

gwpy/table/tests/test_table.py
```diff
@@ -496,6 +514,47 @@ def test_cluster_window(self, clustertable):
...
+        # each cluster, the original table is unchanged, and all points
+        # return their intended values
+        t = clustertable_omicron.cluster('omicron', 'snr', 0.1)
+        assert len(t) == 4
+        assert len(clustertable_omicron) == 10
+        assert all(t['time'] == [15.09375, 18.708007, 23.265625, 23.90625])
+        assert all(t['snr'] == [5.92, 5.18, 5.1, 5.78])
+        assert all(t['amplitude'] == [6.08, 3.13, 8.93, 1.11])
+        assert all(t['tstart'] == [15.0625, 18.707031, 23.25, 23.875])
+        assert all(t['tend'] == [15.125, 18.708985, 23.28125, 24.0])
```
duncanmmacleod, line 528: It might be simpler to use `numpy.testing.assert_array_equal` or similar here.

gwpy/table/table.py
```diff
@@ -754,31 +895,51 @@ def cluster(self, index, rank, window):
...
+            maxidx.append(triggers[mask][numpy.argmax(param[mask])]['idx'])
+            if index == 'omicron':
+                tstart_list.append(numpy.min(triggers[mask]['tstart']))
+                tend_list.append(numpy.max(triggers[mask]['tend']))
+
+        # Apply these values to tile representing the cluster, only for
+        # the omicron method. Might need to use it as well for other method
+        if index == 'omicron':
+            triggers['tstart'][maxidx] = tstart_list
+            triggers['tend'][maxidx] = tend_list
```
duncanmmacleod, line 938: Please add the necessary logic to manipulate all other columns that are modified by the C++ version of this method, including `flow/fhigh` etc. We should strive to have the python method reproduce the C++ method exactly (including all parameter values).

duncanmmacleod (changes requested):
Thanks @dcorre, this will make a nice addition to gwpy!

I have requested some pedantic changes, but also some restructuring to make the code more modular, and more atomic.

I am a little confused why the original cluster method is much more complex when moved into the new `clustering` method, but that might end up being clear once the other changes have been made.

gwpy/table/table.py
dcorre: I included all of your recommandations. 

Appart from codeclimate and cognitive complexity, it looks fine. I'll let you check codeclimate outputs to see if it really matters.

dcorre:
Regarding the reason why I changed so much the original method, well it is quite a while now since I've made theses changes... but  think it was becasue it did not allow to associate a tile/trigger (i.e row) to a given cluster as I intend to do with these changes.

I'll make some tests, probably tomorrow, to see if I can keep the old format with these changes. 

By the way, it would make more sense if the fstart/fend, tstart/tend values of the clusters identified by the original method correspond to the min/max of the rows included in them and not to the values of the row with the highest snr (or whatever quantity you want to maximise over). With these changes it is straightforward, with the original method, I am not sure. 
So if there are no particular reasons, apart from the fact that it was easier to code, we could also update this.

dcorre:
Actually they are the same methods, but I just added some checks for missing triggers.

The reason is that I'm using an array, self.clustersList,  containing the list of rows indexes included in a given cluster. This array is the output of both omicron and the 'new' original method. Then in identify_clusters() these groups of points are assigned to a given cluster and each point/row/tile is assigned to a given cluster and store in self.tile_cluster_id . So I need to associate each point to a cluster, even if it is a single point cluster.

The original method does not identify successive rows that do not fullfill the window criteria.
To illustrate, here is an output of ''padded_sublists''  (whose aim is to group rows together)

[array([0, 1, 2]), array([3, 4, 5, 6, 7]), array([8, 9]), array([12, 13, 14, 15, 16]), array([17, 18, 19])]

You see that rows 10 and 11 are missing here because thay do not fullfill the window criteria.   

In the original method, this is not a problem thanks to the use of :

mask1 = numpy.ones_like(col, dtype=bool)
mask1[numpy.concatenate(padded_sublists)] = False
mask1[maxidx] = True

So missing points (i.e. individual triggers) are set to True.

To summarise, old and new implementation are based on the same method, but as I want to assocaite each point in the EventTable to a given cluster, I had the need to add some code to search for missing points (that can be also at the beginning or end of the EventTable) and then to insert them in the ''padded_sublists''. 

Do you have a better solution?

gwpy/table/table.py
```diff
@@ -720,20 +720,244 @@ def filter(self, *column_filters):
...
+            for i in range(len(missing_trigger)):
+                clustersList.insert(len(clustersList)+i,
+                                    numpy.atleast_1d(missing_trigger[i]))
+
+        self.clustersList = clustersList
+
+        # keep track of the order indexes
+        self.orderidx = orderidx
+
+        return True
```
duncanmmacleod, line 858: It makes more sense to me to have these functions return the list of clusters, rather than storing them in another internal property, since the list is only ever used internal to the `identify_clusters` method.

Can you please consider that change?

duncanmmacleod:
@dcorre, what's the status of this work? Can you please rebase against the upstream master branch to refresh things, and then address the comments above?

eagoetz:
> @dcorre, what's the status of this work? Can you please rebase against the upstream master branch to refresh things, and then address the comments above?

@dcorre A gentle nudge on this PR.

duncanmmacleod:
Ping @dcorre.

duncanmmacleod:
@dcorre, what is the status of this work? I'm sorry that I wasn't able to devote enough time to it when you originally proposed it.

PR #1204, opened 2020-01-29
Improve reading of empty tables from LIGO_LW XML

duncanmmacleod:
This PR improves the reading of empty tables from `LIGO_LW`-format XML. The tricky part is determining the correct `dtype` for dynamic columns (i.e. ones that aren't natively part of the XML structure, but are supported by the `ligo.lw` python API).


scottcoughlin2014:
@duncanmmacleod this fixes my issue I think it looks good!

PR #1217, opened 2020-03-16
Fix test mocks for sqlparse-0.3.1

duncanmmacleod:
This PR fixes test failures relating to mocking of SQL outputs using `sqlparse-0.3.1`.


duncanmmacleod:
@alurban, can you review this please; if we can get this merged then we can start moving properly on the python2 removal.

PR #1220, opened 2020-03-16
Bump minimum scipy and numpy version requirements

alurban:
This PR bumps the minimum scipy version requirement to 1.2.0, and the minimum numpy version requirement to 1.12.0. The former is dictated by a need to support median averaging in spectrograms, and the latter is determined by a desire to support `numpy.geomspace`, both the most encompassing requirements I could find scattered throughout gwpy.

Per instructions left in comment lines, this includes a somewhat large-scale refactoring of `gwpy/signal/spectral/_scipy.py`.

cc @duncanmmacleod

### DIFF
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -31,7 +31,7 @@ import pytest
 import numpy
 from numpy import testing as nptest
 
-from scipy import (signal, __version__ as scipy_version)
+from scipy import signal
 
 from astropy import units
 
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -55,10 +55,7 @@ SKIP_LAL = utils.skip_missing_dependency('lal')
 SKIP_LALFRAME = utils.skip_missing_dependency('lalframe')
 SKIP_PYCBC_PSD = utils.skip_missing_dependency('pycbc.psd')
 
-if scipy_version < '1.2.0':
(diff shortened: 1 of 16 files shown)


alurban:
@duncanmmacleod, this PR is ready for review, note that the two CircleCI failures involve EPEL7 and Debian, for which you are in the process of dropping support.

alurban:
@duncanmmacleod, I've rebased this against #1224, so CircleCI is running without RHEL/Debian and TravisCI is running with python3.8.

alurban:
@duncanmmacleod, this is done in the latest commit

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -55,10 +55,7 @@
 SKIP_LALFRAME = utils.skip_missing_dependency('lalframe')
 SKIP_PYCBC_PSD = utils.skip_missing_dependency('pycbc.psd')
 
-if scipy_version < '1.2.0':
-    SCIPY_METHODS = ('welch', 'bartlett')
-else:
-    SCIPY_METHODS = ('welch', 'bartlett', 'median')
+SCIPY_METHODS = ('welch', 'bartlett', 'median')
```
duncanmmacleod: I think this variable can be removed entirely now, since its not really variable any more, any just replace its usage (decorator to `test_psd`) with the tuple.

alurban: Makes sense to me, thanks @duncanmmacleod, this is done in the latest commit

PR #1222, opened 2020-03-17
Drop dependence on the six package and remove python < 3 workarounds

### LINKED ISSUE #1170: Drop use of six
When we drop support for python2, we can remove all usage of the `six` module.

### LINKED ISSUE #1173: Simplify return_as decorator for python3
When we drop support for python2, the following code can and should be simplified:

https://github.com/gwpy/gwpy/blob/2b4e780bd4323d6c150483bef253e3e0e7b7025e/gwpy/utils/decorators.py#L101-L119

alurban:
This PR drops dependence on the `six` package and removes all workarounds for python < 3, including those related to `python-ligo-lw` which should now be python3 compatible.

Fixes #1170, fixes #1173

cc @duncanmmacleod

### DIFF
--- a/gwpy/table/gravityspy.py
+++ b/gwpy/table/gravityspy.py
@@ -19,12 +19,18 @@
 """Extend :mod:`astropy.table` with the `GravitySpyTable`
 """
 
-import os.path
+import json
+import os
+from urllib.error import HTTPError
+from urllib.parse import urlencode
 from urllib.request import urlopen
+
+import numpy as np
+
+from astropy.utils.data import get_readable_fileobj
 
 from ..utils import mp as mp_utils
 from .table import EventTable
-import numpy as np
 
 __author__ = 'Scott Coughlin <scott.coughlin@ligo.org>'
 __all__ = ['GravitySpyTable']
@@ -71,7 +77,6 @@ class GravitySpyTable(EventTable):
         -------
         Folder containing omega scans sorted by label
         """
-        import os
         # back to pandas
         try:
             images_db = self.to_pandas()
@@ -191,11 +196,6 @@ class GravitySpyTable(EventTable):
         an evaluation of the Euclidean distance of the input image
         to all other images in some Feature Space
         """
-        from astropy.utils.data import get_readable_fileobj
-        import json
-        import urllib
-        from urllib.error import HTTPError
-
         # Need to build the url call for the restful API
         base = 'https://gravityspytools.ciera.northwestern.edu' + \
             '/search/similarity_search_restful_API'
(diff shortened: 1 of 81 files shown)


alurban:
@duncanmmacleod, this is now ready for review, and can be merged without waiting on anything else any RHEL/Debian requirements.

gwpy/utils/decorators.py
```diff
@@ -112,10 +111,6 @@ def wrapped(*args, **kwargs):
                         func.__name__, returntype.__name__, str(exc)),
                 )
                 raise
-        try:
-            return wraps(func)(wrapped)
-        except AttributeError:  # python < 3.0.0
-            wrapped.__doc__ == func.__doc__
-            return wrapped
+        return wraps(func)(wrapped)
```
duncanmmacleod: @alurban, this should just be I think)...:

```python
        @wraps(func)
        def wrapped(*args, **kargs):
            ...
```

rather than calling `wraps` as a function. You can also remove the entire comment about wraps.

alurban: @duncanmmacleod, I was hoping you'd catch that one, I tried to figure it out last night but had gotten stuck. Your suggestion appears to work for me locally, we'll find out soon whether build tests pass

duncanmmacleod:
@alurban, this needs rebased.

alurban:
Thanks @duncanmmacleod, this has been rebased

gwpy/table/gravityspy.py
```diff
@@ -194,8 +193,8 @@ def search(cls, gravityspy_id, howmany=10,
         """
         from astropy.utils.data import get_readable_fileobj
         import json
-        from six.moves.urllib.error import HTTPError
-        from six.moves import urllib
+        import urllib
+        from urllib.error import HTTPError
```
duncanmmacleod: Can these `import`s be moved to the module leve, with `urllib` replaced by the full `from urllib.parse import urlencode`?

gwpy/table/tests/test_table.py
```diff
@@ -391,7 +377,7 @@ def test_filter(self, table):
         utils.assert_table_equal(
             midf, table.filter('frequency > 100').filter('frequency < 1000'))
 
-        # check unicode parsing (PY2)
+        # check unicode parsing
         table.filter(u'snr > 100')
```
duncanmmacleod: We don't need this test in python3.

```suggestion
```

gwpy/table/gravityspy.py
alurban: @duncanmmacleod, I agree with the suggestion, and I took a stab at this with the most recent commit.

alurban:
Ping @duncanmmacleod

Issue #1228, opened 2020-03-25
Change default PSD averaging method


alurban:
@duncanmmacleod, now that we require scipy version 1.2.0 or better, do we also want to change the default PSD averaging method from ‘mean’ (standard Welch averaging) to ‘median’?

**Progressive argument:** gwpy is intended for gravitational wave data analysis, and every existing ground-based interferometer has nonstationary noise with loud, transient glitches, which median averaging is very good at mitigating. We already recommend that users employ this method; after all, it was the motivation for requiring at least scipy-1.2.0. So it is the natural default to use if we adopt the philosophy that sensible defaults should make the code work out-of-the-box.

**Conservative argument:** the default has always been Welch averaging, it is both the simplest method available and the first one most people see when they learn about PSDs. It is also still the default used by scipy, so it is the natural default for gwpy to continue using if cross-package consistency is a higher priority.

Anyone who knows where I land politically will be entirely unsurprised to hear that I personally favor the progressive argument.

cc @tjma12, @dethodav, @areeda, and any others who may have strong feelings

tdent:
Median is *quite* good at mitigating glitches .. as long as enough independent samples are used.

Does the standard median method include the median bias correction? 

In terms of implications, what standard results/products/plots would be affected by the change? And would any of them be affected 'for the worse'?

duncanmmacleod:
> Does the standard median method include the median bias correction?

@tdent: yes, this was implemented in https://github.com/scipy/scipy/pull/8970

alurban:
Salient points as ever @tdent, I know that all of the production services downstream of gwpy already use median averaging where appropriate (particularly the summary pages) as do many users, at least in LIGO DetChar. To be cautious and minimize adverse impact, we can raise a loud `UserWarning` signaling that the default behaviour will change in a future minor release.

PR #1234, opened 2020-04-15
Raise DeprecationWarning for a future change to the default spectral averaging method

alurban:
This PR raises a `DeprecationWarning` loudly warning users that the default spectral averaging method will change to `'median'` in gwpy-2.1.0.

This is related to #1228.

cc @duncanmmacleod

### DIFF
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -36,7 +36,7 @@ from .core import (TimeSeriesBase, TimeSeriesBaseDict, TimeSeriesBaseList,
 
 __author__ = 'Duncan Macleod <duncan.macleod@ligo.org>'
 
-DEFAULT_FFT_METHOD = "welch"
+DEFAULT_FFT_METHOD = None
 
 
 # -- utilities ----------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


gwpy/timeseries/timeseries.py
```diff
@@ -289,6 +289,8 @@ def psd(self, fftlength=None, overlap=None, window='hann',
         - ``'bartlett'`` : a mean average of non-overlapping periodograms
         - ``'median'`` : a median average of overlapping periodograms
         - ``'welch'`` : a mean average of overlapping periodograms
+
+        If ``method=None`` is passed, this will default to 'welch'.
```
duncanmmacleod, line 293: Instead of this, could you just modify the `method` parameter docstring above similar to what is done in `gwpy.astro`?

```
        method : `str`, optional
            FFT-averaging method (default: ``'welch'``), see *Notes* for more details
```
or similar

alurban: @duncanmmacleod, this is done in the latest commit

duncanmmacleod:
@alurban, this seems good to go, but can you please squash the commits together, so that when we come to actually implement the new default, we can just `git revert` this one commit, then add another that swaps `'welch'` with `'median'`

alurban:
> @alurban, this seems good to go, but can you please squash the commits together, so that when we come to actually implement the new default, we can just `git revert` this one commit, then add another that swaps `'welch'` with `'median'`

Good idea, I've replaced these 4 commits with one squashed commit.

alurban:
@duncanmmacleod, this PR finally passes all the CI checks, so should be good to go

PR #1238, opened 2020-05-03
Implement a TimeSeries.mask method

### LINKED ISSUE #1237: Implement segment masking for TimeSeries objects
It would be reasonably straightforward to let users mask out unwanted segments (e.g., non-observing-mode data) from their `TimeSeries`, something like:

```python
import numpy
from gwpy.segments import DataQualityFlag

# retrieve segment and timeseries data
flag = DataQualityFlag.query(...).active
data = TimeSeries.read(...)

# mask out non-observing time
indices = numpy.nonzero([x not in flag for x in data.times.value])
data.value[indices] = numpy.nan
```

(Of course the actual method should be more careful than to call `data.times`, but this is just an example.)

alurban:
This PR implements a new method, `TimeSeries.mask`, that masks away portions of a given timeseries which fall within of a user-defined deadtime. This can be given as either:

* A `SegmentList` custom-defining the deadtime
* The name of a data-quality flag to be queried for

The former will supersede the latter if both are given, and in the second case, the `DataQualityFlag.isgood` attribute is used to determine whether the given flag defines livetime or deadtime.

This method also optionally tapers away data at segment boundaries, with the taper length controlled by the user. The following corner cases are explicitly handled:

* The edges of the original `TimeSeries` **will not** be tapered
* Any livetime shorter than twice the taper length will get a subset of the tapering window, chosen so that (1) the taper length remains fixed by the user and (2) the ramp-up-ramp-down across that segment is still continuous. This reproduces the existing behavior of `TimeSeries.gate`

Lastly, because this PR effectively refactors some of the functionality in `TimeSeries.gate`, I have also simplified that method so that it invokes `TimeSeries.mask`.

This fixes #1237.

cc @duncanmmacleod, @siddharth101

### DIFF
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -36,8 +36,9 @@ from scipy import signal
 from astropy import units
 
 from ...frequencyseries import (FrequencySeries, SpectralVariance)
-from ...segments import Segment
+from ...segments import (Segment, SegmentList, DataQualityFlag)
 from ...signal import filter_design
+from ...signal.window import planck
 from ...table import EventTable
 from ...spectrogram import Spectrogram
 from ...testing import (mocks, utils)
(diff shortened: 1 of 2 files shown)


duncanmmacleod:
@alurban, apologies for letting this linger. I think this needs to be rebased for the CI to pass.

PR #1245, opened 2020-05-21
Changed fetch_catalog to work with new eventapi

rngeorge:


duncanmmacleod:
@rngeorge, can you please also update the version dependency for `gwosc` in the following places:

- `docs/install/index.rst`
- `setup.py`

to `>=0.5.3`, which is the first working version that provides this new API.

gwpy/table/io/losc.py
```diff
@@ -18,7 +18,7 @@
 
 """Fetch and parse an event catalog from GWOSC.
 """
-
+import json
```
duncanmmacleod, line 21: Please remove this unused import

```suggestion
```

rngeorge: Pytest keeps failing in test_table.py when it runs:

def test_fetch_open_data_kwargs(self)
and
def test_fetch_open_data(self)

Specifically, I cannot find a parameter to check for masking since none of the values are masked anymore. This means the following line fails.

assert isinstance(gw170818["snr_pycbc"], MaskedConstant)

duncanmmacleod:
> Pytest keeps failing in test_table.py when it runs:
> 
> def test_fetch_open_data_kwargs(self)
> and
> def test_fetch_open_data(self)
> 
> Specifically, I cannot find a parameter to check for masking since none of the values are masked anymore. This means the following line fails.
> 
> assert isinstance(gw170818["snr_pycbc"], MaskedConstant)

Since there are no masked values any more, we should remove the logic that supports it, and remove that part of the test. Can you take care of that as part of this change please?

gwpy/table/tests/test_table.py
```diff
@@ -742,21 +741,29 @@ def test_fetch_open_data(self):
...
         try:
             table = self.TABLE.fetch_open_data(
                 "GWTC-1-confident",
-                selection="mass1 < 5",
-                columns=["name", "mass1", "mass2", "distance"])
+                selection="mass_1_source < 5",
+                columns=[
+                    "name", "mass_1_source",
+                    "mass_2_source", "luminosity_distance"
+                ]
```
duncanmmacleod, line 760:
```suggestion
                columns=[
                    "name",
                    "mass_1_source",
                    "mass_2_source",
                    "luminosity_distance"
                ]
```

gwpy/table/tests/test_table.py
```diff
@@ -742,21 +741,29 @@ def test_fetch_open_data(self):
...
         except (URLError, SSLError) as exc:  # pragma: no-cover
             pytest.skip(str(exc))
         assert len(table) == 1
-        assert table[0]["name"] == "GW170817"
-        assert set(table.colnames) == {"name", "mass1", "mass2", "distance"}
+        assert table[0]["name"] == "GW170817_R1"
+        assert set(table.colnames) == {
+            "name", "mass_1_source",
+            "mass_2_source", "luminosity_distance"
+        }
```
duncanmmacleod, line 769:
```suggestion
        assert set(table.colnames) == {
            "name",
            "mass_1_source",
            "mass_2_source",
            "luminosity_distance"
        }
```

gwpy/table/io/losc.py
```diff
@@ -102,17 +101,22 @@ def _mask_column(col):
...
 @read_with_selection
 def fetch_catalog(catalog, host=DEFAULT_GWOSC_URL):
-    catalog = download_catalog(catalog, host=host)
-    data = catalog["data"]
-    parameters = catalog["parameters"]
+    catalog = fetch_catalog_json(catalog, host=host)
+    data = catalog["events"]
+
+    firstdata = next(iter(data))
+    parameters = list(data[firstdata])
```
duncanmmacleod, line 108: I'm afraid that this might encode an incorrect assumption about the GWOSC catalogs, namely that all events in the catalog have the same parameters.

@jkanner, can you confirm whether all events in the eventapi catalog views are guaranteed to have the same parameters? Or should we prepare for the eventually that some events have data that others don't? I'm imagining mixed catalogs of CBC and burst events in the future.

@rngeorge, the safest approach would be to take the logical OR of all available parameters:

```python
parameters = set(key for event in data.values() for key in event)
```

and then reinstate the 'masking' we had before to handle rows that have missing parameters.

gwpy/table/io/losc.py
```diff
@@ -102,17 +101,22 @@ def _mask_column(col):
...
+
+    unitlist = {k: v for k, v in data[firstdata].items() if k.endswith('unit')}
+    parameters = [x for x in parameters if 'unit' not in x]
```
duncanmmacleod, line 111: We should probably use identically opposite logic to get the non-unit columns as we do in the preceding line to get the unit columns:

```suggestion
    parameters = [x for x in parameters if not x.endswith('unit')]
```

gwpy/table/io/losc.py
jkanner: Hi Duncan,

That's interesting.  For now, yes, all events have the same parameters at the top level.  Though, individual PE sets may have an arbitrary list of parameters (and each PE set may be different). 

We haven't yet thought about how to represent different types of events (like Supernova or pulsars) - for now, they are all CBC.  I agree that thinking ahead, and imagining a heterogenous list at some point in the future is good planning.

PR #1252, opened 2020-07-02
Remove unused import

duncanmmacleod:
This PR removes an unused import that somehow slipped past the linter.

### DIFF
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -21,7 +21,6 @@
 
 import os.path
 from itertools import (chain, product)
-from math import isnan
 from ssl import SSLError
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
This is bogus, somehow.

alurban:
@duncanmmacleod, do you mean the PR itself was bogus or the result of the linter?

duncanmmacleod:
The PR, I got tangled in some branches (git pun) and had some trouble working my way out.

alurban:
Aha! Happens to the best of us, I just wanted to make sure it wasn't a security issue.

duncanmmacleod:
Nope, it was actually me - the posted commit has my GPG signature on it, which is as good security as we get.

PR #1253, opened 2020-07-02
Update test for matplotlib-3.2.2

duncanmmacleod:
This PR fixes a failing test on matplotlib-3.2.2, which has updated TeX text formatting compared to 3.2.1.

### DIFF
--- a/gwpy/plot/tests/test_log.py
+++ b/gwpy/plot/tests/test_log.py
@@ -26,7 +26,10 @@ except ImportError:  # python < 3
 
 import pytest
 
-from matplotlib import rc_context
+from matplotlib import (
+    __version__ as mpl_version,
+    rc_context,
+)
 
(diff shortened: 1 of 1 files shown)


gwpy/plot/tests/test_log.py
```diff
@@ -44,10 +47,27 @@ def formatter(cls):
...
+            None,
+            r'$\mathdefault{0}$',
+            '$0$',
+            id="0",
+        ),
+        pytest.param(
+            1,
+            None,
+            r'$\mathdefault{10^{0}}$',
+            r'$\mathdefault{10^{0}}$' if mpl_version > "3.2.1" else "$10^{0}$",
```
duncanmmacleod, line 61: This line is the only changed functional line, everything else is just re-formatting.

PR #1255, opened 2020-07-13
Update copyright for release

duncanmmacleod:
This PR updates the copyright for the pending 2.0.0 release.

### DIFF
--- a/examples/frequencyseries/inject.py
+++ b/examples/frequencyseries/inject.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (C) Alexander Urban (2018-2020)
+# Copyright (C) Alex Urban (2018-2020)
 #
 # This file is part of GWpy.
 #
--- a/examples/frequencyseries/inject.py
+++ b/examples/frequencyseries/inject.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (C) Alexander Urban (2018-2019)
+# Copyright (C) Alexander Urban (2018-2020)
 #
 # This file is part of GWpy.
 #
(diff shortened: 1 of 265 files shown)


examples/frequencyseries/inject.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (C) Alexander Urban (2018-2019)
+# Copyright (C) Alexander Urban (2018-2020)
```
alurban, line 3: Can we change this for consistency with the others (and to match my preference)?

```suggestion
# Copyright (C) Alex Urban (2018-2020)
```

duncanmmacleod: I have applied this suggestion, but please note that it is extremely unlikely that you are actually eligible to own the copyright of this file, in the same way that I am probably not eligible to own the copyright on the rest of the codebase. These rights likely go to Cardiff and LSU as appropriate, but they are unlikely to ever care.

alurban:
Thanks @duncanmmacleod, can you explain how that works? I was under the impression that you owned the copyright to the codebase under terms of the GNU license, so I’m not sure I understand how ownership gets assigned under copyright law.

duncanmmacleod:
> Thanks @duncanmmacleod, can you explain how that works? I was under the impression that you owned the copyright to the codebase under terms of the GNU license, so I’m not sure I understand how ownership gets assigned under copyright law.

Standard procedure is that your employer claims copyright ownership over all code you write in the normal course of your employment - this is definitely the case for Cardiff University. The specific terms should have been laid out in your employment contract, I can't remember the terms when I signed at LSU.

The License only dictates the permissions the copyright holder grants to _other_ people who wish to use/reuse/redistribute the code. This is legally independent of the copyright attribution, although your employer may dictate what sorts of licenses they allow you to use.

All of this is not legal advice, so you might want to check with the LSU intellectual property office.

PR #1265, opened 2020-08-19
Add support for reading gstlal trigger files

dethodav:
This pull request aims to add support for reading gstlal trigger files.

The rationale for a specific gstlal format is 

* A desire to calculate some useful derived quantities. This feature is available for only pycbc trigger files at present. 
* The gstlal statistic requires additional inputs beyond what is included in trigger files in order to calculate the ranking statistic. However, the comparable quantity is stored in the `coinc_event` table. Hence reading in both the `coinc_event` and `coinc_inspiral` tables at once would be a useful feature. 

Thew new proposed format heavily relies upon the current ligolw functionality, with the addition of the two features above.

### DIFF
--- a/gwpy/table/io/gstlal.py
+++ b/gwpy/table/io/gstlal.py
@@ -199,7 +199,7 @@ def get_snr_chi(events, snr_pow=2., chi_pow=2.):
     """
     snr = events['snr'][:]
     chisq = events['chisq'][:]
-    snr_chi = snr**snr_pow / chisq**chi_pow 
+    snr_chi = snr**snr_pow / chisq**(chi_pow/2.) 
     return snr_chi
 
 GET_COLUMN['snr_chi'] = get_snr_chi
--- a/gwpy/table/io/gstlal.py
+++ b/gwpy/table/io/gstlal.py
@@ -1,0 +1,164 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) Duncan Macleod (2014-2020)
+#
+# This file is part of GWpy.
+#
+# GWpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# GWpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with GWpy.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Read events from the PyCBC live online GW search
+"""
+
+import re
+from os.path import basename
+
+import numpy
+from astropy.table import join
+
+from ...io.registry import (register_reader, register_identifier)
+from .. import (Table, EventTable)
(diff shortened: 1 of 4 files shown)


dethodav:
At present, this pull request is not ready for merge - one of the remaining items is setting up the necessary registry in the format file. @duncanmmacleod @alurban Would you be able to help explain how the file registry is set up?

duncanmmacleod:
@dethodav, is this still a desired change? If 'yes' (I hope so) then I can help get things straightened out, tested, reviewed, and merged.

First up, please `rebase` against the upstream `master` branch.

dethodav:
@duncanmmacleod Yes, I would still be interested in adding this feature to gwpy. I've gone ahead and rebased to the current main branch and updated the PR. Let me know your suggested next steps.

duncanmmacleod:
@dethodav, I let this linger again, please rebase. I think the PEP8/flake linter issues are real, so please have a look at those.

Please think also about how to add tests of the new functionality.

gwpy/table/io/gstlal.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+### !!! unclear how to set up register
+def identify_gstlal(origin, filepath, fileobj, *args, **kwargs):
+    """Identify a PyCBC Live file as an HDF5 with the correct name
+    """
+    if identify_hdf5(origin, filepath, fileobj, *args, **kwargs) and (
+            filepath is not None and PYCBC_FILENAME.match(basename(filepath))):
+        return True
+    return False
```
duncanmmacleod, line 119: @dethodav, the `identify_gstlal` function should take in the arguments are you have them listed, and inspect them to unimbiguously detect whether the incoming file is a 'gstlal file' (returning `True`) or not (`False`).

For PyCBC we do this with a regular expression on the filename, which isn't great, but seems to work well enough.

What is unique about a GstLAL file?

dethodav:
@duncanmmacleod Here is an example gstlal xml file (available at CIT): `/home/gstlalcbc/observing/4/dev/runs/trigs.edward_O3_replay_MDC06/13443/H1L1V1-0598_noninj_LLOID-1344312382-14414.xml.gz`

dethodav:
@duncanmmacleod I've gone through and made updates to this PR. Things should be ready for review modulo the failing "Dependencies" check. Do you have any suggestions on how best to resolve this import issue?

duncanmmacleod (changes requested):
> @duncanmmacleod I've gone through and made updates to this PR. Things should be ready for review modulo the failing "Dependencies" check. Do you have any suggestions on how best to resolve this import issue?

Thanks @dethodav, the fundamental issue here is that `python-ligo-lw` is not a first-class requirement of GWpy, it's optional. So you need to protect the imports in `gwpy/table/io/gstlal.py` and `gwpy/table/tests/test_io_gstlal.py` so that the test suite executes without it.

Refer to `gwpy/table/io/ligolw.py` and `gwpy/table/tests/test_io_ligolw.py` for inspiration there.

You will also need to add a reference to the new `gstlal` module in `gwpy/table/io/__init__.py` so that the I/O format is registered automatically when the `EventTable` is first imported.

dethodav:
@duncanmmacleod I've now addressed the rest-related issues. Let me know if there's anything else that needs to be fixed before merger.

duncanmmacleod (approved):
Thanks for this @dethodav, looks great.

PR #1267, opened 2020-08-26
Use bash scripts on windows CI as well

duncanmmacleod:
This PR updates the current CI configurations to control everything using the `bash` scripts in the `ci/` subdirectory. This has meant a small amount of restructuring, but means that the CI runs are much more homogeneous across CI providers, especially on Windows.

There is a small amount of extra logic that might not make sense now, but supports a new Azure Pipelines configuration that has been introduced and will be configured properly once it hits the `master` branch of the main repo. I will propose in a following PR to remove the other CI configurations in favour of one-ci-to-rule-them-all from Azure. The current Azure build for this PR is manually configured from my fork, but outputs here: 

https://dev.azure.com/duncanmmacleod/Test/_build/results?buildId=44&view=results

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,30 +8,13 @@ environment:
 install:
   - "set PATH=%MINICONDA%\\condabin;%PATH%"
   - call "%MINICONDA%\\Scripts\\activate.bat"
-  - conda config --set always_yes yes
-  - conda config --add channels conda-forge
-  - conda create --quiet --yes --name gwpy
-        "python=%PYTHON_VERSION%"
-        "pip"
-        "setuptools"
-  - conda activate gwpy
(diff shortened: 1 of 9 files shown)


alurban (approved):
Thanks @duncanmmacleod, I'm going to trust that because the build tests in this PR are passing, that effectively validates the changes. No obvious notes jump out at me after looking over file diffs line-by-line.

One question, does this mean that in the future gwpy will move away from Travis/Appveyor and towards Azure for all continuous integration?

duncanmmacleod:
@alurban, yes. Appveyor has issues with SSL certificates that mean all of the tests that rely on remote data fail. Also, having one config for all platforms rather than 3 separate ones should be easier to maintain.

alurban:
Thanks @duncanmmacleod, that makes sense, I have no specific arguments against Azure but was just asking for clarity.

PR #1269, opened 2020-09-01
Fix internal use of urlparse(x).path to strip file url schemes

duncanmmacleod:
This PR modifies two internal uses of `urlparse(x).path` to instead use the `gwpy.io.utils.file_path` function. Just stripping the scheme out (which is what `urlparse(x).path` does) sometimes causes problems on Windows when it gets passes a string with a drive name in it (e.g. `C:\Users\me\myfile.txt`). This should have no impact on anybody (except future me).

### DIFF
--- a/gwpy/io/datafind.py
+++ b/gwpy/io/datafind.py
@@ -40,7 +40,6 @@ import re
 import warnings
 from functools import wraps
 from http.client import HTTPException
-from urllib.parse import urlparse
 
 from ligo.segments import (
     segment as LigoSegment,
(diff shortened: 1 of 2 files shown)


gwpy/io/datafind.py
```diff
@@ -40,14 +40,13 @@
...
 from http.client import HTTPException
-from urllib.parse import urlparse
 
 from ligo.segments import (
     segment as LigoSegment,
     segmentlist as LigoSegmentList,
 )
 from ..time import to_gps
-from .cache import (cache_segments, read_cache_entry, _iter_cache)
+from .cache import (cache_segments, read_cache_entry, _iter_cache, file_path)
```
alurban, line 49: Just a minor question, should these imports come from `.utils`?

duncanmmacleod: Yes, I have made that change with an amended commit, thanks.

PR #1275, opened 2020-11-06
EventTable.read(): modify API for `hdf5.snax` format

myNameIsPatrick:
This modifies the `channel` argument to be optional (now `channels`) and allows for more flexible behavior for reading SNAX files. In particular, this allows reading multiple channels at once from a single hdf5 file by providing a list of channels. If `channels` is not specified, this will attempt to read in all channels from the file.
    
In order to preserve the uniqueness across channels, a 'channel' column is added into the `EventTable` during the read process.
    
Also update relevant unit tests for all modified behavior as well as the docs on "Reading and writing Table and EventTable objects" in the SNAX section.

### DIFF
--- a/gwpy/table/io/snax.py
+++ b/gwpy/table/io/snax.py
@@ -19,6 +19,8 @@
 """Read events from SNAX
 """
 
+import warnings
+
 from astropy.io.misc.hdf5 import read_table_hdf5
 from astropy.table import vstack
 
@@ -33,7 +35,7 @@ __author__ = 'Patrick Godwin <patrick.godwin@ligo.org>'
 @read_with_columns
 @read_with_selection
 @with_read_hdf5
-def table_from_file(source, channels=None):
+def table_from_file(source, channels=None, on_missing="error"):
     """Read an `EventTable` from a SNAX HDF5 file
 
     Parameters
@@ -41,8 +43,15 @@ def table_from_file(source, channels=None):
     source : `h5py.File`
         the file path of open `h5py` object from which to read the data
 
-    channels : `str` or `list`
-        the channel(s) to read data from
+    channels : `str` or `list`, optional
+        the channel(s) to read data from. if no channels selected,
+        read in all channels
+
+    on_missing: `str`, optional
+        how to proceed when channels requested are missing. choose one from:
+            * ``'warn'``: emit a warning when missing channels are discovered
+            * ``'error'``: raise an exception
+        default is 'error'.
 
     Returns
     -------
@@ -57,7 +66,23 @@ def table_from_file(source, channels=None):
     if channels is None:
         channels = source.keys()
     else:
-        channels = set(channels) & set(source.keys())
+        channels = set(channels)
+        found = set(source.keys())
+        missing = channels - found
+        # check whether missing channels should
+        # be an error or simply a warning
+        if missing:
+            msg = "requested channels not found in SNAX file: '{}'".format(
+                "', '".join(missing),
+            )
+            if on_missing == "error":
+                raise ValueError(msg)
+            elif on_missing == "warn":
+                warnings.warn(msg)
+            else:
+                raise ValueError(
+                    'on_missing argument must be one of "warn" or "error"')
+        channels = channels & found
 
     # read data, adding in 'channel' column
     # to preserve uniqueness across channels
--- a/gwpy/table/io/snax.py
+++ b/gwpy/table/io/snax.py
@@ -35,7 +35,7 @@ __author__ = 'Patrick Godwin <patrick.godwin@ligo.org>'
 @read_with_columns
 @read_with_selection
 @with_read_hdf5
-def table_from_file(source, channels=None, on_missing="error"):
+def table_from_file(source, channels=None, on_missing="error", compact=False):
     """Read an `EventTable` from a SNAX HDF5 file
 
     Parameters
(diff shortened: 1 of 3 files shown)


gwpy/table/io/snax.py
```diff
@@ -33,26 +33,46 @@
...
+    # format channels appropriately
+    if isinstance(channels, str):
+        channels = [channels]
+
+    # only query channels contained in file
+    # if channels not specified, load all of them
+    if channels is None:
+        channels = source.keys()
+    else:
+        channels = set(channels) & set(source.keys())
```
duncanmmacleod, line 60: I'm not a fan of silently ignoring channels that were explicitly requested but aren't included in the file. I would rather this be an error by default.

You could consider adding a keyword `on_missing="error"` to allow users to downgrade to a warning if their channel is missing. Something like this:

```suggestion
        channels = set(channels)
        found = set(source.keys())
        missing = channels - found
        msg = "requested channels not found in SNAX file: '{}'".format(
            "', '".join(missing),
        )
        if missing and on_missing == "error":
            raise ValueError(msg)
        if missing and on_missing == "warn":
            warnings.warn(msg)
        channels = channels & found
```

myNameIsPatrick: Thanks @duncanmmacleod, I agree with all that and I'll add this in.

As an aside, after submitting this PR I realized how much space the 'channel' column takes in memory (somewhere between a `U30` and a `U40` which takes up around 150-200B per row). It's probably fine for typical use cases but would be expensive when you're reading many tables/channels which is typical for workflows like iDQ/hveto. I have an implementation that should get around that with some more work from the user (`compact`, False by default) by storing an integer representation of the channel and a `channel_map` in the table's metadata to map between the integer representation (really just `hash(channel`) and the full channel name).

Switching back involves something like:
```
table = table.group_by('channel')
for key, group in zip(table.groups.keys, table.groups):
    group['channel'] = table.meta['channel_map'][key['channel']]
```

First, how do you feel about something like that? That implementation is sufficient for my use case but I don't know if that's necessarily the best solution to expose in general, even if a user would have to explicitly toggle it on. Second, if you're favorable to that, would you want that as part of this PR or in a follow-up PR?

gwpy/table/tests/test_table.py
```diff
@@ -759,6 +759,19 @@ def test_read_snax(self):
                 filter_table(table, 'snr>.5')[('time', 'snr')],
             )
 
+            # check error handling when missing channels are encountered
+            missing = [channel, 'H1:MISSING']
+            with pytest.raises(ValueError):
+                self.TABLE.read(fp, channels=missing, format='hdf5.snax')
+
+            with pytest.warns(None):
```
duncanmmacleod, line 767: Why is `None` being used here, doesn't that function require a Warning type. I think this should be `UserWarning`
```suggestion
            with pytest.warns(UserWarning):
```

myNameIsPatrick: Thanks @duncanmmacleod, I just updated that. I originally went with `None` since I wasn't sure that `warnings.warn()` without specifying a specific warning always defaulted to a `UserWarning` across python versions. Putting `None` is a catch all for catching any warnings raised: https://docs.pytest.org/en/3.0.5/recwarn.html#recwarn.

myNameIsPatrick:
@duncanmmacleod, @alurban, I wanted to check in to see how I can help in moving this PR forward.

duncanmmacleod:
@myNameIsPatrick, sorry for sitting on this doing nothing, please rebase against the upstream `master` branch, then I'll take a closer look.

myNameIsPatrick:
Thanks @duncanmmacleod and no worries. I rebased against upstream so you should be good to look at this when you get a chance. I see a couple of CI 'Build and test' issues due to missing dependencies (seems unrelated) and a codeclimate issue on code complexity that I don't really understand, but otherwise everything else is passing.

I'd also like to get your thoughts on this when you get a chance: https://github.com/gwpy/gwpy/pull/1275#discussion_r523430886

duncanmmacleod:
> I see a couple of CI 'Build and test' issues due to missing dependencies (seems unrelated) and a codeclimate issue on code complexity that I don't really understand, but otherwise everything else is passing.

There's a problem with the metadata for the pycbc-1.18.0 conda package that is convincing `pip` to reinstall numpy, which in turn causes problems with astropy for some reason. I'm fixing it.

gwpy/table/io/snax.py
duncanmmacleod: So in your case, you need the `channel` information for each row, but are just trying to cut down on memory overhead? Or could you live without the channel column entirely?

myNameIsPatrick: That's right, the former. I would need the channel information if I'm requesting multiple channels since I'd like to find coincidences between a target channel and other channels. For a typical iDQ run, I'm requesting O(100) channels for O(day) stretch in time which can be problematic in terms of memory usage in some cases, but this would likely not be a concern for other typical use cases.

duncanmmacleod: I think I'm ok with the opt-in `compact` representation you have proposed - please implement it, then I can take a proper look and see what I really think.

myNameIsPatrick: @duncanmmacleod, just added this in the latest commit along with the relevant tests.

duncanmmacleod:
@myNameIsPatrick, can you rebase this again, which will pick up any upstream changes, and rerun the CI.

myNameIsPatrick:
@duncanmmacleod, I rebased this, but it looks like due to a security change I can't start the CI so it'll need to be approved by a maintainer.

duncanmmacleod:
@myNameIsPatrick, can you rebase this again then I can see about getting the CI running.

duncanmmacleod (approved):
LGTM, thanks @myNameIsPatrick!

PR #1276, opened 2020-11-10
Add --dt option to TimeSeries plot different times

areeda:
This is in response to a long standinf requests to be able compare time series of the same channel at different times.
An example is https://ldvw-dev.ligo.caltech.edu/ldvw/view?act=getImg&imgId=152709

It uses a switch on the command line (--dt) which forces all time series object to be plotted as though they have the same start time, with the legends extended to show the actual start time.

### DIFF
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -730,7 +730,6 @@ class CliProduct(object, metaclass=abc.ABCMeta):
             end = 0
             for ts in self.timeseries:
                 if start is not None:
-                    # NB: quantity truth value is deprecated
                     start = min(ts.t0, start)
                     end = max(ts.t0+ts.duration, end)
                 else:
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -729,7 +729,7 @@ class CliProduct(object, metaclass=abc.ABCMeta):
             start = None
             end = 0
             for ts in self.timeseries:
-                if start:
+                if start is not None:       # NB quantity truth value is deprecated
                     start = min(ts.t0, start)
                     end = max(ts.t0+ts.duration, end)
                 else:
--- a/gwpy/cli/cliproduct.py
+++ b/gwpy/cli/cliproduct.py
@@ -653,7 +653,6 @@ class CliProduct(object, metaclass=abc.ABCMeta):
         else:
             self.ax.grid(b=False)
 
-
     def save(self, outfile):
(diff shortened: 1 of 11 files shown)


gwpy/cli/cliproduct.py
```diff
@@ -729,7 +729,8 @@ def add_segs(self, args):
             start = None
             end = 0
             for ts in self.timeseries:
-                if start:
+                if start is not None:
+                    # NB: quantity truth value is deprecated
```
alurban, line 733:
```suggestion
```
I suggest removing this note, it's a simple enough change that doesn't touch any functionality that this isn't needed

alurban (changes requested):
@areeda, thanks for this, looks like it'll be a good addition and feature set.

After talking on the DetChar computing call, you asked me to post a couple of comments, which I've done below. One additional comment is: when plots are rendered with the new command-line flag, is it possible/worth it to tweak the x-axis to reflect generic start times?

gwpy/cli/timeseries.py
```diff
@@ -109,3 +135,14 @@ def scale_axes_from_data(self):
...
         self.plot.gca().autoscale_view(scalex=False)
+
+    @classmethod
+    def arg_channels(self, parser):
+        """Add an option to overlay different times rather than on the
+        same time scale"""
+        super().arg_channels(parser)
+        group = parser.add_argument_group(
+                'Data options time series', 'Allow overlay')
+        group.add_argument('--dt', action='store_true',
```
alurban, line 146: @areeda, my comment here doesn't seem to have been added, so I'll repost it:

Can you rename this flag? `--dt` gives the user the impression that they are changing the `TimeSeries.dt` attribute, i.e. that they're resampling the timeseries. Maybe something like `--start-time-offset`?

areeda: I will rename to something to distinguish it from the TimeSeries property dt

areeda:
Alex also mentioned changing the X-axis label, during our call. I'm looking into that while maintaing the auto-gps selection of units.

gwpy/cli/timeseries.py
areeda: Start time offset is too close to what the epoch option does. I think dt is short and clear but if you insist how  about --overlay-times?

alurban: That makes sense to me, `--overlay-times` sounds sensible

duncanmmacleod:
@areeda, please rebase this against upstream/master and resolve the merge conflicts.

PR #1278, opened 2020-11-16
Fix compatibility with matplotlib 3.3.3

### LINKED ISSUE #1277: gwpy is incompatible with Matplotlib 3.3.3
gwpy is incompatible with Matplotlib 3.3.3 because gwpy uses the non-public members `Axis._gridOnMinor` in `gwpy/plot/axes.py`. These non-public members were removed in Matplotlib 3.3.3. See matplotlib/matplotlib#18769.

duncanmmacleod:
This PR fixes #1277 by addressing the new incompatibility with matplotlib 3.3.3. I was hoping to find a better way to do this that doesn't just rely on a different private attribute, but didn't.

### DIFF
--- a/gwpy/plot/axes.py
+++ b/gwpy/plot/axes.py
@@ -81,8 +81,16 @@ def restore_grid(func):
     """
     @wraps(func)
     def wrapped_func(self, *args, **kwargs):
-        grid = (self.xaxis._gridOnMinor, self.xaxis._gridOnMajor,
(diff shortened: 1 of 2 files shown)


alurban (approved):
LGTM thanks @duncanmmacleod, if this is the simplest practical solution I think we can live with that.

duncanmmacleod:
This and #1279 are both needed to fix the failures, I suggest we merge that one first, and then rebase this one.

PR #1279, opened 2020-11-16
Fix compatibility with h5py-3.1.0

duncanmmacleod:
Minor change to `gwpy.timeseries` tests for compatibility with hdf5-3.1.0.


alurban (approved):
LGTM and sounds like a good plan, thanks @duncanmmacleod

PR #1281, opened 2020-11-24
Improve error reporting/handling when reading LIGO_LW Arrays

duncanmmacleod:
This PR fixes an issue seen by @rmacas when reading `<Array>` elements from LIGO_LW files. Fundamentally the problem was just that the wrong error message was printed when multiple `<Array>` elements were found and gwpy didn't know how to pick the right one (fixed in 288c3a20b1c578f977a909e7c8ebfdc185aa4949), but users probably require a wee bit more help to understand how to modify their code to avoid the error in the first place.


alurban (approved):
LGTM, thanks @duncanmmacleod

PR #1282, opened 2020-11-27
Change default spectral averaging method to 'median'

### LINKED ISSUE #1228: Change default PSD averaging method

alurban:
This PR fixes #1228 by changing the default spectral averaging method to `'median'` across the codebase. This includes changes to reference values in the unit tests for `gwpy.astro` to reflect median averaging.

### DIFF
--- a/gwpy/astro/range.py
+++ b/gwpy/astro/range.py
@@ -40,7 +40,7 @@ from ..utils.decorators import deprecated_function
 __author__ = 'Duncan Macleod <duncan.macleod@ligo.org>'
 __credits__ = 'Alex Urban <alexander.urban@ligo.org>'
 
-DEFAULT_FFT_METHOD = None
+DEFAULT_FFT_METHOD = "median"
 
 
 def _get_isco_frequency(mass1, mass2):
@@ -381,7 +381,7 @@ def range_timeseries(hoft, stride=None, fftlength=None, overlap=None,
         formats
 
     method : `str`, optional
-        FFT-averaging method, defaults to Welch's method, see
+        FFT-averaging method, defaults to median averaging, see
         :meth:`~gwpy.timeseries.TimeSeries.spectrogram` for
         more details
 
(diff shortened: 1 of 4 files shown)


alurban:
@duncanmmacleod, it may simplify your review to compare this to the diff in #1234

gwpy/astro/range.py
```diff
@@ -34,7 +34,7 @@
 __author__ = 'Duncan Macleod <duncan.macleod@ligo.org>'
 __credits__ = 'Alex Urban <alexander.urban@ligo.org>'
 
-DEFAULT_FFT_METHOD = None
+DEFAULT_FFT_METHOD = "median"
```
duncanmmacleod, line 37: @alurban, why is this repeated here and not just an import of `gwpy.timeseries.timeseries.DEFAULT_FFT_METHOD`?

duncanmmacleod: Ping @alurban.

areeda (approved):
This is a helpful change.

PR #1285, opened 2020-12-01
Allow generic window control for Rayleigh spectra

### LINKED ISSUE #1242: allowing selection of window in the Rayleigh_Spectrum and Rayleigh_Spectrogram
Current GWPY implementation does not allow to choose an specific window function when calculating the Rayleigh_Spectrum and Rayleigh_Spectrogram, and only the default Hann window is available.

alurban:
This PR fixes #1242 by allowing users to select generic window functions when calculating Rayleigh spectra. As with other spectral density estimation methods, any window format recognized by [`scipy.signal.get_window`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.get_window.html) is acceptable.

Note, Rayleigh spectra are given a hard 0 seconds of overlap between segments by default. Users will be able to either choose their own overlap, or pass `None` to get scipy's default based on the window function. This is reflected in various docstrings.

### DIFF
--- a/gwpy/signal/spectral/_scipy.py
+++ b/gwpy/signal/spectral/_scipy.py
@@ -90,7 +90,7 @@ for func in (welch, bartlett, median):
 
 # -- others -------------------------------------------------------------------
 
-def rayleigh(timeseries, segmentlength, noverlap=0):
+def rayleigh(timeseries, segmentlength, noverlap=0, window='hann'):
     """Calculate a Rayleigh statistic spectrum
 
     Parameters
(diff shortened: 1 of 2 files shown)


duncanmmacleod:
@alurban, this looks fine, can you confirm that this doesn't change the default output from before to after?

alurban:
Thanks @duncanmmacleod, and yes, I can confirm the default before was the Hann window (because that’s the default for scipy.signal.welch). This is also borne out by the unit tests, which use the default window before and after.

duncanmmacleod (approved):
LGTM, thanks @alurban.

alurban:
Thanks @duncanmmacleod, do you mind if I merge or are you intentionally holding off?

PR #1290, opened 2020-12-08
Resolve various warnings during build-and-test

### LINKED ISSUE #1289: Matplotlib colormaps have deprecated behavior
In downstream service software (particularly `gwsumm` and `gwdetchar.omega`) I am seeing the following warnings:

```
/path/to/lib/python3.8/site-packages/gwpy/plot/plot.py:395: MatplotlibDeprecationWarning: The 'cmap' parameter to Colorbar has no effect because it is overridden by the mappable; it is deprecated since 3.3 and will be removed two minor releases later.
    cbar = super().colorbar(mappable, **kwargs)
```

Matplotlib's [release notes](https://matplotlib.org/3.3.0/api/prev_api_changes/api_changes_3.3.0/deprecations.html#effectless-parameters-of-figure-colorbar-and-matplotlib-colorbar-colorbar) are somewhat confusing on the matter of how to address this, but AFAICT it appears we have to set the colormap from the mappable directly, something like:

```python
mappable.set_cmap(cmap)
cbar = ax.colorbar(mappable, **kwargs)
```

alurban:
This PR addresses several runtime warnings that arise during build-and-test:

* Do not duplicate the `'cmap'` parameter between the `mappable` and `ax.colorbar()` (see [matplotlib release notes](https://matplotlib.org/3.3.0/api/prev_api_changes/api_changes_3.3.0/deprecations.html#effectless-parameters-of-figure-colorbar-and-matplotlib-colorbar-colorbar))
* Log-scaled plot axes have deprecated kwargs (see [matplotlib release notes](https://matplotlib.org/3.3.1/api/api_changes.html#log-symlog-scale-base-ticks-and-nonpos-specification))
* The `numpy.ndarray.tostring()` method is deprecated as of 1.19, and its replacement `tobytes` has been available since 1.9 (see [numpy documentation](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tostring.html))
* Warnings about `numpy.ufunc` size changes are benign (they arise from downstream packages built against earlier numpy versions) and can be ignored, see [this stackoverflow post](https://stackoverflow.com/questions/40845304/runtimewarning-numpy-dtype-size-changed-may-indicate-binary-incompatibility)
* CLI plot products carry a redundancy in grid kwargs when the grid is turned off
* CLI coherencegrams carry an ambiguous truth value of an `astropy.units.Quantity`

Other warnings arise from an advertised change to the default spectral averaging method (addressed in #1282) and deprecated behavior related to `gwosc` and `uproot`, which can be handled in a separate PR. There are also upstream warnings arising from `glue.ligolw` running python-3.8, presumably because the package was built with an earlier version of python.

Note, this updates the Matplotlib version requirement to 3.3.0, numpy to 1.15.0 (for consistency with Matplotlib), and `python-ligo-lw` to 1.7.0.

This fixes #1289.

### DIFF
--- a/gwpy/io/ligolw.py
+++ b/gwpy/io/ligolw.py
@@ -533,10 +533,16 @@ def write_tables(target, tables, append=False, overwrite=False, **kwargs):
 
     # write file
     if isinstance(target, str):
-        kwargs.setdefault('compress', target.endswith('.gz'))
+        kwargs.setdefault(
+            'compress',
+            "gz" if target.endswith('.gz') else False,
+        )
         ligolw_utils.write_filename(xmldoc, target, **kwargs)
     elif isinstance(target, FILE_LIKE):
-        kwargs.setdefault('compress', target.name.endswith('.gz'))
+        kwargs.setdefault(
+            'compress',
+            "gz" if target.name.endswith('.gz') else False,
+        )
         ligolw_utils.write_fileobj(xmldoc, target, **kwargs)
 
 
--- a/gwpy/io/ligolw.py
+++ b/gwpy/io/ligolw.py
@@ -531,19 +531,22 @@ def write_tables(target, tables, append=False, overwrite=False, **kwargs):
     # convert table to format
     write_tables_to_document(xmldoc, tables, overwrite=overwrite)
 
-    # write file
-    if isinstance(target, str):
-        kwargs.setdefault(
-            'compress',
-            'gz' if target.endswith('.gz') else False,
-        )
-        ligolw_utils.write_filename(xmldoc, target, **kwargs)
-    elif isinstance(target, FILE_LIKE):
-        kwargs.setdefault(
-            'compress',
-            'gz' if target.name.endswith('.gz') else False,
-        )
-        ligolw_utils.write_fileobj(xmldoc, target, **kwargs)
+    # find writer function and target filename
+    if isinstance(target, FILE_LIKE):
+        writer = ligolw_utils.write_fileobj
+        name = target.name
+    else:
+        writer = ligolw_utils.write_filename
+        name = str(target)
+
+    # handle gzip compression kwargs
+    if name.endswith('.gz') and writer.__module__.startswith('glue'):
+        kwargs.setdefault('gz', True)
+    elif name.endswith('.gz'):
+        kwargs.setdefault('compress', 'gz')
+
+    # write XML
+    writer(xmldoc, target, **kwargs)
 
 
 # -- utilities ----------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


gwpy/io/ligolw.py
```diff
@@ -533,10 +533,16 @@ def write_tables(target, tables, append=False, overwrite=False, **kwargs):
 
     # write file
     if isinstance(target, str):
-        kwargs.setdefault('gz', target.endswith('.gz'))
+        kwargs.setdefault(
+            'compress',
+            "gz" if target.endswith('.gz') else False,
+        )
```
duncanmmacleod, line 539: I have a feeling that this will cause errors on systems that have `lscsoft-glue` and _not_ `python-ligo-lw` (since `glue.ligolw.utils.write_filename` is still expecting `gz`), but I would have hoped that the unit tests should have caught this.

duncanmmacleod: Also, can you please forgive me for asking you to pick one style of quotes for both lines?

alurban: https://youtu.be/hou0lU8WMgo

alurban: @duncanmmacleod, would you prefer that I do some sort of try-except statement?

gwpy/io/ligolw.py
```diff
@@ -533,10 +533,16 @@ def write_tables(target, tables, append=False, overwrite=False, **kwargs):
...
+            'gz' if target.endswith('.gz') else False,
+        )
         ligolw_utils.write_filename(xmldoc, target, **kwargs)
     elif isinstance(target, FILE_LIKE):
-        kwargs.setdefault('gz', target.name.endswith('.gz'))
+        kwargs.setdefault(
+            'compress',
+            'gz' if target.name.endswith('.gz') else False,
+        )
         ligolw_utils.write_fileobj(xmldoc, target, **kwargs)
```
duncanmmacleod, line 546: To support both I think I would rewrite this as

```python
    # find writer function and target filename
    if isinstance(target, FILE_LIKE):
        writer = ligolw_utils.write_fileobj
        name = target.name
    else:
        writer = ligolw_utils.write_filename
        name = str(target)

    # handle gzip compression kwargs
    if name.endswith('.gz') and writer.__module__.startswith('glue'):
        kwargs.setdefault('gz', True)
    elif name.endswith('.gz'):
        kwargs.setdefault('compress', 'gz')

    # write XML
    writer(xmldoc, target, **kwargs)

alurban: Good idea @duncanmmacleod, I've done this in the latest commit

duncanmmacleod:
@alurban, please rebase.

alurban:
@duncanmmacleod, done

duncanmmacleod (approved):
LGTM, thanks @alurban.

Issue #1292, opened 2020-12-14
inspiral_range should use cosmologically-corrected ranges


jrollins:
It looks like the inspiral_range function will now be using the [inspiral_range package](https://git.ligo.org/gwinc/inspiral-range) to calculate ranges, which is good.  However, it looks like it's still just calculating the "sensemon" range, which is inaccurate at cosmological distances.  It would be good if by default it used the cosmologically-corrected [range](https://git.ligo.org/gwinc/inspiral-range/-/blob/master/inspiral_range/inspiral_range.py#L205) function, or at least make the other functions available to users so they're not limited to just the sensemon range.

Another possibility would be to deprecate the gwpy.astro.inspiral_range package and to just point users to the external inspiral_range package, which may in some way be cleaner and less confusing.  At the very least this package should be referenced for people looking for a more accurate range estimate.

duncanmmacleod:
I believe this will be resolved by #1248.

jrollins:
@duncanmmacleod That is the PR I was referring to.  But it looks like it will be replacing the existing sensemon range calculation with the one from the inspiral_range package.  It's the sensemon_range calculation itself that I think should be deprecated, or at least the cosmologically-corrected ranges should be offered (or just advertise the inspiral_range package itself).

duncanmmacleod:
Ah right, I understand now. @alurban, are you able to include the modification to use the cosmologically-corrected range function in your PR?

Maybe include a wrapper around `sensemon_range` for users who prefer it (for whatever reason)?

duncanmmacleod:
@jrollins, we probably can't just tell people to use the `inspiral_range` functions directly because we want to munge the input and output into `gwpy` objects. But the wrappers should be very light and all of the docstrings and documentation pages should defer credit to the `inspiral_range` package.

jrollins:
I'm not worried about credit, just flexibility and not having to duplicate too much code.  A light wrapper around inspiral_range sounds reasonable, something that gives as much access as possible to all the functionality of that library.  Or maybe there are modifications to the inspiral_range that could make it more interoperable with gwpy without having to maintain a whole interoperability module in gwpy...

alurban:
[...]

PR #1293, opened 2020-12-17
Refactor gwpy.astro.inspiral_range to sensemon_range

alurban:
This PR refactors two functions under `gwpy.astro` to more appropriate namespaces:

* `inspiral_range --> sensemon_range`
* `inspiral_range_psd --> sensemon_range_psd`

This also includes a number of minor enhancements for efficiency.

The old functions are now placeholders which (via decoration) raise a `DeprecationWarning` so they can be replaced by more refined models that incorporate cosmological corrections in gwpy-2.2.0. This is related to #1292 and should be merged **well before** #1284.


gwpy/astro/range.py
```diff
@@ -21,22 +21,48 @@
...
 from math import pi
+from scipy.integrate import trapz
 
-from scipy import integrate
-
-from astropy import (units, constants)
+from astropy import (
+    units,
+    constants,
+)
```
duncanmmacleod, line 32: Please reinstate the blank line between the standard library imports and the `scipy` import.

gwpy/astro/range.py
```diff
@@ -75,9 +101,31 @@ def decorated_func(psd, *args, **kwargs):
...
+            "this estimator currently uses the LIGO SenseMon (LIGO-T030276), "
+            "which will be replaced by a model with cosmological corrections "
+            "in gwpy-2.2.0.\n\n"
+            "The proper SenseMon method is now gwpy.astro.{}".format(
+                func.__name__),
+            DeprecationWarning,
+        )
+        target.__doc__ = func.__doc__
+        return target
+    return decorated_func
```
duncanmmacleod, line 117: This looks like a reimplementation of the `gwpy.utils.decorators.deprecated_function` decorator, please consider using that instead.

gwpy/astro/range.py
```diff
@@ -180,46 +215,46 @@ def inspiral_range(psd, snr=8, mass1=1.4, mass2=1.4, fmin=None, fmax=None,
...
-    mass2 = units.Quantity(mass2, 'solMass').to('kg')
-    mtotal = mass1 + mass2
-
-    # compute ISCO
-    fisco = (constants.c ** 3 / (constants.G * 6**1.5 * pi * mtotal)).to('Hz')
-
+    fisco = _get_isco_frequency(mass1, mass2)
     # format frequency limits
-    fmax = units.Quantity(fmax or fisco, 'Hz')
+    fmin = units.Quantity(fmin or psd.df.value, 'Hz')  # avoid DC value
```
duncanmmacleod, line 233: I think we should still be using `psd.df` here in case the units aren't `Hz`.

duncanmmacleod (changes requested):
This looks good, and should set things up for #1284 nicely, however, one comment (possibly requiring a change) and one likely change required.

gwpy/astro/range.py
alurban: Turns out you were right about this, this has been fixed in the latest commit

gwpy/astro/range.py
alurban: Aha! This has been addressed, #1294 will need to go in first

gwpy/astro/range.py
alurban: Done

alurban:
@duncanmmacleod, this one is now ready and rebased against #1294

duncanmmacleod (approved):
This is ready.

PR #1294, opened 2021-01-11
Fix deprecated_function decorator

duncanmmacleod:
This PR fixes the handling of custom messages being passed to the `gwpy.utils.decorators.deprecated_function` decorator - this was clearly never tested properly.

### DIFF
--- a/gwpy/utils/tests/test_decorators.py
+++ b/gwpy/utils/tests/test_decorators.py
@@ -49,3 +49,30 @@ def test_return_as_error():
     with pytest.raises(ValueError) as exc:
         myfunc('test')
     assert 'failed to cast return from myfunc as int: ' in str(exc.value)
+
+
+def test_deprecated_function():
+    """Test that `deprecated_function` works without a message
+
+    (and without any functional brackets)
+    """
+    @decorators.deprecated_function
+    def myfunc(value):
+        return str(value)
+
+    with pytest.warns(DeprecationWarning) as record:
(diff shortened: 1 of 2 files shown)


gwpy/utils/tests/test_decorators.py
```diff
@@ -49,3 +49,30 @@ def myfunc(value):
...
+def test_deprecated_function_message():
+    """Test that `deprecated_function` works with a message
+    """
+    @decorators.deprecated_function(message="don't use {0.__name__}")
+    def myfunc(value):
+        return str(value)
+
+    with pytest.warns(DeprecationWarning) as record:
+        myfunc('test')
+    assert record[0].message == "don't use myfunc"
```
alurban, line 78: I think this `message` attribute needs to be converted to a string:
```suggestion
    assert str(record[0].message) == "don't use myfunc"
```

alurban (changes requested):
LGTM @duncanmmacleod, I can verify this works in testing #1293, just one minor comment to resolve build tests.

gwpy/utils/tests/test_decorators.py
duncanmmacleod: I could have sworn that worked for me locally, maybe only on the latest greatest pytest version (or python 3.9). Anyways, fixed, thanks for the heads up.

alurban (approved):
LGTM, thanks @duncanmmacleod

PR #1295, opened 2021-01-21
Add a Histogram product to gwpy-plot

areeda:
I've wanted to have this for a while.  It's a good way to look at the noise distributions. See [this set of plots](https://ldvw.ligo.caltech.edu/ldvw/view?act=imagehistory&strt=0&pageNum=1&usrSel=Joseph+Areeda+%284392%29&group=gaussian+noise+histograms+6&size=med&op_group=Favorites&newGrpName=%3Cnew+group+name%3E&pageNum2=1) comparing colored Gaussian noise to our Strain channel (LVK credentials needed)

### DIFF
--- a/gwpy/cli/histogram.py
+++ b/gwpy/cli/histogram.py
@@ -31,6 +31,7 @@ __myname__ = 'histogram.py'
 import math
 import sys
 from scipy.stats import kurtosis, norm, skew
+from astropy.units import Quantity
 
 from .cliproduct import TimeDomainProduct
 from ..plot import Plot
@@ -45,11 +46,12 @@ class Histogram(TimeDomainProduct):
 
     def __init__(self, args):
         super().__init__(args)
-        self.mean = None
-        self.std = None
-        self.skew = None
-        self.kurtosis = None
-        self.nbins = None
+        self.args = args
+        self.mean = Quantity(0, '')    # value needed for pytests
+        self.std = Quantity(0, '')
+        self.skew = Quantity(0, '')
+        self.kurtosis = 0
+        self.nbins = 0
         self.ymin = sys.maxsize
         self.ymax = 0
 
@@ -67,14 +69,15 @@ class Histogram(TimeDomainProduct):
         """
         units = self.units
         ret = 'Timeseries amplitude'
-        if len(units) == 1 and str(units[0]) == '':  # dimensionless
-            pass
-        elif len(units) == 1 and self.usetex:
-            ret += ' [{}]'.format(units[0].to_string('latex'))
-        elif len(units) == 1:
-            return ' ({})'.format(units[0].to_string())
-        elif len(units) > 1:
-            ret += ' (multiple units)'
+        if isinstance(units, list):
+            if len(units) == 1 and str(units[0]) == '':  # dimensionless
+                pass
+            elif len(units) == 1 and self.usetex:
+                ret += ' [{}]'.format(units[0].to_string('latex'))
+            elif len(units) == 1:
+                return ' ({})'.format(units[0].to_string())
+            elif len(units) > 1:
+                ret += ' (multiple units)'
         return ret
 
     @classmethod
@@ -83,13 +86,11 @@ class Histogram(TimeDomainProduct):
         """
         return 'Count'
 
-    @classmethod
     def get_suptitle(self):
         """Start of default super title, first channel is appended to it
         """
         return 'Histogram: {0}'.format(self.chan_list[0])
 
-    @classmethod
     def get_title(self):
         suffix = super().get_title()
         # limit significant digits for minute trends
@@ -106,7 +107,6 @@ class Histogram(TimeDomainProduct):
                    self.skew, self.kurtosis)
         return ret
 
-    @classmethod
     def make_plot(self):
         """Generate the plot from time series and arguments
         """
@@ -161,14 +161,12 @@ class Histogram(TimeDomainProduct):
 
         return plot
 
-    @classmethod
     def _finalize_arguments(self, args):
         """ We cannot guess at default plot params
         until histogram is calculated"""
         if args.xscale is None:  # set default x-axis scale
             args.xscale = 'linear'
 
-    @classmethod
     def scale_axes_from_data(self):
         """Restrict data limits for Y-axis based on what you can see
         """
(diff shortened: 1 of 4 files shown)


gwpy/cli/histogram.py
```diff
@@ -0,0 +1,187 @@
...
+
+import math
+import sys
+from scipy.stats import kurtosis, norm, skew
+
+from .cliproduct import TimeDomainProduct
+from ..plot import Plot
+from ..plot.tex import label_to_latex
+import warnings
+
```
alurban, line 39: `flake8` complains about two blank lines:

```suggestion


```

gwpy/cli/histogram.py
```diff
@@ -0,0 +1,187 @@
...
+        self.skew = None
+        self.kurtosis = None
+        self.nbins = None
+        self.ymin = sys.maxsize
+        self.ymax = 0
+
+    @classmethod
+    def init_cli(self, parser):
+        """We add an option to add a fit ot gaussian PDF"""
+        super().init_cli(parser) # add all timeseries options
```
alurban, line 58: `flake8` complains about two whitespaces before an inline comment:

```suggestion
        super().init_cli(parser)  # add all timeseries options
```

gwpy/cli/histogram.py
```diff
@@ -0,0 +1,187 @@
...
+        """Generate the plot from time series and arguments
+        """
+        plot = Plot(figsize=self.figsize, dpi=self.dpi)
+        ax = plot.gca(xscale='linear')
+
+        # handle user specified plot labels
+        if self.args.legend:
+            nlegargs = len(self.args.legend[0])
+        else:
+            nlegargs = 0
```
alurban, line 118: This can be done in one line:

```suggestion
        nlegargs = len(self.args.legend[0]) if self.args.legend else 0
```

gwpy/cli/histogram.py
```diff
@@ -0,0 +1,187 @@
...
+        group.add_argument('--gauss', action='store_true',
+                           help='Add gaussian fit')
+
+    @classmethod
+    def get_xlabel(self):
+        """Text for x-axis label
+        """
+        units = self.units
+        if len(units) == 1 and str(units[0]) == '':  # dimensionless
+            return 'Sample value'
```
alurban, line 69: It seems like lines 68-69 can be omitted because they fall into the use case on line 76, right?

```suggestion
```

gwpy/cli/histogram.py
```diff
@@ -0,0 +1,187 @@
...
+                          'There are {:d} series and {:d} legends'.format(
+                            len(self.timeseries), len(self.args.legend)))
+            nlegargs = 0    # don't use  them
+
+        for i in range(0, self.n_datasets):
+            series = self.timeseries[i]
+            if nlegargs:
+                label = self.args.legend[0][i]
+            else:
+                label = series.channel.name
```
alurban, line 132: This can be done on one line:

```suggestion
            label = self.args.legend[0][i] if nlegargs else series.channel.name
```

gwpy/cli/tests/test_histogram.py
```diff
@@ -0,0 +1,38 @@
...
+
+"""Unit tests for :mod:`gwpy.cli.histogram`
+"""
+
+from ... import cli
+from .base import (_TestTimeDomainProduct, update_namespace)
+
+
+class TestCliTimeSeries(_TestTimeDomainProduct):
+    TEST_CLASS = cli.TimeSeries
```
alurban, line 27: Looks like CircleCI is seeing a failure related to class definitions:

```
self = <gwpy.cli.tests.test_histogram.TestCliTimeSeries object at 0x7fd8860f6400>
prod = <gwpy.cli.timeseries.TimeSeries object at 0x7fd885b9d340>

    def test_action(self, prod):
        if self.ACTION is None:
            raise NotImplementedError
>       assert prod.action is self.ACTION
E       AssertionError

/usr/local/lib/python3.8/site-packages/gwpy/cli/tests/base.py:170: AssertionError
```

Is there a simple fix for this?

alurban (changes requested):
@areeda, this is going to be an awesome feature and methinks will prove extremely useful to LDVW users. I have a few minor suggestions for changes, most of which include specific line suggestions.

One more general comment: if I read your plots correctly, the x-axis (i.e., the thing being histogram'd) represents the amplitude of a timeseries, is that right? If so, I think a better x-axis label would be something like "Timeseries amplitude", "Timeseries amplitude [unit]", etc.

gwpy/cli/tests/test_histogram.py
areeda: I copied that from the time series test, maybe I missed something. Checking

gwpy/cli/histogram.py
areeda: Thanks Alex. I'll make both suggested changes

gwpy/cli/histogram.py
areeda: I don't think that's true. I rewrote that method to make it clearer. The situation is when a dimensionless channel has an emtpy string for units we want the label to be `Timeseries amplitude` without the separate test it would come out as `Timeseries amplitude[]`
I think it's better code now (not saying much)

gwpy/cli/tests/test_histogram.py
areeda: Found that I believe I was not consistent trying to test the histogram product with the cli.TimeSeries class.

gwpy/cli/histogram.py
areeda: These both now use the conditional assignment

gwpy/cli/histogram.py
areeda: I agree

gwpy/cli/histogram.py
areeda: fixed

gwpy/cli/tests/test_histogram.py
areeda: I made enough mods to get the tests to pass

gwpy/cli/tests/test_histogram.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) Duncan Macleod (2018-2021)
```
duncanmmacleod, line 2: Please update the copyright, I didn't write this file:

```suggestion
# Copyright (C) CSUF (2021)
```

duncanmmacleod (changes requested):
There's a bunch of formatting/style changes I would like to see in here. I don't think there's anything functional - if you tell me it works, I'm happy.

PR #1302, opened 2021-02-25
Migrate CI to GitHub Actions (2)

duncanmmacleod:
Duplicate of #1301 from a branch on the main repo.

PR #1303, opened 2021-02-26
Fix handling of figsize=None in matplotlib 3.4.0

duncanmmacleod:
This PR fixes an incompatibility with the release candidate for matplotlib-3.4.0 in which it seems that `figsize=None` is actively passed a lot more to figure constructors. Luckily, it's an almost trivial fix.


duncanmmacleod:
This needs to be rebased when #1302 is merged to pick up the new test suite.

alurban (approved):
LGTM and the build tests are passing, thanks @duncanmmacleod

PR #1304, opened 2021-02-26
Add new gwpy.testing.errors module to provide error-handling utilities for the test suite

duncanmmacleod:
This PR adds a new `gwpy.testing.errors` module that provides a decorator function to catch and apply `pytest.skip` when common network errors are found - we typically don't want to fail the suite because of these. This allows us to clean up a bunch of repeated usage through the test suite.


alurban (approved):
LGTM and this is a much appreciated reorganization/enhancement, thanks @duncanmmacleod

PR #1309, opened 2021-03-03
Add basic support for uproot4

duncanmmacleod:
This PR adds basic support for `uproot >=4`. It's hacky because uproot 4 doesn't support writing ROOT files yet, so there's an import hook in place to advise users to install `uproot3` to do that if they need to.


alurban (approved):
LGTM and I'm personally happy to see this change as it's related to #1290, thanks @duncanmmacleod

PR #1310, opened 2021-03-08
Install texlive on macos and ubuntu during CI

duncanmmacleod:
This PR adds a texlive installation step to the github actions workflows for the full build-and-test.


alurban (approved):
LGTM, this is a very beneficial change methinks, thank you @duncanmmacleod

PR #1312, opened 2021-03-09
Simplify Axes.scatter to better handle arguments

duncanmmacleod:
This PR simplifies the overloaded `Axes.scatter` method as follows:

- handle an `s` (size) array as well, seaborn 0.11.1 expects these to be acceptable as positional arguments,
- renamed `c_sort{=True}` to `sortbycolor=False`, to have a better name and have the same default action as matplotlib's `Axes.scatter`

@alurban @eagoetz, this probably has implications for the summary pages, which I think rely upon the default sorting by colour.

### DIFF
--- a/gwpy/plot/axes.py
+++ b/gwpy/plot/axes.py
@@ -176,27 +176,38 @@ class Axes(_Axes):
 
     # -- overloaded plotting methods ------------
 
-    def scatter(self, x, y, c=None, **kwargs):
-        # scatter with auto-sorting by colour
-        try:
-            if c is None:
-                raise ValueError
-            c_array = numpy.asanyarray(c, dtype=float)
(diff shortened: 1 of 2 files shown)


eagoetz:
@duncanmmacleod thanks for including me in the loop. I hope @alurban will let me shadow him as he would follow up on any checks needed before giving review approval. This begs the question, once Alex leaves, will you be requesting me as a reviewer for issues such as this? I'm happy to go for it, but I still feel rather green and will need time to spin up to Alex-level-expertise.

duncanmmacleod:
@alurban, @eagoetz, can someone please approve this?

alurban (approved):
LGTM, thanks @duncanmmacleod

PR #1317, opened 2021-04-22
Fix NDS2Warnings in TimeSeries tests

duncanmmacleod:
This PR fixes a test suite failure introduced by #1290 that an `NDSWarning` isn't emitted when you can actually use gwdatafind. This also updates the `LOSC_` variables in the timeseries test suite to use `GWOSC_` as the prefix.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,6 +112,33 @@ jobs:
             texlive-latex-extra \
         ;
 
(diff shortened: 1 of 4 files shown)


duncanmmacleod:
I think this is ready to go, as soon as #1320 is merged to include 10bb7bb.

alurban (approved):
LGTM, thanks @duncanmmacleod

PR #1318, opened 2021-04-22
Replace use of numpy alias type with builtin

duncanmmacleod:
This PR replaces `numpy.complex` with `numpy.complex128` to resolve a deprecation warning from numpy 1.20.0, see
https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated.


alurban (approved):
Seems pretty open-and-shut, thanks @duncanmmacleod

PR #1319, opened 2021-04-22
Hardcode PSD method in all TimeSeries tests

duncanmmacleod:
This PR hardcodes the PSD `method` argument in all appropriate `TimeSeries` test functions, mainly to resolve deprecation warnings, but also so that the tests are a bit more robust.

cc @alurban


alurban (approved):
LGTM, what’s the status on changing the default averaging method overall?

duncanmmacleod:
> LGTM, what’s the status on changing the default averaging method overall?

That's scheduled in #1282, which (in principle) can be merged any time.

PR #1329, opened 2021-05-06
Add --average-method command-line option for gwpy-plot products

duncanmmacleod:
This PR adds a `--average-method` command-line option for the relevant `gwpy-plot` products, which defaults to the `DEFAULT_FFT_METHOD` from `gwpy.timeseries.timeseries`. This is currently `None` (`'welch'`) but is due to change in #1282 to `'median'`.


hoyc1 (commented):
This looks good to me!

hoyc1 (approved):

duncanmmacleod:
@areeda, can you please review this?

areeda (approved):
It's not clear to me what methods are available. I think listing them or linking to external docs in the --average-method help would be useful. But not necessary

PR #1330, opened 2021-05-06
Enable FFT options for gwpy-plot qtransform product

duncanmmacleod:
This PR enables the FFT command-line options for the `gwpy-plot` qtransform product to allow specification of fftlength/overlap/method for the whitening ASD calculation.

This is dependent upon #1329.


areeda (approved):
LGTM

PR #1339, opened 2021-06-03
Fix detection of uproot4 writing capability

duncanmmacleod:
The latest uproot 4.x includes `create()` and `recreate()` functions, but don't include the `newtree` function that we need to translate numpy arrays into ROOT-serialisable objects, so this PR just updates the `_import_uproot_that_can_write_root_files()` function to check for `newtree`.


hoyc1 (commented):
This looks good to me!

hoyc1 (approved):

PR #1341, opened 2021-06-09
Add custom pytest marks

duncanmmacleod:
This PR adds support for custom, registered pytest marks. The only mark added at the moment is `cvmfs`, which has been applied to all of those tests that attempt to use the GWOSC CVMFS repo.

### DIFF
--- a/gwpy/conftest.py
+++ b/gwpy/conftest.py
@@ -30,6 +30,8 @@ use('agg', force=True)
 
 # register custom fixtures for all test modules
 from .testing.fixtures import *  # noqa: E402,F401,F403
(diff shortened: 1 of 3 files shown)


eagoetz (approved):
@duncanmmacleod Myself, @areeda, and Robert had a look at this PR today and it looks like you're extending the decorators for functions relying on CVMFS, but we don't fully understand everything about how this is supposed to work. It feels like the decorators are getting quite nested in a way that it is not clear what they are doing, but it looks like you know what you are doing. This is also in the test functions, so we don't want to hold up the next release over this matter.

We approve.

PR #1342, opened 2021-06-09
Handle CVMFS failures more gracefully

duncanmmacleod:
This PR adds some error-handling to skip over CVMFS-related failures in the test suite.

There's currently an issue with gwosc.osgstorage.org that is failing the tests, but since this isn't really a GWpy issue we should skip those failures.


hoyc1 (commented):
Looks good to me!

hoyc1 (approved):

PR #1298, opened 2021-02-04
Problem with TimeSeries.get fixed for reduced channels

areeda:
This trivial update resolves issue #1297 
The channel type needed by NDS2 is "rds" not "reduced".

The problem is demonstrated with:
```
from gwpy.timeseries import TimeSeries
ts = TimeSeries.get('L1:DCS-CALIB_STRAIN_CLEAN_SUB60HZ_C01,rds', 1252939480, 1252939490, verbose=True)
```

PR #1343, opened 2021-06-09
Dynamically generate NDS2 enums

duncanmmacleod:
This PR refactors the Enum generation in `gwpy.io.nds2` to dynamically generate the `Nds2ChannelType` and `Nds2DataType` enums if possible using the `nds2.channel` object itself. This should mean that the enums are always an accurate reflection of the underlying NDS2 interface, if present.

This does present a minor breaking change in that the enums no longer overload the `name` property, instead providing a `nds2name` property instead.

### DIFF
--- a/gwpy/detector/channel.py
+++ b/gwpy/detector/channel.py
@@ -265,7 +265,7 @@ class Channel(object):
         if type_ is None:
             self._type = None
         else:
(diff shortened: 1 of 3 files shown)


duncanmmacleod:
cc @areeda, this is relevant to #1298.

areeda:
I test with gwpy-plot and it looks good

PR #1345, opened 2021-06-21
Fix enable for imperial units

duncanmmacleod:
This PR fixes the enabling of the `imperial` units from astropy. astropy/astropy#11204 removed `imperial` from the `astropy.units` namespace, but it's still importable if you go at it directly.


hoyc1 (approved):

hoyc1 (commented):
Looks good to me!

PR #1346, opened 2021-06-21
Update CI to run on main branch and run some jobs nightly

duncanmmacleod:
This PR makes two small changes to the Github Actions CI

- run all jobs on the `main` branch - this just means that if we rename `master` to `main` we don't have to update the CI again
- run the `Dependencies` workflow on a cron schedule every night - this should allow us to catch breaking changes in dependencies as early as possible, rather than 'whenever the next person submits a PR'

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,10 +13,12 @@ name: Build and test
 on:
   push:
     branches:
+      - main
       - master
       - release/**
   pull_request:
     branches:
+      - main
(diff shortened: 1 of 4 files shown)


duncanmmacleod:
Needs to be rebased after #1345 is merged.

hoyc1 (approved):

hoyc1 (commented):
Looks good to me!

PR #1347, opened 2021-06-21
Support output kwarg in gwpy.signal.filter_design.notch()

duncanmmacleod:
This PR adds support for specifying the `output` keyword for the `gwpy.signal.filter_design.notch` function, allowing notches to be designed in SOS format.


hoyc1 (approved):

hoyc1 (commented):
Looks good to me!

PR #1349, opened 2021-06-22
Migrate pytest configuration to pyproject.toml

duncanmmacleod:
This PR migrates the pytest configuration from `setup.cfg` to `pyproject.toml`, and then adds a few warnings that aren't our problem.

This is extracted from #1331 to get it through quickly so that it can be used elsewhere.


hoyc1 (approved):

hoyc1 (commented):
Looks good to me!

PR #1351, opened 2021-06-22
Fail tests on any warning

duncanmmacleod:
This PR modifies the pytest configuration to fail a test on any warnings, forcing us to stay on top of upstream dependency changes, etc.

This will likely need to stay open in 'Draft' form for a while until all of the different warnings are addressed.

This closes #1331 which should be split up into separate PRs that fix specific warnings.

Depends #1484.


duncanmmacleod:
Migrated to https://gitlab.com/gwpy/gwpy/-/merge_requests/1351.

PR #1352, opened 2021-06-22
Fix physical_type str formatting for astropy 4.3+

duncanmmacleod:
This PR addresses a warning emitted by Astropy 4.3+ relating to formatting of `<unit>.physical_type`, which used to just be a `str`, but is now a structured object that complains about using `str` methods.


hoyc1 (approved):

hoyc1 (commented):
Looks good to me!

PR #1353, opened 2021-06-22
Deprecate TemporaryFile context manager and tmpfile test fixture

duncanmmacleod:
This PR deprecates the `gwpy.testing.utils.TemporaryFile` context manager function and the `gwpy.testing.fixtures.tmpfile` test fixture. These are basically no longer required now that we can rely properly on the [`tmp_path`](https://docs.pytest.org/en/stable/tmpdir.html?) fixture from pytest.

### DIFF
--- a/gwpy/cli/tests/base.py
+++ b/gwpy/cli/tests/base.py
@@ -19,7 +19,6 @@
 """Unit tests for :mod:`gwpy.cli`
 """
 
-import os
 from argparse import ArgumentParser
 from unittest import mock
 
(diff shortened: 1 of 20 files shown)


eagoetz:
@duncanmmacleod we understand this is still a draft and that you may edit this further after #1362 gets merged. We'll come back to this with you once that is done.

duncanmmacleod:
@gwpy/maintainers this is now ready for review.

pesummary (approved):
LGTM.

hoyc1 (approved):

PR #1356, opened 2021-06-22
Remove s6 segment server query support

### LINKED ISSUE #1350: Remove support for S6-style segment database
The LIGO S6-style segment databases are no longer operational, so we should remove the support in the codebase to read from them.

duncanmmacleod:
This PR closes #1350 by removing all support for querying an S6-style DB2 database using `lscsoft-glue`.

cc @rpfisher, @robertbruntz


areeda (approved):
LGTM

PR #1357, opened 2021-06-22
Remove unnecessary import protection for python < 3

duncanmmacleod:
This PR removes a lingering `ImportError` catch relating to `unittest.mock`; we don't support python < 3 any more.


eagoetz (approved):
LGTM

PR #1360, opened 2021-06-25
Support loading files from Path objects

lpsinger:


duncanmmacleod:
This has been subsumed into #1362.

duncanmmacleod:
Thanks for the contribution!

PR #1362, opened 2021-06-29
Fix pathlib compatibility in file I/O

duncanmmacleod:
This PR fixes up a number of functions to enable reading from `pathlib.Path` objects as well as `file` and `str`. This probably supersedes #1360.

### DIFF
--- a/gwpy/detector/io/omega.py
+++ b/gwpy/detector/io/omega.py
@@ -25,7 +25,7 @@ from collections import OrderedDict
 
 from astropy.io import registry
 
-from ...io.cache import FILE_LIKE
+from ...io.utils import with_open
 from .. import (Channel, ChannelList)
 
 __author__ = 'Duncan Macleod <duncan.macleod@ligo.org>'
@@ -36,6 +36,7 @@ WPIPELINE = OMEGA_LOCATION and os.path.join(OMEGA_LOCATION, 'bin', 'wpipeline')
 
 # -- read ---------------------------------------------------------------------
 
+@with_open
 def read_omega_scan_config(source):
     """Parse an Omega-scan configuration file into a `ChannelList`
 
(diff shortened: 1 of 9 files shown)


duncanmmacleod:
Ping @gwpy/maintainers, please review this.

gwpy/io/utils.py
```diff
@@ -98,6 +100,50 @@ def gopen(name, *args, **kwargs):
...
+
+    >>> @with_open
+    >>> def my_func(pathorfile, *args, **kwargs)
+    >>>     ...
+
+    To ensure that the second argument (position 1) is a file open for writing:
+
+    >>> @with_open(mode="r", pos=1)
+    >>> def my_func(stuff, pathorfile, *args, **kwargs)
+    >>>     stuff.write_to(pathorfile, *args, **kwargs)
```
eagoetz, line 130: This example is a bit confusing: which is readable (`pathorfile`?) and which is writable(`stuff`?)? Perhaps give a bit more explanation here for non-experts.

gwpy/io/utils.py
```diff
@@ -98,6 +100,50 @@ def gopen(name, *args, **kwargs):
...
+            if not isinstance(args[pos], FILE_LIKE):
+                with open(args[pos], mode=mode) as fobj:
+                    args = list(args)
+                    args[pos] = fobj
+                    return func(*args, **kwargs)
+            return func(*args, **kwargs)
+        return wrapped_func
+    if func:
+        return _decorator(func)
+    return _decorator
```
eagoetz, line 144: Can you add some in-line code comments here? We're not able to follow exactly what is going on.

gwpy/io/utils.py
eagoetz: It looks like mode in the example should be `w`, or `rw`, not `r` (unless we don't understand something fundamental...)

eagoetz:
Trying to leave comments but github is preventing us from leaving comments. We will come back to this

eagoetz:
@duncanmmacleod We think this is basically fine, though our understanding of exactly the changes to `gwpy/io/tests/test_utils.py` are not well understood, but there is a potential bug in `gwpy/detector/io/clf.py` line 141 (see comment).

gwpy/io/utils.py
duncanmmacleod: I've added inline comments to explain the logic internal to the decorated function, the outer scaffolding is just the standard structure to define a decorator that can be called with or without functional brackets, i.e. allowing both

```python
@with_open
def thing():
    ...
```

and

```python
@with_open(mode="w")
def thing():
    ...
```

duncanmmacleod:
> @duncanmmacleod We think this is basically fine, though our understanding of exactly the changes to `gwpy/io/tests/test_utils.py` are not well understood,

That change is just pytest fanciness, there isn't actually anything substantive in the changes to what the tests actually do.

> but there is a potential bug in `gwpy/detector/io/clf.py` line 141 (see comment).

This has been corrected, good catch.

I hope this is now ready to go.

eagoetz (approved):
LGTU

gwpy/detector/io/omega.py
```diff
@@ -56,30 +57,18 @@ def read_omega_scan_config(source):
...
+        if not line or line.startswith('#'):
+            continue
+        if line.startswith('['):
+            section = line[1:-1]
+        elif line.startswith('{'):
+            append(parse_omega_channel(source, section))
+        else:
+            raise RuntimeError(
+                "Failed to parse Omega config line: {!r}".format(line),
+            )
```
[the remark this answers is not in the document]
eagoetz: @duncanmmacleod just want to make sure you saw this

PR #1363, opened 2021-07-02
Handle cases in EventTable.cluster() where table is empty or clustered multiple times

myNameIsPatrick:
This PR handles a few edge cases more gracefully when calling `EventTable.cluster()`:

1. The table is empty.
2. The table is already clustered.

In both cases, the error encountered looks like:

```
>>> table.cluster('time', 'snr', 0.1)
  File "/home/patrick.godwin/workspace/projects/gwpy/gwpy/table/table.py", line 782, in cluster
    padded_sublists = [numpy.append(s, numpy.array([s[-1]+1]))
  File "/home/patrick.godwin/workspace/projects/gwpy/gwpy/table/table.py", line 782, in <listcomp>
    padded_sublists = [numpy.append(s, numpy.array([s[-1]+1]))
IndexError: index -1 is out of bounds for axis 0 with size 0
```

In both cases, the original table is now returned as a result. I have also added a test which checks that running `table.cluster()` with the same parameters twice returns the same table.

### DIFF
--- a/gwpy/table/table.py
+++ b/gwpy/table/table.py
@@ -779,14 +779,23 @@ class EventTable(Table):
         if window <= 0.0:
             raise ValueError('Window must be a positive value')
 
+        # If no rows, no need to cluster
+        if len(self) == 0:
+            return self.copy()
+
         # Generate index and rank vectors that are ordered
         orderidx = numpy.argsort(self[index])
         col = self[index][orderidx]
         param = self[rank][orderidx]
 
         # Find all points where the index vector changes by less than window
-        # and divide the resulting array into clusters of adjacent points
         clusterpoints = numpy.where(numpy.diff(col) <= window)[0]
+
+        # If no such cluster points, no need to cluster
+        if len(clusterpoints) == 0:
+            return self.copy()
+
+        # Divide points into clusters of adjacent points
         sublists = numpy.split(clusterpoints,
                                numpy.where(numpy.diff(clusterpoints) > 1)[0]+1)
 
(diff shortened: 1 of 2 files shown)


myNameIsPatrick:
After looking this over again, I realize I am returning the same table rather than a new table in both cases. I should easily be able to modify it depending on what the desired behavior is.

gwpy/table/table.py
```diff
@@ -779,14 +779,23 @@ def cluster(self, index, rank, window):
         if window <= 0.0:
             raise ValueError('Window must be a positive value')
 
+        # If no rows, no need to cluster
+        if len(self) == 0:
+            return self
```
duncanmmacleod, line 784: I like this short-circuit, however the documented behvaiour is to return a _new_ table, so we might need to make a copy:

```suggestion
            return self.copy()
```

gwpy/table/table.py
```diff
@@ -779,14 +779,23 @@ def cluster(self, index, rank, window):
...
         col = self[index][orderidx]
         param = self[rank][orderidx]
 
         # Find all points where the index vector changes by less than window
-        # and divide the resulting array into clusters of adjacent points
         clusterpoints = numpy.where(numpy.diff(col) <= window)[0]
+
+        # If no such cluster points, no need to cluster
+        if len(clusterpoints) == 0:
+            return self
```
duncanmmacleod, line 796: As above, maybe this:

```suggestion
            return self.copy()
```

duncanmmacleod (changes requested):
I think this is great, sorry it has taken a ludicrous amount of time to review, just one pedantic change.

myNameIsPatrick:
Thanks @duncanmmacleod and no worries. I've changed it so it returns copies in both cases now. I also noticed that one of the newly introduced lines wasn't covered by the test suite, so I added an additional test to handle the empty table case.

PR #1368, opened 2021-07-14
Improve unit parsing for GWOSC tables

duncanmmacleod:
This PR improves the auto-parsing of units in event tables downloaded from GWOSC as part of `EventTable.fetch_open_data`. In practice this means that the `_lower` and `_upper` columns now share the same unit as the main column for the relevant parameter, previously those columns didn't have a unit.


eagoetz (approved):
LGTU

PR #1369, opened 2021-07-14
Improve auto-detection of time-like columns in tables

duncanmmacleod:
This PR introduces a refactoring of the internal method `EventTable._get_time_column` in an attempt to simplify the logic. This method is used mainly in plotting to try and automatically detect that a column represents 'time' and to use a GPS-scaled axis for it.


areeda (approved):
Looks good to me

PR #1371, opened 2021-07-16
Refactor `Series.is_compatible`

duncanmmacleod:
This PR refactors the `Series.is_compatible` method by separating internal logic into two functions, and then improve the tests. This also adds one small improvement that irregular `Series` can now be compatible if their X-index arrays match identically.


eagoetz:
@duncanmmacleod While we think we know what is going on with this PR, we would really appreciate if you could assist with some higher-bandwidth discussion on Zoom leading us through the changes. Thanks!

areeda (approved):
I approved it but I wonder if it may be too strict by demanding exact data type match. I wonder if one type can be converted error free to another if they are compatible eq: int16 and int32?
It's not not clear to me whether numpy/astropy deal with that situation seamlessly.  Your call, of course.

PR #1373, opened 2021-07-22
Redesign online documentation using sphinx-material

### LINKED ISSUE #1327: Hard to read examples on gwpy docs
When looking at the following page https://gwpy.github.io/docs/latest/timeseries/io#hdf5-gwosc on Safari/Chrome the white and yellow text on a grey background is hard to read.

duncanmmacleod:
This PR introduces a revamp of the online documentation using [`sphinx-material`](https://bashtage.github.io/sphinx-material/) in place of the current `sphinx_bootstrap_theme`.

An example can be viewed by LIGO.ORG account holders here:

https://ldas-jobs.ligo.caltech.edu/~duncan.macleod/gwpy/sphinx-material/

I would appreciate feedback on the new design from anyone, especially @gwpy/maintainers.

EDIT: fixes #1327


myNameIsPatrick:
Looks pretty nice! I like that all the sections are visible no matter where you are in the docs.

One thing I noticed browsing through the API docs is that in some cases if you scroll up, it'll create a second navbar row rather than replacing the text in the navbar. Example here: https://ldas-jobs.ligo.caltech.edu/~duncan.macleod/gwpy/sphinx-material/api/gwpy.table.EventTable/#gwpy.table.EventTable

duncanmmacleod:
> Looks pretty nice! I like that all the sections are visible no matter where you are in the docs.
> 
> One thing I noticed browsing through the API docs is that in some cases if you scroll up, it'll create a second navbar row rather than replacing the text in the navbar. Example here: https://ldas-jobs.ligo.caltech.edu/~duncan.macleod/gwpy/sphinx-material/api/gwpy.table.EventTable/#gwpy.table.EventTable

@myNameIsPatrick, thanks. Yes, the second navbar row is a feature designed to show the lineage of the page, so you can click back up the hierarchy with ease, but I don't like it either and haven't figured out an elegant way to live with it.

areeda (approved):
I like the new look

PR #1376, opened 2021-08-03
Refactor datafind.find_frametype function

duncanmmacleod:
This PR refactors the `gwpy.io.datafind.find_frametype` function, which was a beastly monolith, and is now a little tamer. This also includes a few fixes for the following

- gracefully handle failing to read the channel list (TOC) from a GWF file (`RuntimeError`)
- add Virgo types to ranking system
- some optimisations for reading cache files

This should have zero impact on users, other than hiding errors.

### DIFF
--- a/gwpy/io/datafind.py
+++ b/gwpy/io/datafind.py
@@ -38,6 +38,7 @@ import os
 import os.path
 import re
 import warnings
+from collections import defaultdict
 from functools import wraps
 from http.client import HTTPException
 
@@ -45,6 +46,7 @@ from ligo.segments import (
     segment as LigoSegment,
     segmentlist as LigoSegmentList,
 )
+
 from ..time import to_gps
 from .cache import (cache_segments, read_cache_entry, _iter_cache)
 from .gwf import (num_channels, iter_channel_names)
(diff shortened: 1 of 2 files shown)


gwpy/io/datafind.py
```diff
@@ -132,7 +136,10 @@ def _read_ffl_cache(self, site, tag):
             def _update_metadata(entry):
                 return type(entry)(site, tag, entry.segment, entry.path)
             with open(path, 'r') as fobj:
-                cache = list(map(_update_metadata, _iter_cache(fobj)))
+                cache = list(map(
+                    _update_metadata,
+                    _iter_cache(fobj, gpstype=float),
```
eagoetz, line 141: Just a curiosity question: can `gpstype` be any valid python type (or `LIGOTimeGPS) or must it be only one of `LIGOTimeGPS` or `int`?

eagoetz (approved):
It's complicated, but we think it looks ok. LGTU

gwpy/io/datafind.py
duncanmmacleod: Officially it can be anything, but in practice you would only want `LIGOTimeGPS`, `float`, or `int`.

PR #1378, opened 2021-08-06
Improve gwpy.cli example handling in docs

duncanmmacleod:
This PR improves the handling/rendering of the `gwpy.cli` examples as part of the documentation. I think the changes basically boil down to:

- only run the example if it has changed or the output doesn't exist yet
- resolve a couple of sphinx warnings


areeda (approved):
I only understand about 95% of the code but that part looks good and I trust the rest.

PR #1379, opened 2021-08-09
Remove usage of tag keyword in calls to TimeSeries.fetch_open_data

duncanmmacleod:
This PR removes all usage of the deprecated `tag` keyword in calls to `TimeSeries.fetch_open_data` - this should resolve a bunch of `DeprecationWarning`s emitted when building the documentation.


hoyc1 (approved):
This looks good to me!

PR #1380, opened 2021-08-09
Improve usage of pathlib.Path objects

duncanmmacleod:
This PR replaces any calls to `pathlib.Path.open()` with simpler calls to `pathlib.Path.read_text` as appropriate. It's just cleaner.


areeda (approved):
learned the new call. Hx

PR #1385, opened 2021-08-09
Execute the examples as part of the test suite in CI

duncanmmacleod:
This PR adds an argument to `pytest` to execute the examples as part of the full `build-and-test` suite run. This used to happen, but I think got lost during the transition to github actions.


hoyc1 (commented):
Looks good to me!

hoyc1 (approved):

PR #1386, opened 2021-08-10
Fix datafind trend discovery

duncanmmacleod:
This PR fixes an apparently critical problem with `gwpy.io.datafind.find_frametype` that _may_ have been introduced by #1376 where trend requests weren't properly parsed, so the wrong trend type may have been selected.

This was flagged by the `test_find_best_frametype_ligo` test function, but that only runs during the cron nightly build at Caltech, which isn't part of the github CI.

### DIFF
--- a/gwpy/io/datafind.py
+++ b/gwpy/io/datafind.py
@@ -361,7 +361,7 @@ def _parse_ifos_and_trends(chans):
     for name in chans:
         chan = Channel(name)
         try:
-            found.add((chan.ifo[0], chan.trend))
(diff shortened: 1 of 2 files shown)


eagoetz:
@duncanmmacleod I'm guessing you've checked but didn't write it: this PR passes the test you mention at Caltech, correct?

duncanmmacleod:
@eagoetz, yes, the test now passes when run at Caltech.

hoyc1 (approved):
Looks good to me!

PR #1387, opened 2021-08-10
Use 'dev' as the version for development documentation builds

duncanmmacleod:
This PR modifies the sphinx configuration to just use `dev` as the version number when building outside of tags. This is needed since the tags aren't always synchronised to the development branch (since we use a `release/X.Y.z` branch for bug-fix releases). E.g. the current dev docs say `2.0.1` because the 2.0.2-2.0.4 tags aren't on `master`.


eagoetz (approved):
LGTM

PR #1388, opened 2021-08-10
Fix a critical bug in the documentation configuration

duncanmmacleod:
This PR fixes a bug in the use of `ConfigParser.get` introduced by #1378: the `fallback` argument to `ConfigParser.get` can _only_ be used as a keyword argument.


eagoetz (approved):
LGTM

PR #1391, opened 2021-08-13
Add support for filter functions to use multiple columns with `EventTable.filter`

duncanmmacleod:
This PR modifies `gwpy.table.filter` to support slicing multiple columns out of an `EventTable` and passing it into a filter function. I also updated the tests and beefed up the documentation to demonstrate the existing and the new functionality.

### DIFF
--- a/docs/table/filter.rst
+++ b/docs/table/filter.rst
@@ -11,80 +11,146 @@ extract portions of a table based on some criteria, this is called filtering.
 The `EventTable` object comes with a :meth:`~EventTable.filter` method
 that provides an intuitive interface to down-selecting rows in a table.
 
-To demonstrate, first we create a catalogue of gravitational-wave detections using data available from `LOSC <https://losc.ligo.org/>`_::
+To demonstrate, we can download |GWTC-2| from |GWOSC|:
 
-    >>> from gwpy.table import EventTable
-    >>> table = EventTable.read(
-    ...     """name      gps           m1      m2         snr   distance network
-    ...        GW150914  1126259462.00 36.2    23.7       23.7  420      HL
-    ...        LVT151012 1128678900.44 23      13         9.7   440      HL
-    ...        GW151226  1135136350.65 14.2    7.5        13    880      HL
-    ...        GW170104  1167559936.60 31.2    19.4       13    1000     HL
-    ...        GW170814  1186741861.53 30.5    25.3       18    540      HLV""",
-    ...     format='ascii')
+.. code-block:: python
+
+   >>> from gwpy.table import EventTable
+   >>> events = EventTable.fetch_open_data("GWTC-2")
+   >>> print(events)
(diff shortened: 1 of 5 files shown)


docs/table/filter.rst
```diff
@@ -11,80 +11,146 @@ extract portions of a table based on some criteria, this is called filtering.
...
+   GW190426_152155-v1          0.32 ... 1240327333.3                    None
+          GW190814-v2          0.06 ... 1249852257.0                     1.1
+   GW190929_012149-v1          0.34 ... 1253755327.5                    33.6
 
 ======================
 Using multiple filters
 ======================
 
-Filters can be trivially chained (either in `str` form, or functional form)::
+Filters can be trivially chained (either in `str` form, or functional form):
```
eagoetz, line 131: Minor suggestion: remove the word "trivially" as the word doesn't really aid the documentation, and can feel demoralizing to someone who might struggle to make use of the multiple filtering

gwpy/table/filter.py
```diff
@@ -205,12 +205,14 @@ def _flatten(container):
...
-        and len(tup) == 3
-        and isinstance(tup[0], str)
-        and callable(tup[1])
-    )
+    try:
+        names, func, args = tup
+        return (
+            (isinstance(names, str) or all(isinstance(x, str) for x in names))
+            and callable(func)
+        )
```
eagoetz, line 213: Seems like the logic conditionals have changed here. Was there something incorrect about the original and/or what is the motivation for the updated version?

docs/table/filter.rst
duncanmmacleod: Agree 100%, will change.

gwpy/table/filter.py
duncanmmacleod: The latter, the first two conditions are now taken care of by the tuple unpacking, and the rest is just to support the first element of the tuple being a list of names.

eagoetz: Sounds good, thanks for confirming

gwpy/table/filter.py
```diff
@@ -251,6 +253,6 @@ def filter_table(table, *column_filters):
     """
     keep = numpy.ones(len(table), dtype=bool)
     for name, op_func, operand in parse_column_filters(*column_filters):
-        col = table[name].view(numpy.ndarray)
+        col = table[name]
```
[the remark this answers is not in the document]
duncanmmacleod: Removing the `view` is required to ensure that the `op_func` receives a formatted `Column` object (or a `Table` slice), rather than just an array; this in turn is a requirement of the new documented functionality.

duncanmmacleod:
@eagoetz, any further review comments?

gwpy/table/filter.py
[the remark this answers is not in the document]
eagoetz: Ok, thanks

eagoetz (approved):
LGTM

PR #1392, opened 2021-08-13
Update 'LOSC' references to 'GWOSC'

duncanmmacleod:
This PR updates references to LOSC to use 'GWOSC', which is the newer name. This includes renaming the `hdf5.losc` I/O format to `hdf5.gwosc`, but the old name is retaining in a deprecated manner (with tests added to check the deprecation warning is emitted).


hoyc1 (approved):
Looks good to me!

PR #1393, opened 2021-08-13
Refactor GravitySpyTable

duncanmmacleod:
This PR refactors the `GravitySpyTable` code to simplify the methods and remove the requirement on Pandas. This has been replaced by a requirement on `requests`, which is fine since `gwosc` already requires `requests`, and `gwpy` requires `gwosc`.

cc @scottcoughlin2014

### DIFF
--- a/gwpy/table/gravityspy.py
+++ b/gwpy/table/gravityspy.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) Scott Coughlin (2017-2020)
+# Copyright (C) Scott Coughlin (2017-2021)
+#               Cardiff University (2021)
 #
 # This file is part of GWpy.
 #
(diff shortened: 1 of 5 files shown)


duncanmmacleod:
@scottcoughlin2014, this is finally actually ready for a review if you are able to sign off. 🦃

gwpy/table/gravityspy.py
```diff
@@ -19,22 +20,83 @@
...
+    row,
+    outdir,
+    training_set=False,
+    labelled_samples=False,
+):
+    """Construct where to download a GravitySpy image
+
+    This returns a `str.format`-style template that just needs
+    the duration to form a complete target path.
+    """
```
areeda, line 83: Took me a while to figure out outdir must be a Path object and / is an overload for join. Defining the argument data types might save the next person some time

areeda (approved):
Looks good to me

gwpy/table/gravityspy.py
```diff
@@ -69,187 +138,129 @@ def download(self, **kwargs):
...
-        output = mp_utils.multiprocess_with_queues(
+        download_path = Path(download_path)
+
+        # work out which columns contain URLs
+        url_columns = list(filter(URL_COLUMN.match, self.colnames))
+
+        # construct a download target location for each requested image
+        urls = dict()
+        for row in self:
+            if not self["url1"]:  # skip rows without URLs
```
scottcoughlin2014, line 155: Hey Duncan,

thanks for doing this! just need to change this line 155 to

```
            if row["url1"] == "?":  # skip rows without URLs
                continue
```

scottcoughlin2014: This will let the download skip any rows that do not have a public url associated with them in the databases/table.

scottcoughlin2014 (changes requested):
Other than that one line, things seem good!

gwpy/table/gravityspy.py
duncanmmacleod: @scottcoughlin2014, do you mean

```suggestion
            if not row["url1"]:  # skip rows without URLs
```

to just fix the typo, or 

```suggestion
            if row["url1"] == "?":  # skip rows without URLs
```

to actually change the refence 'no url' value? If the latter should the test actually be

```suggestion
            if self["url1"] in {None, "", "?"}:  # skip rows without URLs
```

to catch actual empty values?

scottcoughlin2014: Hey @duncanmmacleod , I think it would be most robust to be

`            if row["url1"] in {None, "", "?"}:  # skip rows without URLs`

duncanmmacleod: This change has been made.

PR #1394, opened 2021-08-16
Catch KeyError when filtering errors from parse-conda-requirements.py

duncanmmacleod:
This PR improves the error handling in the `parse-conda-requirements.py` script. When an UnsatisfiableError is raised by `conda install` there won't be a `'packages'` key in the JSON output, so we need to catch `KeyError` then run it again to expose the underlying problem. See build errors in #1366 for an example of why this is needed.


hoyc1 (approved):
Looks good to me!

PR #1396, opened 2021-08-19
Handle default ADC slope of 0. in new frameCPP versions

duncanmmacleod:
This PR fixes a failing test when running with the latest versions of `ldas-tools-framecpp`, in which `FrAdcData` structures are created with a `DEFAULT_SLOPE` of `0.`, with no way to change the slope from the Python bindings. This means that if you write a GWF from Python that contains ADC channels, the slope you read back is _always_ `0.`. The patch here is to replace a default of `0.` with `1.` if the bias is also at its default (`0.`).

See https://git.ligo.org/ldastools/LDAS_Tools/-/issues/114

### DIFF
--- a/gwpy/timeseries/io/gwf/framecpp.py
+++ b/gwpy/timeseries/io/gwf/framecpp.py
@@ -312,6 +312,10 @@ def read_frdata(frdata, epoch, start, end, scaled=True,
         slope = None
         bias = None
     else:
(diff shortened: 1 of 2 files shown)


pesummary (approved):
Looks good to me!

hoyc1:
> Looks good to me!

Sorry, I was logged in with the wrong account.

hoyc1 (approved):

PR #1397, opened 2021-08-24
Fix compatibility with matplotlib-3.5.0

duncanmmacleod:
This PR fixes a couple of instances of `assert ax.lines == []`, which doesn't work any more as of matplotlib-3.5.0b1. This is easy fixed by comparing only the lengths of the objects, which is independently defined for the `ArtistList` object.

### DIFF
--- a/gwpy/spectrogram/tests/test_spectrogram.py
+++ b/gwpy/spectrogram/tests/test_spectrogram.py
@@ -144,7 +144,7 @@ class TestSpectrogram(_TestArray2D):
         with rc_context(rc={'text.usetex': False}):
             plot = array.plot(method=method)
             ax = plot.gca()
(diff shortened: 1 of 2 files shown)


hoyc1 (approved):
Looks good to me! You could possibly simplify `assert len(ax.lines) == 0` to `assert not len(ax.lines)` which might be easier on the eye but this is your choice.

duncanmmacleod:
> Looks good to me! You could possibly simplify `assert len(ax.lines) == 0` to `assert not len(ax.lines)` which might be easier on the eye but this is your choice.

This was considered, but `len(ax.lines) == 0` was chosen for visual consistency with the other asserted comparisons in those test functions. Thanks @hoyc1!

PR #1411, opened 2021-08-27
Deprecate gwpy.utils.shell.which

duncanmmacleod:
~~This PR modifies `gwpy.utils.shell.which` to use `shutil.which` instead of `distutils.spawn.find_executable`, which doesn't work as well on windows.~~

This PR deprecates the `gwpy.utils.shell.which` function, people should use use `shutil.which` directly.

### DIFF
--- a/gwpy/io/kerberos.py
+++ b/gwpy/io/kerberos.py
@@ -31,8 +31,6 @@ import subprocess
 import sys
 from collections import OrderedDict
 
-from ..utils.shell import which
-
 __author__ = "Duncan Macleod <duncan.macleod@ligo.org>"
 
 __all__ = ['kinit']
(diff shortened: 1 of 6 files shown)


gwpy/io/kerberos.py
```diff
@@ -49,12 +47,18 @@ class KerberosError(RuntimeError):
...
 
 
-def kinit(username=None, password=None, realm=None, exe=None, keytab=None,
-          krb5ccname=None, verbose=None):
-    """Initialise a kerberos (krb5) ticket.
+def kinit(
+        username=None,
+        password=None,
+        realm=None,
+        exe="kinit",
```
eagoetz, line 54: Just checking: unless I missed something, is it intended behaviour that the default changed from `exe=which('kinit')` to `exe='kinit'`? Those seem like two different cases for `exe`

duncanmmacleod: @eagoetz, this is a deliberate change. `subprocess` effectively does a `which` call internally anyway so I decided to remove it and let `subprocess` handle it.

eagoetz (commented):
@duncanmmacleod LGTM, but check the comment in `gwpy/io/kerberos.py`. I'll approve once confirmed

eagoetz (approved):
LGTM

PR #1414, opened 2021-09-17
Allowing cropping of irregular Series

mattpitkin:
I would like to be able to crop irregular TimeSeries objects, but because the `dx` attribute is not defined for these it doesn't work. This PR makes a change to the `crop` method of the `Series` object, so that cropping can be performed on irregularly sampled series.

### DIFF
--- a/gwpy/types/series.py
+++ b/gwpy/types/series.py
@@ -978,19 +978,37 @@ class Series(Array):
                  % type(self).__name__)
             end = None
 
+        # check if series is irregular
+        try:
+            _ = self.dx
+            irregular = False
+        except AttributeError:
+            irregular = True
+
         # find start index
         if start is None:
             idx0 = None
         else:
-            idx0 = int((xtype(start) - x0) // self.dx.value)
+            if not irregular:
+                idx0 = int((xtype(start) - x0) // self.dx.value)
+            else:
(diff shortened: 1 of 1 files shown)


gwpy/types/series.py
```diff
@@ -978,19 +978,41 @@ def crop(self, start=None, end=None, copy=False):
                  % type(self).__name__)
             end = None
 
+        # check if series is irregular
+        try:
+            _ = self.dx
+            irregular = False
+        except AttributeError:
+            irregular = True
```
duncanmmacleod, line 986: There isn't a need to assign the `self.dx` to a variable in this case:

 ```suggestion
        try:
            self.dx
        except AttributeError:
            irregular = True
        else:
            irregular = False
```

duncanmmacleod (changes requested):
Thanks for this @mattpitkin. I've requested one small pedantic change to the code you have proposed, but would also love to see a new unit test to cover the new use case.

mattpitkin:
@duncanmmacleod Sure, I'm happy to add a unit test.

mattpitkin:
I've added a unit test. Hopefully codecov will agree that this no longer reduced the coverage of the package.

duncanmmacleod (approved):
LGTM, thanks @mattpitkin.

PR #1419, opened 2021-10-13
Use uproot4 for writing ROOT files as well

duncanmmacleod:
This PR updates the ROOT I/O interface to use `uproot` (v4) for writing ROOT files, rather than relying up uproot3.

Depends on #1441.

### DIFF
--- a/gwpy/table/io/root.py
+++ b/gwpy/table/io/root.py
@@ -75,33 +75,11 @@ def table_from_root(source, treename=None, **kwargs):
             return Table(trees[treename].arrays(namedecode="utf-8"), **kwargs)
 
 
-def _import_uproot_that_can_write_root_files():
-    """uproot v4 can't handle writing files (ATOW),
-    so try it and fall back to uproot3
-    """
-    import uproot
-    try:
-        uproot.newtree
-    except AttributeError:
-        try:
-            import uproot3 as uproot
-        except ModuleNotFoundError as exc:  # pragma: no cover
-            exc.msg = (
-                "you have uproot {} installed, which does not support writing "
-                "ROOT files (yet), please install uproot3 "
-                "(see https://pypi.org/project/uproot3/)".format(
-                    uproot.__version__,
-                )
-            )
-            raise
-    return uproot
-
-
 def table_to_root(table, filename, treename="tree",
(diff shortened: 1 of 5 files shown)


duncanmmacleod:
This can be rebased after #1441 is merged.

gwpy/table/io/root.py
```diff
@@ -75,47 +75,38 @@ def table_from_root(source, treename=None, **kwargs):
...
-            "uproot currently doesn't support appending to existing files",
-        )
+    import uproot
 
-    tree = uproot.newtree(dict(table.dtype.descr), **kwargs)
+    createkw = {
+        k: kwargs.pop(k) for k in {
+            "initial_directory_bytes",
+            "uuid_function",
+        } if k in kwargs
```
eagoetz, line 94: Just checking that the change in the `kwargs` is not problematic. I don't know the details

gwpy/table/io/root.py
```diff
@@ -75,47 +75,38 @@ def table_from_root(source, treename=None, **kwargs):
...
+        create_func = uproot.create
 
     with create_func(filename, **createkw) as outf:
-        outf[treename] = tree
-        outf[treename].extend(dict(table.columns))
+        tree = outf.mktree(
+            treename,
+            dict(table.dtype.descr),
+            **kwargs,
+        )
```
eagoetz, line 108: Is this just a difference with uproot4? I guess it's reasonable, but I don't really know.

duncanmmacleod: Yes this is just a feature of uproot4 vs uproot3.

gwpy/table/io/root.py
duncanmmacleod: This is a necessary evil of this migration. I don't think these kwargs are ever used in the wild.

eagoetz: Ok, thanks for confirming. I'll resolve this then

gwpy/table/io/root.py
eagoetz: Sounds good.

eagoetz (approved):
By no means am I an expert in this area, but I think everything seems reasonable. Duncan answered my questions about some details.

PR #1423, opened 2021-10-18
Remove ldas-tools-framecpp from conda extra in setup.cfg

duncanmmacleod:
This PR removes a redundant entry from the `dev` extra spec in `setup.cfg`; this should have been removed as part of #1396.


eagoetz (approved):
LGTM

PR #1428, opened 2021-10-25
Resolve "Bug in Series index comparison"

### LINKED ISSUE #1426: Bug in Series index comparison
In the [`_compare_index()`](https://github.com/gwpy/gwpy/blob/master/gwpy/types/series.py#L661) method of the [`Series`](https://github.com/gwpy/gwpy/blob/master/gwpy/types/series.py#L37) class I think there is a bug in the [`except`](https://github.com/gwpy/gwpy/blob/master/gwpy/types/series.py#L680) clause. Both `idxa` and `idxb` get set to the same values, but I think one of them should be set from `other` rather than `self`.

mattpitkin:
Resolves #1426.

### DIFF
--- a/gwpy/types/series.py
+++ b/gwpy/types/series.py
@@ -664,8 +664,8 @@ class Series(Array):
         Raises
         ------
         ValueError
-            if ``dx`` doesn't match, or ``xindex`` doesn't match
-            (as appropriate)
+            if ``dx`` doesn't match, or ``xindex`` values are not present/are
+            identical (as appropriate)
         """
         try:  # check step size, if possible
             _delta = "d{}".format(axis)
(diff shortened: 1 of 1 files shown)


gwpy/types/series.py
```diff
@@ -680,11 +680,11 @@ def _compare_index(self, other, axis="x"):
...
             idxa = getattr(self, _index, None)
-            idxb = getattr(self, _index, None)
+            idxb = getattr(other, _index, None)
             if (
-                idxa is None  # index on 'other' but not on 'self'
-                or idxb is None  # index on 'self' but not on 'other'
-                or not numpy.array_equal(idxa, idxb)  # indexes don't match
+                idxa is None  # no index on 'self'
+                or idxb is None  # no index on 'other'
+                or numpy.array_equal(idxa, idxb)  # indexes are identical
```
duncanmmacleod, line 687: For series with a manual index, there isn't a clean concept of 'appending' or 'prepending' so I don't really like the idea that we are presuming that is the use case here. Purely for this function (which is independent of the prepend/append idea) I would like to keep the concept that equal indexes means 'compatible'.

Can you please revert the change on this one line? This may mean adding a different extra check in the relevant prepend/append function.

duncanmmacleod (changes requested):

duncanmmacleod (approved):
LGTM, thanks @mattpitkin!

Issue #1432, opened 2021-11-22
Incompatible with astropy 5.0


dethodav:
The changes to the astropy I/O architecture that were introduced in 5.0 are not compatible with the current version of gwpy. There may be other issues, but this is the first one we identified. Details on the new astropy I/O are here: 

https://docs.astropy.org/en/stable/whatsnew/5.0.html#whatsnew-5-0-io-unified

This mainly shows up as import errors, but there may be deeper issues. For example, the imports here return errors: 

https://github.com/gwpy/gwpy/blob/3a740c920b78001495eccd417ae5f1fd2ad789e4/gwpy/io/registry.py#L28-L34

This issue can be reproduced by running this import line with astropy 5.0.

PR #1433, opened 2021-11-22
Pin the experimental dependencies check to Python 3.9

duncanmmacleod:
The experimental dependencies check has been failing for a while because the check is running on Python 3.10 which doesn't have any lalsuite wheels yet. For now we need to pin back to Python 3.9. This may expose #1432.


hoyc1 (approved):
LGTM

PR #1435, opened 2021-11-23
Update gwpy.io to support Astropy 5 class-based registries

### LINKED ISSUE #1432: Incompatible with astropy 5.0

duncanmmacleod:
This PR fixes #1432 by updating the use of a private function from `astropy.io.registry`. This is still a hack, but is the easiest thing to do while attempting to support the old registry format and the new class-based system at the same time.

### DIFF
--- a/gwpy/io/registry.py
+++ b/gwpy/io/registry.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) Duncan Macleod (2014-2020)
+# Copyright (C) Louisiana State University (2014-2017)
+#               Cardiff University (2017-2021)
 #
 # This file is part of GWpy.
 #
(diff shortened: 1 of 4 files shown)


eagoetz (approved):
LGTM

PR #1438, opened 2021-11-24
Update compatibility with matplotlib 3.5.0

duncanmmacleod:
This PR addresses a number of `DeprecationWarning`s from matplotlib 3.5.0.

### DIFF
--- a/gwpy/plot/legend.py
+++ b/gwpy/plot/legend.py
@@ -45,9 +45,9 @@ class HandlerLine2D(legend_handler.HandlerLine2D):
         self._linewidth = linewidth
 
     def create_artists(self, *args, **kwargs):
-        line, marker = super().create_artists(
+        artists = super().create_artists(
             *args,
             **kwargs
         )
-        line.set_linewidth(self._linewidth)
-        return line, marker
+        artists[0].set_linewidth(self._linewidth)
+        return artists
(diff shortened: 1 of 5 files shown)


gwpy/plot/legend.py
```diff
@@ -45,9 +45,9 @@ def __init__(self, linewidth=6., **kw):
...
     def create_artists(self, *args, **kwargs):
-        line, marker = super().create_artists(
+        artists = super().create_artists(
             *args,
             **kwargs
         )
-        line.set_linewidth(self._linewidth)
-        return line, marker
+        artists[0].set_linewidth(self._linewidth)
+        return artists
```
eagoetz, line 53: @duncanmmacleod I'm not super familiar with the internal workings (obviously) but just wanted to check that the change of this method returning two variables to one (`line, marker` now `artists`) is not going to break anything. I see that tests passed, so presumably this is all good. Just wanted to check 😉

duncanmmacleod: Yes. The internals of this one are strange, but the return from this function is identical to what it was before, just now we're not separating out the two items returned from the matplotlib function (matplotlib now complains loudly if you do that).

eagoetz: Ok, sounds good. Thanks for confirming

eagoetz (approved):
LGTM

PR #1440, opened 2021-11-26
Improvements to the conda requirements parser

duncanmmacleod:
This PR introduces a bunch of improvements to the `parse-conda-requirements.py` script that is used to translate PyPI requirements into conda-forge requirements, these probably boil down to:

- make it more modular
- use grayskull to map pypi names to conda-forge names
- don't write requirements to a tmp file, just add them to the command
- if mamba fails, rerun the thing with conda since mamba doesn't actually return json output on failure, see https://github.com/mamba-org/mamba/issues/1032


hoyc1 (approved):
LGTM (after private discussions on mattermost)

PR #1441, opened 2021-11-26
Drop support for python3.6

duncanmmacleod:
This PR removes support for Python 3.6.

As per NEP29 this version has already been dropped by a number of upstream packages as well as the conda-forge packaging group.


hoyc1 (approved):
LGTM

PR #1442, opened 2021-11-26
Use f-strings for string formatting in gwpy.cli

duncanmmacleod:
This PR updates the `gwpy.cli` module to use [f-strings](https://www.python.org/dev/peps/pep-0498/) for all string formatting, instead of a mix of `str.format` and `%`-formatting.

I think f-strings are a nice, readable way to do most of this formatting, and normally result in shorter statements.


areeda (approved):
LGTM

PR #1445, opened 2021-11-26
Use f-strings for string formatting in gwpy.io

duncanmmacleod:
This PR updates the `gwpy.io` module to use [f-strings](https://www.python.org/dev/peps/pep-0498/) for all string formatting, instead of a mix of `str.format` and `%`-formatting.


areeda (approved):
LGTM

PR #1446, opened 2021-11-29
Use f-strings for string formatting in gwpy.plot

duncanmmacleod:
This PR updates the `gwpy.plot` module to use [f-strings](https://www.python.org/dev/peps/pep-0498/) for all string formatting, instead of a mix of `str.format` and `%`-formatting.

### DIFF
--- a/gwpy/plot/axes.py
+++ b/gwpy/plot/axes.py
@@ -181,8 +181,7 @@ class Axes(_Axes):
                 unit = trans.get_unit_name()
                 iso = Time(epoch, format='gps', scale='utc').iso
                 utc = iso.rstrip('0').rstrip('.')
-                ax.set_label_text('Time [{0!s}] from {1!s} UTC ({2!r})'.format(
-                    unit, utc, epoch))
+                ax.set_label_text(f"Time [{unit}] from {utc} UTC ({epoch!r})")
 
         try:
             super().draw(*args, **kwargs)
@@ -538,7 +537,7 @@ class Axes(_Axes):
                 return ((x-w/2., y-h/2.), (x-w/2., y+h/2.),
                         (x+w/2., y+h/2.), (x+w/2., y-h/2.))
         else:
-            raise ValueError("Unrecognised tile anchor {!r}".format(anchor))
+            raise ValueError(f"Unrecognised tile anchor '{anchor}'")
 
         # build collection
         cmap = kwargs.pop('cmap', rcParams['image.cmap'])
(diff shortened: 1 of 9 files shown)


gwpy/plot/axes.py
```diff
@@ -143,8 +143,7 @@ def draw(self, *args, **kwargs):
                 unit = trans.get_unit_name()
                 iso = Time(epoch, format='gps', scale='utc').iso
                 utc = iso.rstrip('0').rstrip('.')
-                ax.set_label_text('Time [{0!s}] from {1!s} UTC ({2!r})'.format(
-                    unit, utc, epoch))
+                ax.set_label_text(f"Time [{unit}] from {utc} UTC ({epoch!r})")
```
eagoetz, line 146: Checking here: the formatting options `s` are dropped here but not `r`. Meanwhile, on L509, the `r` format option is dropped. Can you confirm this is all fine?

gwpy/plot/gps.py
```diff
@@ -137,13 +138,13 @@ def set_unit(self, unit):
...
         if dec.bases != [units.second]:
-            raise ValueError("Cannot set GPS unit to %s" % unit)
+            raise ValueError(f"cannot set GPS unit to '{unit}'")
         # check equivalent units
         for other in TIME_UNITS:
             if other.decompose().scale == dec.scale:
                 self._unit = other
                 return
-        raise ValueError("Unrecognised unit: %s" % unit)
+        raise ValueError("unrecognised unit '{unit}'")
```
eagoetz, line 147:
```suggestion
        raise ValueError(f"unrecognised unit '{unit}'")
```

gwpy/plot/axes.py
duncanmmacleod: the 's' format is the default for f-strings, so it is unnecessary, while 'r' is required. On L509 the 'r' format was used to enclose the `anchor` value in quotes for the error message, but that behaviour only holds for `str` objects, it's more accurate to just use enclosing quotes manually.

duncanmmacleod:
@eagoetz, any other review comments?

gwpy/plot/axes.py
eagoetz: (sorry for the long-delayed reply) Sounds like you've got this understood and all is good. Resolving.

eagoetz:
@duncanmmacleod No, however, there is a failing test. I'll go ahead and approve it anyway, but thought you should know...

eagoetz (approved):
LGTM

PR #1456, opened 2021-12-09
Add skip decorator to network test

duncanmmacleod:
This PR adds `@pytest_skip_network_error` to another network-dependent test, so that a network error doesn't fail the whole suite.


hoyc1 (approved):
LGTM

PR #1457, opened 2021-12-13
Restructure imports of ilwdchar

duncanmmacleod:
This PR restructures the way that the myriad `ilwdchar` type objects are imported, mainly to fix compatibility with python-ligo-lw 1.8.0; the latest version of that library doesn't include any ilwd stuff at all, so we need import protections.


areeda (approved):
I had to run the old and new code in ipython to understand it.  
So now ILGTM

PR #1458, opened 2021-12-13
Fix import of scipy.signal.windows for scipy 1.8.0

duncanmmacleod:
This PR fixes compatibility with scipy 1.8.0 by updating the usage of `scipy.signal.windows`.

### DIFF
--- a/gwpy/signal/window.py
+++ b/gwpy/signal/window.py
@@ -23,7 +23,10 @@ import numpy
 
 from math import ceil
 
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@gwpy/maintainers the CI on this will continue to fail because #1457 is also needed urgently, but if you review the CI you should see that the scipy-related failures exhibited in the #1457 CI are resolved here (and vice-versa).

areeda (approved):
LGTM

PR #1462, opened 2021-12-20
Simplify reading LIGO_LW documents

duncanmmacleod:
This PR simplifies reading LIGO_LW-format XML documents by replacing the annoying `ilwdchar_compat` system where documents are parsed twice, with a content handler-based system where `ilwdchar` types are swapped with the relevant `int`-like types on-the-fly automatically as documents are being parsed.

This allows to simplify everything else about reading LIGO_LW documents.


areeda (approved):
LGTM
I just wonder why we did this to ourselves to avoid a string to int conversion

PR #1463, opened 2022-01-02
Update minimum astropy requirement to 4.0

duncanmmacleod:
This PR bumps the minimum requirement on Astropy to `>= 4.0`, principally to resolve warnings from erfa relating to leap seconds for 2022. This also updates numpy since astropy will require at least 1.16 anyway.


eagoetz (approved):
LGTM

PR #1466, opened 2022-01-12
Remove gwpy.table.EventColumn

duncanmmacleod:
This PR removes the `EventColumn` object which has been deprecated since v1.0.0, see 812a7127deee8cff0433a9ee934e3de6a44c45c7.


gwpy/table/table.py
```diff
@@ -28,15 +29,14 @@
...
-from astropy.table import (Table, Column, vstack)
+from astropy.table import (Table, vstack)
 from astropy.io import registry
 
 from ..io.mp import read_multi as io_read_multi
 from ..time import gps_types
 from .filter import (filter_table, parse_operator)
 
 __author__ = 'Duncan Macleod <duncan.macleod@ligo.org>'
-__all__ = ['EventColumn', 'EventTable']
```
robertbruntz, line 39: Should this entire line be deleted or just `'EventColumn', `?  AFAIK, this will only [affect `import *` statements](https://stackoverflow.com/questions/44834/can-someone-explain-all-in-python), but I wanted to make sure the change is intentional.

duncanmmacleod: Yes, I intentionally removed the `__all__`, it doesn't really help for a module with such a small set of defined objects.

robertbruntz (approved):
LGTM

PR #1468, opened 2022-01-12
Remove deprecated colorbar behaviour

duncanmmacleod:
This PR finalises two deprecations of the colorbar functionality:

- remove the `log` keyword that has been deprecated for years
- implement the change to the default `fraction`, now is 0.0 which means 'don't shrink the parent axes'

### DIFF
--- a/gwpy/plot/colorbar.py
+++ b/gwpy/plot/colorbar.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) Duncan Macleod (2017-2020)
+# Copyright (C) Louisiana State University (2017)
+#               Cardiff University (2017-2022)
 #
 # This file is part of GWpy.
 #
@@ -18,8 +19,6 @@
 
 """Utilities for generating colour bars for figures
 """
-
-import warnings
 
 from matplotlib.axes import SubplotBase
 from matplotlib.colors import LogNorm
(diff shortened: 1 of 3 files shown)


gwpy/plot/colorbar.py
```diff
@@ -69,24 +59,8 @@ def process_colorbar_kwargs(figure, mappable=None, ax=None, use_axesgrid=True,
...
-                          "behaviour, however, this will change to `0.0` in "
-                          "an upcoming release. To keep the current default, "
-                          "manually specify `fraction=0.15`",
-                          PendingDeprecationWarning)
-            kwargs['fraction'] = 0.15
-
-        # create axes if requesting not to resize the parent
-        if use_axesgrid:
-            cax, kwargs = make_axes_axesgrid(ax, **kwargs)
-    else:
```
robertbruntz, line 88: @duncanmmacleod Is there a reason that this scenario (cax is not None) is no longer handled in the modified code?

gwpy/plot/colorbar.py
```diff
@@ -69,24 +59,8 @@ def process_colorbar_kwargs(figure, mappable=None, ax=None, use_axesgrid=True,
...
-        kwargs.pop('fraction', None)  # don't need this if already have axes
+    if cax is None and use_axesgrid:
+        cax, kwargs = make_axes_axesgrid(ax, **kwargs)
```
robertbruntz, line 63: @duncanmmacleod Should the scenario of 'cax is None' and 'use_axesgrid' is False be handled in the code?  (I haven't been able to infer enough about what `cax` and `use_axesgrid` do to determine this myself.)

duncanmmacleod: Explicitly no. That condition falls back to the default `matplotlib` behaviour. I will update the documentation to better explain this.

robertbruntz:
LGTM

robertbruntz (approved):
LGTM

PR #1471, opened 2022-01-12
Remove deprecated GWF I/O options

duncanmmacleod:
This PR removes the `resample` and `dtype` options for reading GWF files, these have been deprecated since v0.11.0,
see 84672af3c4d3c285b758736eda4487fbe40fa1a0.

### DIFF
--- a/docs/timeseries/io.rst
+++ b/docs/timeseries/io.rst
@@ -182,15 +182,6 @@ To read multiple channels from one or more GWF files (rather than opening and cl
 
    >>> data = TimeSeriesDict.read('HLV-HW100916-968654552-1.gwf', ['H1:LDAS-STRAIN', 'L1:LDAS-STRAIN'])
 
-In this case the ``resample`` and ``dtype`` keywords can be given as a single value used for all data channels, or a `dict` mapping an argument for each data channel name:
-
-.. code-block:: python
-
-   >>> data = TimeSeriesDict.read('HLV-HW100916-968654552-1.gwf', ['H1:LDAS-STRAIN', 'L1:LDAS-STRAIN'],
-   ...                            resample={'H1:LDAS-STRAIN': 2048})
-
-The above example will resample only the ``'H1:LDAS-STRAIN'`` `TimeSeries` and will not modify that for ``'L1:LDAS-STRAIN'``.
-
 .. note::
 
    A mix of `TimeSeries` and `StateVector` objects can be read by using only `TimeSeriesDict` class, and casting the returned data to a `StateVector` using :meth:`~TimeSeries.view`.
--- a/gwpy/timeseries/core.py
+++ b/gwpy/timeseries/core.py
@@ -368,10 +368,10 @@ class TimeSeriesBase(Series):
             prefix for the progress meter
 
         type : `int`, optional
-            NDS2 channel type integer
+            NDS2 channel type integer or string name to match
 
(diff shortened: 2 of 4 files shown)


robertbruntz:
@duncanmmacleod It looks to me like these are the same `resample` and `dtype` as are mentioned in `docs/timeseries/io.rst`, correct?  If so, do you want to excise them there and rebase?

Also, to be clear, the `resample` option for reading GWF files is being removed, but the `resample` method is being kept, i.e., in `gwpy/timeseries/core.py`, `gwpy/timeseries/statevector.py`, `gwpy/timeseries/timeseries.py`, `gwpy/spectrogram/coherence.py` and testing in `gwpy/timeseries/tests/{test_statevector.py,test_core.py,test_timeseries.py}`, correct?

robertbruntz:
@duncanmmacleod Am I correct that the `fetch` method gets data from NDS, so `dtype` is retained, whereas the `find` method gets data from frame files, so `dtype` is being removed?  If so, what about the `get` method starting on line 549?  It can get data from either frames or NDS.  Is that handled by the `get` method starting on line 1284, which seems to know to retain the `dtype` keyword for NDS queries but not for frame file queries?  (This is getting out of my depth at this level, so I'll take your word for it.)

robertbruntz:
@duncanmmacleod ping on the above questions

duncanmmacleod:
> @duncanmmacleod It looks to me like these are the same `resample` and `dtype` as are mentioned in `docs/timeseries/io.rst`, correct? If so, do you want to excise them there and rebase?

Good catch, those have been removed in 15ddb2d.
 
> Also, to be clear, the `resample` option for reading GWF files is being removed, but the `resample` method is being kept, i.e., in `gwpy/timeseries/core.py`, `gwpy/timeseries/statevector.py`, `gwpy/timeseries/timeseries.py`, `gwpy/spectrogram/coherence.py` and testing in `gwpy/timeseries/tests/{test_statevector.py,test_core.py,test_timeseries.py}`, correct?

Yes, that is correct. Rather than fancy on-the-fly resampling, users should just call the `.resample()` method manually.

duncanmmacleod:
> @duncanmmacleod Am I correct that the `fetch` method gets data from NDS, so `dtype` is retained, whereas the `find` method gets data from frame files, so `dtype` is being removed? If so, what about the `get` method starting on line 549? It can get data from either frames or NDS. Is that handled by the `get` method starting on line 1284, which seems to know to retain the `dtype` keyword for NDS queries but not for frame file queries? (This is getting out of my depth at this level, so I'll take your word for it.)

So, this is a little complicated. For `TimeSeries.fetch` (NDS2) the `dtype` argument is used to identify the right channels, but searching only for channels explicitly with the specified data type, see:

https://github.com/gwpy/gwpy/blob/7a27a9c0b4552934f328c58dca7645402fa431fd/gwpy/io/nds2.py#L514

That argument is not (and has never been) used to actually _change_ the data type, 4aba30d clarifies this in the docstrings for the relevant methods.

So now passing `dtype` to `TimeSeries{Dict}.get` is only used in the context of identifying the correct data source from NDS2, see

https://github.com/gwpy/gwpy/blob/15ddb2d9c178f7617aa054e16ac5abfccd023682/gwpy/timeseries/core.py#L1322-L1327

robertbruntz (approved):
LGTM

PR #1477, opened 2022-01-21
Remove equivalence between custom strain unit and 'dimensionless_unscaled'

duncanmmacleod:
This PR improves the handling of the 'strain' unit by removing the equivalence with astropy's dimensionless_unscaled. This means that it won't be silently dropped during unit operations.

Before:

```python
>>> import gwpy.detector  # register the 'strain' unit
>>> from astropy import units
>>> units.Unit('strain') * units.Unit('second')
Unit("s")
```

After:

```python
>>> import gwpy.detector
>>> from astropy import units
>>> units.Unit('strain') * units.Unit('second')
Unit("s strain")
```

### DIFF
--- a/gwpy/detector/units.py
+++ b/gwpy/detector/units.py
@@ -177,7 +177,7 @@ units.def_unit(['NONE', 'undef'], namespace=_ns,
                doc='No unit has been defined for these data')
 
 # other dimenionless units
-units.def_unit('strain', represents=units.Unit(''), namespace=_ns)
+units.def_unit('strain', namespace=_ns)
 units.def_unit('coherence', namespace=_ns)
 
 # alias for 'second' but with prefices


eagoetz (commented):
Strain is a dimensionless quantity, so the purpose of this is to include "strain" within any operation involving units or conversions. Wouldn't that be confusing to people? Or have you received feedback that "strain" should be carried as a unit?

duncanmmacleod:
> Strain is a dimensionless quantity, so the purpose of this is to include "strain" within any operation involving units or conversions. Wouldn't that be confusing to people? Or have you received feedback that "strain" should be carried as a unit?

This change was adapted from @jolien-creighton's prototype for a new waveform interface. The functional purpose here is to have the GWpy unit handling match that of LAL, where `strain` is carried in unit conversions, as in the example I posted at the top. However, I'm not totally sold on this myself.

jolien-creighton:
On Thu, Feb 3, 2022 at 2:54 PM Duncan Macleod ***@***.***>
wrote:

> Strain is a dimensionless quantity, so the purpose of this is to include
> "strain" within any operation involving units or conversions. Wouldn't that
> be confusing to people? Or have you received feedback that "strain" should
> be carried as a unit?
>
> This change was adapted from @jolien-creighton
> <https://github.com/jolien-creighton>'s prototype for a new waveform
> interface. The functional purpose here is to have the GWpy unit handling
> match that of LAL, where strain is carried in unit conversions, as in the
> example I posted at the top. However, I'm not totally sold on this myself.
>
>
All units are fabrications anyway (particularly now that so many
fundamental physical constants are exact).  In LAL there are some
inequivalent "dimensionless" units used for making sure that the wrong time
series isn't accidentally passed to a function.  For example, the response
function R(f) has units of "strain / count" (both dimensionless) so that
h(f) = R(f) e(f) has units of "strain s" where e(f) has units of "count s"
and the routine that produces h(t) takes in one time series and an optional
response function and makes sure that the product of those two will have
units of "strain".

(As a side effect, in GWpy I think the unit name "strain" should appear on
plots of strain vs. time, whereas it doesn't now because it gets converted
to dimensionless which has no name.)

duncanmmacleod:
Thanks @jolien-creighton, that is very helpful context.

eagoetz:
@jolien-creighton @duncanmmacleod Ok, I don't want to be the one to hold this up. So I'll approve it, but I do have reservations about calling strain a unit

eagoetz (approved):
Perhaps not my preference, but for now LGTM

PR #1479, opened 2022-01-21
Fix bugs in LAL unit conversion

### LINKED ISSUE #1478: Incorrect conversion from LAL units with fractional power
@jolien-creighton reports a bug in converting from LAL units with fractional power:

```python
>>> import lal
>>> from gwpy.utils.lal import from_lal_unit
>>> from_lal_unit(lal.Unit("m^1/2"))
Unit("m")
```

duncanmmacleod:
This PR fixes #1478 by implementing changes proposed by @jolien-creighton to improve the LAL to/from unit conversions.
This has required an API change in the `gwpy.utils.lal.to_lal_unit` function, but I believe this to be sparingly used out of GWpy, so shouldn't be heavily disruptive.

### DIFF
--- a/gwpy/utils/lal.py
+++ b/gwpy/utils/lal.py
@@ -159,13 +159,7 @@ LAL_UNIT_INDEX = [  # the order corresponds to how LAL stores compound units
 
 
 def to_lal_unit(aunit):
-    """Convert the input unit into a `LALUnit`
-
-    For example::
-
-       >>> u = to_lal_unit('m**2 / kg ** 4')
-       >>> print(u)
-       m^2 kg^-4
+    """Convert the input unit into a `lal.Unit` and a scaling factor.
 
     Parameters
     ----------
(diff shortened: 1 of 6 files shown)


gwpy/utils/lal.py
```diff
@@ -190,25 +182,36 @@ def to_lal_unit(aunit):
     ValueError
         if LAL doesn't understand the base units for the input
     """
+    # format incoming unit
     if isinstance(aunit, str):
         aunit = units.Unit(aunit)
     aunit = aunit.decompose()
-    lunit = lal.Unit()
+
+    # handle scaled units
```
eagoetz, line 190: Minor comment: it would be helpful to explain what "handle scaled units" means in this context. Handle in what way?

duncanmmacleod: This is now explained in the docstring for the function, see 1ee6961b134323ebbd5fc122ee96dbaf9b6c34b4.

eagoetz (approved):
Thanks! LGTM

PR #1480, opened 2022-02-01
Use pip2conda to parse requirements for conda

duncanmmacleod:
The `parse-conda-requirements.py` script has been migrated out of GWpy and into its own `pip2conda` project, this PR updates this project to use that one.

Currently this is installing directly from the git repo, but will be updated to install from conda-forge when `pip2conda` has been migrated to that platform.


hoyc1 (approved):
Looks good to me!

PR #1483, opened 2022-02-03
Update Axes.tile to use sortbycolor

duncanmmacleod:
This PR updates `Axes.tile()` to recognise the `sortbycolor=False` keyword, to bring it inline with `Axes.scatter`.
I also added deprecated support for the old `c_sort` variable to assist downstream users with legacy code.

cc @eagoetz.

### DIFF
--- a/gwpy/plot/axes.py
+++ b/gwpy/plot/axes.py
@@ -117,6 +117,44 @@ def restore_grid(func):
     return wrapped_func
 
 
+def deprecate_c_sort(func):
+    """Wrap ``func`` to replace the deprecated ``c_sort`` keyword.
+
+    This was renamed ``sortbycolor``.
+    """
+    @wraps(func)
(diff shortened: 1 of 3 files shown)


eagoetz:
@duncanmmacleod I think this is all ok, but I am not that familiar with how wrapping works. It looks like the tests try using both `sortbycolor` and `c_sort`, so I think both cases are covered. Is there a particular reason to deprecate `c_sort`; or is it just that the keyword is renamed to `sortbycolor`?

duncanmmacleod:
>... or is it just that the keyword is renamed to `sortbycolor`?

Exactly, the new name is more descriptive.

eagoetz (approved):
I'm not familiar at all with how python wrapping of functions works, but I think this looks ok and the unit tests Duncan implemented test different conditionals of using `c_sort` and `sortbycolor`.

So I'll give it a LGTM

PR #1488, opened 2022-03-17
Avoid regex 2022.3.15 in requirements

duncanmmacleod:
This MR patches `setup.cfg` to avoid `regex` 2022.3.15, see https://github.com/scrapinghub/dateparser/issues/1045.


eagoetz (approved):
LGTM

PR #1490, opened 2022-03-18
Allow irregular non-contiguous time series to be appended

mattpitkin:
In a follow-up to https://github.com/gwpy/gwpy/pull/1428, this PR edits the `Series.append()` method to allow irregular non-consecutive time series to be appended/joined. For example, currently the following will fail:

```python
from gwpy.timeseries import TimeSeries, TimeSeriesList
ts1 = TimeSeries([1, 2, 3, 4], times=[0, 4, 5, 7])
ts2 = TimeSeries([5, 6, 7, 8, 9], times=[10, 11, 12, 15, 16])
l = TimeSeriesList()
l.append(ts1)
l.append(ts2)
newd = l.join(gap="ignore")
```

With this PR it will work.

### DIFF
--- a/gwpy/types/series.py
+++ b/gwpy/types/series.py
@@ -771,7 +771,11 @@ class Series(Array):
 
         # check metadata (do not do this if trying to join to unevenly sampled
         # discontiguous series')
-        if not (gap == 'ignore' and (not hasattr(self, 'dx') or not hasattr(other, 'dx'))):
+        if not (
+            gap == 'ignore' and (
+                not hasattr(self, 'dx') or not hasattr(other, 'dx')
+            )
+        ):
             self.is_compatible(other)
 
         # make copy if needed
--- a/gwpy/types/series.py
+++ b/gwpy/types/series.py
@@ -769,37 +769,39 @@ class Series(Array):
         if pad is None and gap == 'pad':
             pad = 0.
 
-        # check metadata
-        self.is_compatible(other)
+        # check metadata (do not do this if trying to join to unevenly sampled
+        # discontiguous series')
+        if not (gap == 'ignore' and (not hasattr(self, 'dx') or not hasattr(other, 'dx'))):
+            self.is_compatible(other)
+
         # make copy if needed
         if not inplace:
             self = self.copy()
         # fill gap
-        if self.is_contiguous(other) != 1:
-            if gap == 'pad':
-                ngap = floor(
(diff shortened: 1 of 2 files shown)


gwpy/types/series.py
```diff
@@ -769,37 +769,43 @@ def append(self, other, inplace=True, pad=None, gap=None, resize=True):
...
 
-        # check metadata
-        self.is_compatible(other)
+        # check metadata (do not do this if trying to join to unevenly sampled
+        # discontiguous series')
+        if not (
+            gap == 'ignore' and (
+                not hasattr(self, 'dx') or not hasattr(other, 'dx')
+            )
+        ):
```
duncanmmacleod, line 778: @mattpitkin, is this a simpler version of what you are suggesting:

```suggestion
        if (
            gap != "ignore"
            or (hasattr(self, "dx") and hasattr(other, "dx"))
        ):
```

gwpy/types/series.py
```diff
@@ -845,8 +851,22 @@ def append(self, other, inplace=True, pad=None, gap=None, resize=True):
                     self.xindex.resize((s[0],), refcheck=False)
                 except ValueError as exc:
                     if 'cannot resize' in str(exc):
-                        self.xindex = self.xindex.copy()
-                        self.xindex.resize((s[0],))
+                        try:
+                            self.xindex = self.xindex.copy()
+                            self.xindex.resize((s[0],))
+                        except ValueError as exc2:
```
duncanmmacleod, line 857: Does the `.copy()` operation need to be inside the try/except?

```suggestion
                        self.xindex = self.xindex.copy()
                        try:
                            self.xindex.resize((s[0],))
                        except ValueError as exc2:
```

duncanmmacleod (commented):
@mattpitkin, thanks for this proposal, and apologies it's taken me _so_ long to look at it. I am concerned about the complexity of the `append()` method after this change, I might recommend refactoring the whole thing to handle the different conditions in separate internal methods.

mattpitkin:
@duncanmmacleod I'd originally thought I'd need this patch for cwinpy, but it turns out I found another workaround. I'd be happy for this PR to be closed and a more general refactor be made instead.

duncanmmacleod:
> @duncanmmacleod I'd originally thought I'd need this patch for cwinpy, but it turns out I found another workaround. I'd be happy for this PR to be closed and a more general refactor be made instead.

Ok, thanks for the update; let's close this and see how we get on (I don't have time to take on any refactoring right now).

PR #1491, opened 2022-03-21
Simplify parsing LTI objects

duncanmmacleod:
This PR simplifies the handling of `scipy.signal.LinearTimeInvariant` objects in `gwpy.signal.filter_design`; the change in #1484 doesn't actually work with scipy-1.8.0.


areeda (approved):
LGTM

PR #1494, opened 2022-05-04
Split up functionality of TimeSeries.gate()

dethodav:
This PR splits `TimeSeries.gate()` into two functions:
* `TimeSeries.find_gates()` - find peaks in whitened timeseries and returns the segments to gate
* `TimeSeries.gate()` - now only calls `TimeSeries.find_gates()` and `TimeSeries.mask()` to find and apply gates

The motivation for this change is that it would be helpful to be able to record the segments that were gated. Users hoping to do this can now use `TimeSeries.find_gates()` and `TimeSeries.mask()` separately. 

Since this PR is only reorganizing code, the result of `TimeSeries.gate()` remains unchanged.

### DIFF
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1806,7 +1806,7 @@ class TimeSeries(TimeSeriesBase):
         return out * numpy.sqrt(2 * in_.dt.decompose().value)
 
     def find_gates(self, tzero=1.0, whiten=True,
-             threshold=50., cluster_window=0.5, **whiten_kwargs):
+                   threshold=50., cluster_window=0.5, **whiten_kwargs):
         """Identify points that should be gates using a provided threshold
         and clustered within a provided time window.
 
@@ -1929,9 +1929,10 @@ class TimeSeries(TimeSeriesBase):
         TimeSeries.whiten
             for the whitening filter used to identify gating points
         """
-        deadtime = self.find_gates(tzero=tzero, 
-             whiten=whiten, threshold=threshold, 
-             cluster_window=cluster_window, **whiten_kwargs)
+        deadtime = self.find_gates(tzero=tzero, whiten=whiten,
+                                   threshold=threshold,
+                                   cluster_window=cluster_window,
+                                   **whiten_kwargs)
         return self.mask(deadtime=deadtime, const=0, tpad=tpad)
 
     def convolve(self, fir, window='hanning'):
--- a/gwpy/timeseries/timeseries.py
+++ b/gwpy/timeseries/timeseries.py
@@ -1933,7 +1933,7 @@ class TimeSeries(TimeSeriesBase):
         TimeSeries.whiten
             for the whitening filter used to identify gating points
         """
-        deadtime = self.find_gates(self, tzero=tzero, 
+        deadtime = self.find_gates(tzero=tzero, 
              whiten=whiten, threshold=threshold, 
              cluster_window=cluster_window, **whiten_kwargs)
         return self.mask(deadtime=deadtime, const=0, tpad=tpad)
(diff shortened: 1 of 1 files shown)


gwpy/timeseries/timeseries.py
```diff
@@ -1805,6 +1805,57 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+            other keyword arguments that will be passed to the
+            `TimeSeries.whiten` method if it is being used when discovering
+            gating points
+
+        Returns
+        -------
+        out : `~gwpy.segments.SegmentList`
+            a list of segments that should be gated based on the
+            provided parameters
+        See also
```
duncanmmacleod, line 1840: Need a blank line here

```suggestion

        See also
```

gwpy/timeseries/timeseries.py
```diff
@@ -1805,6 +1805,57 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+        --------
+        TimeSeries.gate
+            for a method that applies the identified gates
+        """
+        from scipy.signal import find_peaks
```
duncanmmacleod, line 1845: `scipy.signal` is already imported at the module level, so this is not required

```suggestion
```

(I realise that this line was already here, but let's fix it anyway)

gwpy/timeseries/timeseries.py
```diff
@@ -1805,6 +1805,57 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+        # Find points to gate based on a threshold
+        sample = self.sample_rate.to('Hz').value
+        data = self.whiten(**whiten_kwargs) if whiten else self
+        window_samples = cluster_window * sample
+        gates = find_peaks(abs(data.value), height=threshold,
```
duncanmmacleod, line 1850: As above

```suggestion
        gates = signal.find_peaks(abs(data.value), height=threshold,
```

gwpy/timeseries/timeseries.py
```diff
@@ -1805,6 +1805,57 @@ def whiten(self, fftlength=None, overlap=0, method=DEFAULT_FFT_METHOD,
...
+                           distance=window_samples)[0]
+        # represent gates as time segments
+        deadtime = SegmentList([Segment(
+            self.t0.value + (k / sample) - tzero,
+            self.t0.value + (k / sample) + tzero,
+        ) for k in gates]).coalesce()
+        return deadtime
```
duncanmmacleod, line 1857:
```suggestion
        return SegmentList([Segment(
            self.t0.value + (k / sample) - tzero,
            self.t0.value + (k / sample) + tzero,
        ) for k in gates]).coalesce()
```

gwpy/timeseries/timeseries.py
```diff
@@ -1873,22 +1924,15 @@ def gate(self, tzero=1.0, tpad=0.5, whiten=True,
...
-        deadtime = SegmentList([Segment(
-            self.t0.value + (k / sample) - tzero,
-            self.t0.value + (k / sample) + tzero,
-        ) for k in gates]).coalesce()
-        # return the self-gated timeseries
+        deadtime = self.find_gates(tzero=tzero, whiten=whiten,
+                                   threshold=threshold,
+                                   cluster_window=cluster_window,
+                                   **whiten_kwargs)
         return self.mask(deadtime=deadtime, const=0, tpad=tpad)
```
duncanmmacleod, line 1936: The name `gates` I think would be clearer here.

```suggestion
        gates = self.find_gates(tzero=tzero, whiten=whiten,
                                   threshold=threshold,
                                   cluster_window=cluster_window,
                                   **whiten_kwargs)
        return self.mask(deadtime=gates, const=0, tpad=tpad)
```

duncanmmacleod (changes requested):
Thanks @dethodav, this will be a nice addition, I have a few pedantic requests.

dethodav:
@duncanmmacleod All your suggestions should now be implemented

dethodav:
While making a "main" branch on my fork, I seem to have inadvertently closed this PR. I've now reopened it.

duncanmmacleod (approved):
LGTM, thanks @dethodav.

PR #1500, opened 2022-05-06
Add ValueError when improper starting frequency used with `to_pycbc()`

dethodav:
This PR ensures that a `ValueError` is returned when the user attempts to use `frequencyseries.FrequencySeries.to_pycbc()` in cases when the starting frequency is not `0 Hz`. When a PyCBC frequency series is created with `f0 != 0 Hz`, the PyCBC frequencies are not correct (They are offset by `f0`).

Here is an example of code that now returns a `ValueError` since the starting frequency is `10 Hz`:
```
from gwpy.frequencyseries import FrequencySeries
psd = (FrequencySeries.read('O3-H1-C01_CLEAN_SUB60HZ-1262197260.0_sensitivity_strain_asd.txt')).to_pycbc()
print(psd.sample_frequencies)
```

The above example uses this file: https://dcc.ligo.org/public/0174/G2100674/001/O3-H1-C01_CLEAN_SUB60HZ-1262197260.0_sensitivity_strain_asd.txt

### DIFF
--- a/gwpy/frequencyseries/frequencyseries.py
+++ b/gwpy/frequencyseries/frequencyseries.py
@@ -437,7 +437,7 @@ class FrequencySeries(Series):
             epoch = self.epoch.gps
         if self.f0.to('Hz').value:
             raise ValueError(
-                "Cannot convert FrequencySeries to PyCBC with f0 = {self.f0}. "
+                f"Cannot convert FrequencySeries to PyCBC with f0 = {self.f0}. "
                 "Starting frequency must be equal to 0 Hz."
             )
         return types.FrequencySeries(self.value,
--- a/gwpy/frequencyseries/frequencyseries.py
+++ b/gwpy/frequencyseries/frequencyseries.py
@@ -437,8 +437,8 @@ class FrequencySeries(Series):
             epoch = self.epoch.gps
         if self.f0.to('Hz').value:
             raise ValueError(
-                f"Cannot convert FrequencySeries to PyCBC with f0 = {self.f0}. "
-                "Starting frequency must be equal to 0 Hz."
+                f"Cannot convert FrequencySeries to PyCBC with f0 = {self.f0}."
+                " Starting frequency must be equal to 0 Hz."
             )
         return types.FrequencySeries(self.value,
                                      delta_f=self.df.to('Hz').value,
--- a/gwpy/frequencyseries/frequencyseries.py
+++ b/gwpy/frequencyseries/frequencyseries.py
@@ -435,6 +435,10 @@ class FrequencySeries(Series):
             epoch = None
         else:
             epoch = self.epoch.gps
+        if self.f0.to('Hz').value > 0.0:
+            raise ValueError("Cannot convert FrequencySeries to PyCBC with "
+                             "f0 = {0}. Starting frequency must be equal "
+                             "to 0 Hz.".format(self.f0))
         return types.FrequencySeries(self.value,
                                      delta_f=self.df.to('Hz').value,
                                      epoch=epoch, copy=copy)
--- a/gwpy/frequencyseries/frequencyseries.py
+++ b/gwpy/frequencyseries/frequencyseries.py
@@ -435,10 +435,11 @@ class FrequencySeries(Series):
             epoch = None
         else:
             epoch = self.epoch.gps
-        if self.f0.to('Hz').value > 0.0:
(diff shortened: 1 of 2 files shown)


gwpy/frequencyseries/frequencyseries.py
```diff
@@ -435,6 +435,10 @@ def to_pycbc(self, copy=True):
             epoch = None
         else:
             epoch = self.epoch.gps
+        if self.f0.to('Hz').value > 0.0:
```
duncanmmacleod, line 438: Since the requirement is that the `f0` be identically zero, we should just check the truthiness of the value:

```suggestion
        if self.f0.to('Hz').value:
```

gwpy/frequencyseries/frequencyseries.py
```diff
@@ -435,6 +435,10 @@ def to_pycbc(self, copy=True):
...
+            raise ValueError("Cannot convert FrequencySeries to PyCBC with "
+                             "f0 = {0}. Starting frequency must be equal "
+                             "to 0 Hz.".format(self.f0))
```
duncanmmacleod, line 441: It might be cleaner here to use f-strings:

```suggestion
            raise ValueError(
                "Cannot convert FrequencySeries to PyCBC with f0 = {self.f0}. "
                "Starting frequency must be equal to 0 Hz."
            )
```

duncanmmacleod (changes requested):
This looks good, thanks @dethodav, just a few pedantic changes (as per usual). I might also move the `f0` check above the `if self.epoch is None` clause, so that isn't executed unnecessarily.

gwpy/frequencyseries/frequencyseries.py
```diff
@@ -435,6 +435,11 @@ def to_pycbc(self, copy=True):
             epoch = None
         else:
             epoch = self.epoch.gps
+        if self.f0.to('Hz').value:
+            raise ValueError(
+                "Cannot convert FrequencySeries to PyCBC with f0 = {self.f0}. "
```
duncanmmacleod, line 440: The f-string prefix character is missing here:

```suggestion
                f"Cannot convert FrequencySeries to PyCBC with f0 = {self.f0}. "
```

duncanmmacleod (changes requested):
Thanks @dethodav, a single character is missing...

duncanmmacleod:
@dethodav, sorry that I should have suggested this earlier, but can you add a test in `gwpy/frequencyseries/tests/test_frequencyseries.py` that asserts that the new error is actually raised appropriately?

The coverage test seems to think that it's already covered, but it's unclear to me where that is.

dethodav:
@duncanmmacleod Just added in a test - let me know if this should be structured differently

gwpy/frequencyseries/tests/test_frequencyseries.py
```diff
@@ -228,6 +228,12 @@ def test_to_from_pycbc(self, array):
         a2 = type(array).from_pycbc(array.to_pycbc(copy=False), copy=False)
         assert shares_memory(array.value, a2.value)
 
+        # test conversion when f0 != 0
```
duncanmmacleod, line 231: It would be nice to separate this into its own test function, so that it is independently asserted (i.e. so that the error handling is checked even when the `test_to_from_pycbc` function fails early.

```suggestion
    @pytest.mark.requires("pycbc")
    def test_to_from_pycbc_nonzero_f0(self, array):
        """Test `FrequencySeries.to_pycbc` conversion when ``f0 != 0``.
        """
```

gwpy/frequencyseries/tests/test_frequencyseries.py
```diff
@@ -228,6 +228,12 @@ def test_to_from_pycbc(self, array):
...
+        array.f0 = 1.
+        with pytest.raises(ValueError) as exc:
+            array.to_pycbc()
+        assert str(exc.value).startswith("Cannot convert FrequencySeries")
```
duncanmmacleod, line 235: I think the message matching can be handled directly by `pytest.raises`:

```suggestion
        with pytest.raises(
            ValueError,
            match="^Cannot convert FrequencySeries",
        ):
            array.to_pycbc()
```

(this is not followed universally in other locations in GWpy)

duncanmmacleod (changes requested):
Thanks @dethodav, two pedantic requests relating to the new (excellent) test method.

duncanmmacleod (approved):
LGTM, thanks @dethodav!

PR #1501, opened 2022-05-12
LALFrame handling of empty arrays.

pchanial:
In case no frames match a time segment, it would be handy if reading by LALframe of a gwf file could return an empty ndarray instead of None for the data.data attribute of the LAL timeseries.

Here is a first attempt to do so.
Currently the writing of an empty LAL series fails when it is added (add_ cython or c function) to the frame in gwpy.timeseries.io.gwf.lalframe.write:
```python
        # add time series to frame
        add_(frame, lalseries)
```
So I've marked the test as xfailed, I would need more pointers to fix it (where is the relevant code ?)

### DIFF
--- a/gwpy/timeseries/io/gwf/lalframe.py
+++ b/gwpy/timeseries/io/gwf/lalframe.py
@@ -153,9 +153,8 @@ def read(source, channels, start=None, end=None, series_class=TimeSeries,
     # read data
     out = series_class.DictClass()
     for name in channels:
-        out[name] = series_class.from_lal(
-            _read_channel(stream, str(name), start=start, duration=duration),
-            copy=False)
+        ts = _read_channel(stream, str(name), start=start, duration=duration)
+        out[name] = series_class.from_lal(ts, copy=False)
         lalframe.FrStreamSeek(stream, epoch)
     return out
 
--- a/gwpy/timeseries/tests/test_io_gwf_lalframe.py
+++ b/gwpy/timeseries/tests/test_io_gwf_lalframe.py
@@ -23,6 +23,7 @@ from pathlib import Path
 from urllib.parse import urlparse
 
 import pytest
+import numpy
 
 from gwdatafind.utils import file_segment
 
@@ -43,8 +44,9 @@ gwpy_lalframe = pytest.importorskip("gwpy.timeseries.io.gwf.lalframe")
 TEST_GWF_PATH = Path(TEST_GWF_FILE).absolute()
 TEST_GWF_URL = TEST_GWF_PATH.as_uri()
 
-# get epoch corresponding to this file
+# get info corresponding to this file
 TEST_GWF_SEGMENT = file_segment(TEST_GWF_FILE)
+TEST_GWF_DELTA_T = 1 / 16384
 
 # channels to read
 CHANNELS = ["H1:LDAS-STRAIN", "L1:LDAS-STRAIN", "V1:h_16384Hz"]
@@ -103,7 +105,8 @@ def test_get_stream_duration(stream):
     (None, None),
     (None, TEST_GWF_SEGMENT[0] + .5),
     (TEST_GWF_SEGMENT[0] + .5, None),
-    (TEST_GWF_SEGMENT[0] + .25, TEST_GWF_SEGMENT[1] - .25)
+    (TEST_GWF_SEGMENT[0] + .25, TEST_GWF_SEGMENT[1] - .25),
+    (TEST_GWF_SEGMENT[1] - TEST_GWF_DELTA_T / 2, None),
 ])
 def test_read(start, end):
     data = gwpy_lalframe.read(
@@ -124,7 +127,7 @@ def test_read(start, end):
         assert ts.name == name
 
         # check data span is what we asked for
-        assert ts.xspan == (start, end)
+        assert numpy.allclose(ts.xspan, (start, end), atol=TEST_GWF_DELTA_T)
 
 
 def test_read_channel_error():
@@ -158,9 +161,22 @@ def test_write(tmp_path):
     assert_dict_equal(data, data2, assert_quantity_sub_equal)
 
 
-def test_write_no_ifo(tmp_path):
+@pytest.mark.parametrize(
+    'data',
+    [
+        pytest.param(
+            [],
+            marks=pytest.mark.xfail(
+                raises=RuntimeError,
+                reason='Cannot add an empty series to a frame',
+            )
+        ),
+        [1, 2, 3, 4, 5],
+    ]
+)
+def test_write_no_ifo(tmp_path, data):
     # create timeseries with no IFO
-    data = TimeSeries([1, 2, 3, 4, 5])
+    data = TimeSeries(data, dtype=float)
     tmp = tmp_path / "test.gwf"
     gwpy_lalframe.write(
         {None: data},
--- a/gwpy/utils/lal.py
+++ b/gwpy/utils/lal.py
@@ -22,6 +22,7 @@ This module requires lal >= 6.14.0
 """
 
 import operator
+import re
 from collections import OrderedDict
 from fractions import Fraction
 from functools import reduce
@@ -71,6 +72,9 @@ LAL_TYPE_FROM_NUMPY = {
 
 LAL_TYPE_STR_FROM_NUMPY = {k: LAL_TYPE_STR[v] for
                            (k, v) in LAL_TYPE_FROM_NUMPY.items()}
+LAL_NUMPY_FROM_TYPE_STR = {v: k for k, v in LAL_TYPE_STR_FROM_NUMPY.items()}
+
+LAL_TYPE_REGEX = re.compile(r'(U?INT|REAL|COMPLEX)\d+')
 
 
 def to_lal_type_str(pytype):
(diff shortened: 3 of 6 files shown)


duncanmmacleod:
@pchanial, I'm not entirely sure what problem you are trying to solve. Can you post a minimal reproducer that demonstrates a current failure?

pchanial:
Steps to reproduce:

```from gwpy.timeseries import TimeSeries
ts = TimeSeries.read('H-H1_llhoft-1336320993-1.gwf', channel='H1:GDS-CALIB_STATE_VECTOR', start=1336320993.999024, end=1336320994)
```
Result:
```python
ile ~/work/repos/gwpy/gwpy/timeseries/core.py:702, in TimeSeriesBase.from_lal(cls, lalts, copy)
    699     warnings.warn("%s, defaulting to 'dimensionless'" % str(exc))
    700     unit = None
    701 channel = Channel(lalts.name, sample_rate=1/lalts.deltaT, unit=unit,
--> 702                   dtype=lalts.data.data.dtype)
    703 out = cls(lalts.data.data, channel=channel, t0=lalts.epoch,
    704           dt=lalts.deltaT, unit=unit, name=lalts.name, copy=False)
    705 if copy:

AttributeError: 'NoneType' object has no attribute 'dtype'
```

gwpy/utils/lal.py
```diff
@@ -144,6 +157,33 @@ def find_typed_function(pytype, prefix, suffix, module=lal):
...
+    """
+    cls = type(lalts)
+    try:
+        return NUMPY_DTYPE_FROM_LAL_CLASS[cls]
+    except KeyError:
+        raise ValueError(
+            f'Invalid class {cls.__name__!r} for the LAL time-series'
+        ) from None
+
+
```
duncanmmacleod, line 186: If we need another function to infer types from LAL objects (which you have demonstrated that we do), I would like to try and make it more general, perhaps something like this:

```python
def from_lal_type(laltype):
    if not isinstance(laltype, type):
        laltype = type(laltype)
    name = laltype.__name__
    for nptyp, laltyp in LAL_TYPE_STR_FROM_NUMPY.items():
        if laltyp in name:
            return nptyp
    raise ValueError(
        f"'{name}' has no known numpy type equivalent"
    )
```

I sort of hate this, but I can't find any better way of inferring the LAL type from the LAL object without hardcoding every possible object.

gwpy/timeseries/tests/test_io_gwf_lalframe.py
```diff
@@ -162,11 +167,21 @@ def test_write(tmp_path):
...
     # create timeseries with no IFO
-    data = TimeSeries([1, 2, 3, 4, 5])
-    tmp = tmp_path / "test.gwf"
-    gwpy_lalframe.write(
-        {None: data},
-        tmp
-    )
+    data = TimeSeries(data, dtype=float)
+    with NamedTemporaryFile(dir=tmp_path) as f:
+        gwpy_lalframe.write({None: data}, f.name)
```
duncanmmacleod, line 187: I don't understand the switch from using `tmp_path` as it is to using `NamedTemporaryFile`, can you explain this please?

duncanmmacleod (changes requested):
This will be a useful change, a couple of questions and one suggestion to make it more general.

Thanks @pchanial!

gwpy/timeseries/io/gwf/lalframe.py
```diff
@@ -153,9 +153,8 @@ def read(source, channels, start=None, end=None, series_class=TimeSeries,
     # read data
     out = series_class.DictClass()
     for name in channels:
-        out[name] = series_class.from_lal(
-            _read_channel(stream, str(name), start=start, duration=duration),
-            copy=False)
+        ts = _read_channel(stream, str(name), start=start, duration=duration)
+        out[name] = series_class.from_lal(ts, copy=False)
```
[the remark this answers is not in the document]
pchanial: This is for readability and our future debugging selves. That's one line less and more importantly, remove the indentations. Please confirm if you want me to revert this change.

gwpy/utils/lal.py
pchanial: I understand the motivation to generalize. I get the idea, but as a note, this won't work for UINT2TimeSeries, the loop will begin with (np.int16, 'INT2') and 'INT2' will match 'UINT2TimeSeries'.
The condition `if name.startswith(laltyp)` should work though but that is still fragile.
Otherwise a regex `(^(U?INT|REAL|COMPLEX)\d+)` to extract the LAL type would be fairly foolproof.

gwpy/timeseries/tests/test_io_gwf_lalframe.py
pchanial: That's me not liking hard-coded files. In this case it is harmless. I will revert the change.

gwpy/timeseries/tests/test_io_gwf_lalframe.py
```diff
@@ -162,9 +166,22 @@ def test_write(tmp_path):
...
 
-def test_write_no_ifo(tmp_path):
+@pytest.mark.parametrize(
+    'data',
+    [
+        pytest.param(
+            [],
+            marks=pytest.mark.xfail(
+                raises=RuntimeError,
+                reason='Cannot add an empty series to a frame',
```
duncanmmacleod, line 176: Please add `strict=True` here so that this cannot be fixed 'accidentally' and have the test still marked as xfail.

```suggestion
                reason='Cannot add an empty series to a frame',
                strict=True,
```

gwpy/utils/lal.py
```diff
@@ -70,6 +71,9 @@
 
 LAL_TYPE_STR_FROM_NUMPY = {k: LAL_TYPE_STR[v] for
                            (k, v) in LAL_TYPE_FROM_NUMPY.items()}
+LAL_NUMPY_FROM_TYPE_STR = {v: k for k, v in LAL_TYPE_STR_FROM_NUMPY.items()}
+
+REGEX_LAL_DTYPE = re.compile(r'(U?INT|REAL|COMPLEX)\d+')
```
duncanmmacleod, line 76: Can you please rename this to appease my pedantry?

```suggestion
LAL_TYPE_REGEX = re.compile(r'(U?INT|REAL|COMPLEX)\d+')
```

gwpy/utils/lal.py
```diff
@@ -144,6 +148,33 @@ def find_typed_function(pytype, prefix, suffix, module=lal):
     return getattr(module, '{0}{1}{2}'.format(prefix, laltype, suffix))
 
 
+def from_lal_type(laltype):
+    """Convert the data type of a LAL instance or type into a numpy data type.`
```
duncanmmacleod, line 152:
```suggestion
    """Convert the data type of a LAL instance or type into a numpy data type.
```

gwpy/utils/lal.py
```diff
@@ -144,6 +148,33 @@ def find_typed_function(pytype, prefix, suffix, module=lal):
...
+
+    Parameters
+    ----------
+    laltype : `SwigPyObject` or `type`
+        the input LAL instance or type
+
+    Returns
+    -------
+    dtype : `type`
+        the numpy data type, such as numpy.uint32, numpy.float64, etc.
```
duncanmmacleod, line 162:
```suggestion
        the numpy data type, such as `numpy.uint32`, `numpy.float64`, etc.
```

gwpy/utils/tests/test_lal.py
```diff
@@ -25,6 +25,7 @@
 
 from astropy import units
 
+
```
duncanmmacleod, line 28: Please remove this empty line.
```suggestion
```

gwpy/utils/tests/test_lal.py
```diff
@@ -52,11 +53,36 @@ def test_find_typed_function():
...
+        module=lalframe) is lalframe.FrStreamReadREAL4TimeSeries
+
+
+@pytest.mark.parametrize(
+    'dtype',
+    [
+        numpy.int16, numpy.int32, numpy.int64,
+        numpy.uint16, numpy.uint32, numpy.uint64,
+        numpy.float32, numpy.float64,
+        numpy.complex64, numpy.complex128,
```
duncanmmacleod, line 68: Please move these to one per line, so that other types can be added inline with minimal diffs.

```suggestion
        numpy.int16,
        numpy.int32,
        numpy.int64,
        numpy.uint16,
        numpy.uint32,
        numpy.uint64,
        numpy.float32,
        numpy.float64,
        numpy.complex64,
        numpy.complex128,
```

gwpy/utils/tests/test_lal.py
```diff
@@ -52,11 +53,36 @@ def test_find_typed_function():
...
+    func = utils_lal.find_typed_function(dtype, 'Create', 'TimeSeries')
+    lalts = func(None, None, 0, 1., None, 1)
+    assert utils_lal.from_lal_type(lalts) is dtype
+    assert utils_lal.from_lal_type(type(lalts)) is dtype
+
+
+@pytest.mark.parametrize(
+    'name',
+    ['XXX', 'XXXCOMPLEX64ZZZ', 'INNT2', 'INT5', 'REAL42'],
+)
```
duncanmmacleod, line 81: Can you move the parametrize list to one per line, possibly like this:

```suggestion
@pytest.mark.parametrize('name', [
    'XXX',
    'XXXCOMPLEX64ZZZ',
    'INNT2',
    'INT5',
    'REAL42',
])
```

duncanmmacleod (changes requested):
@pchanial, some more pedantic change requests. I like this change a lot, thanks again for putting it together.

gwpy/utils/lal.py
duncanmmacleod: Yes, good catch.

gwpy/utils/lal.py
pchanial: Done.

gwpy/timeseries/tests/test_io_gwf_lalframe.py
pchanial: Here it is: https://github.com/gwpy/gwpy/pull/1507

duncanmmacleod:
@pchanial, can you please rebase this again, there is now a conflicts. I'm in work today and tomorrow between weeks of leave, I'd like to try and get this finished off in this window if we can. I'm sorry it's taken so long already.

duncanmmacleod (approved):
Thanks @pchanial!

PR #1502, opened 2022-05-24
Improve dependencies workflow

duncanmmacleod:
This PR improves the dependencies workflow by removing the custom requirements parsing, instead just configure `pip` as much as we can to do what we want.


hoyc1 (approved):
LGTM!

PR #1507, opened 2022-06-01
Strict xfail.

pchanial:
Set pytest option so that XPASSED tests are considered as failures.


duncanmmacleod:
@pchanial, can you please rebase this to get the last failing CI job to pass? That should have been fixed by #1508.

duncanmmacleod (approved):
LGTM, thanks @pchanial.

PR #1510, opened 2022-06-08
Use plot directive to build CLI examples for online docs

duncanmmacleod:
Thus PR updates the sphinx `conf.py` to render the CLI examples to use the matplotlib `.. plot` directive to actually execute the example and generate the plot, which means that the examples are now executed as part of the main build in the same way as every other plot, rather than as part of the pre-processor function.

### DIFF
--- a/docs/cli/index.rst
+++ b/docs/cli/index.rst
@@ -1,20 +1,18 @@
-.. include: ..
-
 *******************************
 Command line plotting with GWpy
 *******************************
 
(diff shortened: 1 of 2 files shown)


duncanmmacleod:
An example docs build for this PR can be viewed by persons with LIGO.ORG credentials at <https://ldas-jobs.ligo.caltech.edu/~duncan.macleod/gwpy/1510/cli/#examples>.

duncanmmacleod:
This was included in #1511.

PR #1508, opened 2022-06-08
Update rstcheck to 6.x

duncanmmacleod:
This PR updates `rstcheck` in the Lint CI workflow to use `>=6.0.0`, which has renamed command-line options.

PR #1511, opened 2022-06-08
Update docs theme to sphinx-immaterial

### LINKED ISSUE #1509: Migrate docs theme to sphinx-immaterial
The `sphinx-immaterial` theme is a much more faithful copy of mkdocs-material than sphinx-material, we should migrate to that.

duncanmmacleod:
This PR closes #1509 by implementing updates to the docs to use the `sphinx-immaterial` theme.

Depends on #1508.


eagoetz:
@duncanmmacleod The example page seems to be rendered fine, so I give it my stamp of approval, but I'm also no expert on what the underlying code should look like. 

For my own curiosity: id you use another package to generate the code for the docs or did you do this all yourself by hand?

One other note: I see that you're fixing some of the docs by replacing single-backticks with double-backticks that had been missing. I'll note that there are other places in the docs where single-backticks are used where (I think) double-backticks should be used instead. I don't think it prevents this PR from being merged, just wanted to bring that to your attention.

eagoetz (approved):
LGTM

duncanmmacleod:
> For my own curiosity: id you use another package to generate the code for the docs or did you do this all yourself by hand?

I don't really understand what you are asking, so I'll cast a wide net, everything in the GWpy repo is written by hand.

> One other note: I see that you're fixing some of the docs by replacing single-backticks with double-backticks that had been missing. I'll note that there are other places in the docs where single-backticks are used where (I think) double-backticks should be used instead. I don't think it prevents this PR from being merged, just wanted to bring that to your attention.

Please open tickets identifying those issues, thanks.

eagoetz:
> > For my own curiosity: id you use another package to generate the code for the docs or did you do this all yourself by hand?
>
> I don't really understand what you are asking, so I'll cast a wide net, everything in the GWpy repo is written by hand.

Sorry that was unclear. I'm really surprised there's not an easier way to generate a file like `docs/conf.py`. Maybe my question is more basic: do you use any kind of automatically generated code that came from a package. Does Sphinx do this? I guess my questions are fairly basic...

>
> Please open tickets identifying those issues, thanks.

Ok, I'll try to flag some of them, but not sure I really have time to dig into the .rst files

duncanmmacleod:
> do you use any kind of automatically generated code that came from a package. Does Sphinx do this?

The `sphinx-quickstart` tool (from the Sphinx Python package) can build a default `conf.py`, which is what I started with originally. However, since then (~10 years ago) the entire thing has been rewritten several times.

PR #1512, opened 2022-06-10
Updating the name of the gravity spy server that hosts the glitch tables

scottcoughlin2014:
Updating the name of the gravity spy server that hosts the glitch tables


duncanmmacleod (approved):
`gravityspyplus`, sounds fancy. Thanks @scottcoughlin2014.

PR #1517, opened 2022-06-24
Workaround a bug in framecpp-swig

duncanmmacleod:
This PR works around an issue with python-ldas-tools-framecpp 2.6.13, see [computing/ldastools/LDAS_Tools#146](https://git.ligo.org/computing/ldastools/LDAS_Tools/-/issues/146).


hoyc1 (approved):
LGTM!

PR #1525, opened 2022-07-19
Pin experimental dependencies build to Python 3.9

duncanmmacleod:
This PR pins the experimental dependencies build to Python 3.9, since the crappy old version of PyYAML we need to support PyCBC (blame `pegasus-wms.common`) doesn't build with the advanced versions of cython and/or setuptools.


hoyc1 (approved):
LGTM!

PR #1530, opened 2022-08-01
Fix some line length issues

duncanmmacleod:
This PR addresses some line-length issues exposed by `flake8` (see [here](https://github.com/gwpy/gwpy/runs/7608965583)).

I don't understand why flake8 is only now catching these.


hoyc1 (approved):
LGTM!

PR #1533, opened 2022-08-02
Change the way Index.define creates arrays to prevent precision issues

### LINKED ISSUE #1506: Bug with irregular sampling rate
Hello! I found what looks like a bug in GWpy v2.1.3. 
Here's how to reproduce. This code works:

```python
import gwpy
import math
import numpy

a=numpy.zeros(500)
t=numpy.zeros(500)
for k in range(500):
    t[k]=k/100.
    a[k] = math.cos(50*float(k))+ 0.1*math.cos(float(k))


ts = TimeSeries(a, t0 = 1337089517.154, sample_rate = 100)
#ts.plot()
downsampled_ts = ts.resample(10)
downsampled_ts.plot()
downsampled_ts.psd()
```

And this code results in error ``AttributeError("This series has an irregular x-axis " "index, so 'dx' is not well defined")``
```python
import gwpy
import math
import numpy

a=numpy.zeros(500)
t=numpy.zeros(500)
for k in range(500):
    t[k]=k/100.
    a[k] = math.cos(50*float(k))+ 0.1*math.cos(float(k))


ts = TimeSeries(a, t0 = 1337089517.154, sample_rate = 100)
#Note this line below is UNCOMMENTED
ts.plot()
downsampled_ts = ts.resample(10)
downsampled_ts.plot()
downsampled_ts.psd()
```
Note just one line uncommented - to plot the time series, which supposedly should not _change_ anything further on, even if AttributeError is expected in some cases.

P.S.: Sorry if I formatting of this report is wrong (and please let me know why), this is the first time I'm opening an Issue.

duncanmmacleod:
This PR closes #1506 by refactoring the way that `Index.define()` creates arrays to prevent precision issues. A regression test has been introduced which demonstrates the precision issues describe in #1506.

### DIFF
--- a/gwpy/types/index.py
+++ b/gwpy/types/index.py
@@ -60,9 +60,14 @@ class Index(Quantity):
                 numpy.array(start, subok=True, copy=False).dtype,
                 numpy.array(step, subok=True, copy=False).dtype,
             )
-        start = start.astype(dtype, copy=False)
-        step = step.astype(dtype, copy=False)
-        return cls(start + numpy.arange(num, dtype=dtype) * step, copy=False)
+        start = Quantity(start, dtype=dtype, copy=False)
+        step = Quantity(step, dtype=dtype, copy=False).to(start.unit)
+        stop = start + step * num
+        return cls(
+            numpy.arange(start.value, stop.value, step.value, dtype=dtype),
+            unit=start.unit,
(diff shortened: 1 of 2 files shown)


eagoetz (approved):
LGTM

PR #1542, opened 2022-09-01
Update usage of pytest.raises throughout the codebase

### LINKED ISSUE #1541: Update use of pytest.raises to include match keyword
We should update the common pattern:

```python
with pytest.raises(Exception) as exc:
    thing()
assert str(exc.value) == message
```

to use the supported `match` keyword for pytest.raises, see https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions.

The pattern would then be

```python
with pytest.raises(Exception, match=message):
    thing()
```

duncanmmacleod:
This PR closes #1541 by updating the use of `pytest.raises` throughout the codebase to use the `match` keyword. This cleans things up a wee bit, and naturally allows more flexibility in matching the error message (since pytest uses a regex, rather than simple string comparisons).

### DIFF
--- a/gwpy/timeseries/tests/test_core.py
+++ b/gwpy/timeseries/tests/test_core.py
@@ -60,21 +60,30 @@ class TestTimeSeriesBase(_TestSeries):
         assert array.sample_rate == units.Quantity(1, 'Hertz')
         assert array.dt == units.Quantity(1, 'second')
 
-        # check handling of epoch vs t0
+    def test_new_epoch_t0(self):
+        """Test `gwpy.timeseries.TimeSeriesBase` handling of epoch vs t0.
+        """
         a = self.create(epoch=10)
         b = self.create(t0=10)
         utils.assert_quantity_sub_equal(a, b)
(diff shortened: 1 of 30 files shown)


duncanmmacleod:
Ping @gwpy/maintainers, please review this change.

gwpy/timeseries/tests/test_core.py
```diff
@@ -269,9 +278,13 @@ def test_span(self, instance):
         assert instance.span == reduce(
             operator.or_, (ts.span for ts in instance.values()), Segment(0, 0),
         )
-        with pytest.raises(ValueError) as exc:
+
+    def test_span_error_empty(self):
```
js01, line 282: could write docstrings for tests, at least for new ones

js01 (approved):
From what I can see, the rewritten pytest.raises look correct

For any new tests added, could be worth adding docstrings if desired

js01:
Codecov fails -- if I read that report correctly, coverage % is checked in the test files themselves

PR #1545, opened 2022-09-12
#1453 change whiten.py example to use open data

asouthgate:
See #1453.

Could not build with sphinx-immaterial on my machines, but should be OK. I guess a version incompatibility. Will open an issue.

### DIFF
--- a/examples/timeseries/whiten.py
+++ b/examples/timeseries/whiten.py
@@ -19,10 +19,6 @@
 
 """Whitening a `TimeSeries`
 
-.. warning::
-
-   This example requires LIGO.ORG credentials for data access.
-
 Most data recorded from a gravitational-wave interferometer carry information
 across a wide band of frequencies, typically up to a few kiloHertz, but
 often it is the case that the low-frequency amplitude dwarfs that of the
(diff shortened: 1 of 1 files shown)


examples/timeseries/whiten.py
```diff
@@ -31,31 +27,52 @@
...
+# We can see that, in the times series without pre-whitening, 
+# the gravitational wave signal at approximately 8 seconds is 
+# not easily visible. 
+# We can also see
 # tapering effects at the boundaries as the whitening filter settles in,
 # meaning that the first and last ~second of data are corrupted and should
 # be discarded before further processing.
+
+# 1. Abbott, et al. "Observation of gravitational waves from a binary black hole merger." Physical review letters 116.6 (2016): 061102. 
+# 2. Abbott, et al. "A guide to LIGO–Virgo detector noise and extraction of transient gravitational-wave signals." Classical and Quantum Gravity 37.5 (2020): 055002.
```
duncanmmacleod, line 78: It would be nice to use formal RST citations for the references, see https://docutils.sourceforge.io/docs/user/rst/quickref.html#citations.

duncanmmacleod (commented):
This is a great improvement, thanks @js01!

As we discussed on the phone, I think it would be nice to display the whitened-only data, as well as the whitened-and-bandpassed, to demonstrate that whitening improves the visibility of the signal, but you need bandpassing as well.

It might also be nice to zoom in to a 1-second window around the event.

PR #1547, opened 2022-09-30
Migrate all metadata out of setup.cfg into pyproject.toml

duncanmmacleod:
This PR migrates all package metadata into `pyproject.toml` now that setuptools supports it. This way everything is in `pyproject.toml`, rather than being split into `setup.cfg` as well.

This should have no practical impact on the package.


areeda (approved):
LGTM

PR #1553, opened 2022-11-01
Update veto-definer test example XML

duncanmmacleod:
This PR updates the example XML in the `gwpy.segments` tests for read/write of a Veto definer file.


areeda (approved):
LGTM

PR #1554, opened 2022-11-01
Handle optional kerberos with requests-ecp 0.3.0

duncanmmacleod:
This PR fixes the current experimental-dependencies test failure by updating the tests for `Channel.query` to not attempt to use kerberos.


areeda (approved):
LGTM

PR #1556, opened 2022-11-02
Minor restructuring of CI pipeline

duncanmmacleod:
This PR restructures the CI pipeline a bit to expand the `pip install` tests now that ligo-segments has wheels available for Windows.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -8,7 +8,7 @@
 #
 # -----------------------
 
-name: Build and test
+name: Conda tests
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


areeda:
I looked at this PR but don't understand enough of it to comment.

duncanmmacleod:
> I looked at this PR but don't understand enough of it to comment.

@areeda, thanks for looking; this PR is mainly housekeeping and doesn't do much. The main actual change is to expand the `dist` github actions workflow to try a standard `pip install gwpy && pytest` using both the tarball and the wheel (separately).

I also expanded that workflow to include Windows now that `ligo-segments` has wheels on pypi.org.

areeda (approved):
While I have a lot to learn about the CI, I think I now understand these changes well enough to approve thee PR

PR #1559, opened 2022-11-07
Improve handling/propagation of maya/dateparser errors and refactor gwpy.time tests

duncanmmacleod:
This PR improves the handling and propagation of errors from `maya/dateparser` when attempting to convert times.

This also refactors the test suite for `gwpy.time` to improve the tests and make them more readable. As part of this I replace the `freezegun` requirement with `pytest-freezegun`.

### DIFF
--- a/gwpy/time/_tconvert.py
+++ b/gwpy/time/_tconvert.py
@@ -263,9 +263,13 @@ def _str_to_datetime(datestr):
     # use maya
     try:
         import maya
-        return maya.when(datestr).datetime()
     except ImportError:
-        pass
+        MAYA_ERROR = None
+    else:
+        try:
+            return maya.when(datestr).datetime()
+        except Exception as exc1:
+            MAYA_ERROR = exc1
 
     # use dateutil.parse
     with warnings.catch_warnings():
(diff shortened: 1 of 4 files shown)


gwpy/time/_tconvert.py
```diff
@@ -263,19 +263,28 @@ def _str_to_datetime(datestr):
     # use maya
     try:
         import maya
-        return maya.when(datestr).datetime()
     except ImportError:
-        pass
+        MAYA_ERROR = None
```
js01, line 267: should MAYA_ERROR be none on ImportError? Is None how this error is encoded?

duncanmmacleod: This is eventually raise as a `RuntimeError` on L281-287. We hide the `ImportError` because we use `dateutil.parser` as a backstop if `maya` isn't available, so the `ImportError` isn't very informative for the user.

gwpy/time/tests/test_time.py
```diff
@@ -86,46 +72,108 @@ def _test_with_errors(func, in_, out):
...
+        ctx = mock.patch.dict("sys.modules", {"maya": None})
+    with ctx:
+        assert time.to_gps(in_) == out
+
+
+@pytest.mark.requires("maya")
+@pytest.mark.parametrize(("in_", "out"), [
+    ('Oct 30 2016 12:34 CST', 1161887657),
+])
+def test_to_gps_maya(in_, out):
```
js01, line 101: What's diff between 

Test that :func:`to_gps` works with and without maya

and

Test that :func:`gwpy.time.to_gps` works with maya.

js01 (approved):
Other than minor queries -- approved

PR #1563, opened 2022-11-10
Improve documentation of GWF I/O libraries

### LINKED ISSUE #1561: Improve documentation on how to install a GWF I/O library on various platforms
Inspired by https://git.ligo.org/computing/helpdesk/-/issues/2919 (LIGO.ORG login required) we should improve the documentation of the optional extra libraries required for I/O with GWF.

Currently the docs just say

https://github.com/gwpy/gwpy/blob/46ebb08d51f5c44116f60591947cca04216b8fea/docs/timeseries/io.rst#L108-L113

with no detail on how to install those tools.

duncanmmacleod:
This PR closes #1561 by improving the documentation of the optional GWF I/O libraries.

My solution is to provide our own 'external' pages for the various tools where the upstream reference doesn't do a good job of instructed users on how to install the library.

cc @jkanner

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -55,4 +55,4 @@ jobs:
             "rstcheck[sphinx,toml]>=6.0.0" \
         ;
     - name: Lint with rstcheck
-      run: rstcheck . --recursive --report-level error
+      run: rstcheck . --recursive --report-level error --log-level DEBUG
--- a/docs/external/framecpp.rst
+++ b/docs/external/framecpp.rst
@@ -1,0 +1,30 @@
+.. _gwpy-external-framecpp:
+
+########
+FrameCPP
+########
+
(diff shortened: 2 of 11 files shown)


eagoetz (approved):
I approve of the changes, but one thing remains fairly opaque to non-experts: which one or both of framecpp and framel need to be installed in order to use the GWF I/O within GWpy? If only one is needed, how does one decide which to install? Perhaps a sentence or two in the installation documentation would be helpful.

duncanmmacleod:
> I approve of the changes, but one thing remains fairly opaque to non-experts: which one or both of framecpp and framel need to be installed in order to use the GWF I/O within GWpy? If only one is needed, how does one decide which to install? Perhaps a sentence or two in the installation documentation would be helpful.

@eagoetz, this is a good point, so I've made some modifications to try and address this. Please see the demo docs here:

https://ldas-jobs.ligo.caltech.edu/~duncan.macleod/gwpy/1563/timeseries/io/#gwf

PR #1567, opened 2022-11-10
Update a number of CLI examples to use GWTC-3 data

duncanmmacleod:
This PR updates a number of CLI examples to use an event from GWTC-3, mainly to allow them to use more recent data that is easier to access.


duncanmmacleod:
@areeda, you can view a build of the docs with the updated examples at 

https://ldas-jobs.ligo.caltech.edu/~duncan.macleod/gwpy/1567/cli/#examples

PR #1568, opened 2022-11-10
Use consistent formatting for all argparse stuff in gwpy.cli

duncanmmacleod:
This PR modifies the style/layout of the argparse code in `gwpy.cli.cliproduct`, mainly to enable minimal diffs when future changes are made.

This should have zero functional impact (none of the keyword arguments _should_ have changed, just the layout of the code).

PR #1569, opened 2022-11-10
Support GWOSC as a data source for gwpy-plot

duncanmmacleod:
This PR adds a `--gwosc` option for `gwpy-plot` to support getting data from GWOSC.

Depends on #1568.


areeda (approved):
Looks good

PR #1570, opened 2022-11-23
Fix reading 'old-style' LIGO_LW files (with ILWD characters)

duncanmmacleod:
This PR fixes reading old-style LIGO_LW XML files that contain ILWD characters.

python-ligo-lw >=1.8.0 removed the ilwd char support, so we need to in-house some of that stuff.


duncanmmacleod:
@areeda, this fixes the bug you found trying to read that Omicron XML file, can you review this please?

areeda (approved):
LGTM

PR #1575, opened 2023-01-09
Update GWF writing for frameCPP 3.0.0

duncanmmacleod:
This PR addresses an incompatibility between `gwpy.io` and python-ldas-tools-framecpp 3.0.0; the compression options changed compared to frameCPP 2.9.


eagoetz (approved):
This seems reasonable to me, though I cannot comment on the `DefaultCompressionLevel` line. That line seems ok, but I'm not familiar enough with GZIP compressions versus others.

PR #1585, opened 2023-01-26
Enhance LaTeX detection

duncanmmacleod:
This PR enhances the `gwpy.plot.tex.has_tex` function that attempts to detect whether LaTeX can be used as a text renderer by attempting to draw (but not render) an image that actually uses latex.

This should amount to an end-to-end test such that if `has_tex` returns `True`, then the user has significant assurance that the LaTeX installation is complete enough for MPL.

### DIFF
--- a/gwpy/plot/tex.py
+++ b/gwpy/plot/tex.py
@@ -32,20 +32,40 @@ MACROS = [
 ]
 
 
+def _test_usetex():
+    """Draw (but don't show) a test image using matplotlib and LaTeX.
+    """
+    from matplotlib import (pyplot, rc_context)
+    with rc_context({"text.usetex": True}):
+        fig = pyplot.figure()
+        ax = fig.gca()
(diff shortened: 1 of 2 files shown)


gwpy/plot/tex.py
```diff
@@ -32,20 +32,40 @@
...
-    Checks for ``latex``, ``pdflatex``, and ``dvipng`` on the path.
+    Checks for ``latex``, ``pdflatex``, and ``dvipng`` on the path, and
+    then attemps to draw an image using LaTeX syntax.
 
     Returns
     -------
     hastex : `bool`
-        `True` if all required executables are found on the path, otherwise
-        `False`
+        `True` if the test image is drawn correctly, otherwise `False`
```
js01, line 56: pedantic maybe -- drawn at all or drawn correctly?

js01 (approved):

Issue #1596, opened 2023-03-03
Floating point errors when defining an Index


dethodav:
When creating a new `Index`, rounding of float values can result in an extra x-index being created. For example, the following code should generate 1001 indices, but instead generates 1002 indices: 

```
from gwpy.types.index import Index
x = Index.define(1262936373.4853957, 0.051, 1001)
print(x.shape)
``` 

This creates issues when creating spectrograms, as the number of points in the 2d array will not end up matching the number of points in the x-index: 
https://github.com/gwpy/gwpy/blob/de2299162736d745b7847b31cd1277805dcdb9a5/gwpy/signal/qtransform.py#L549-L553

In the above case, this bug leads to `nx != len(out.xindex.value)`.This then leads to errors when plotting the created spectrogram.

dethodav:
Issue addressed in #1600 - closing issue

PR #1600, opened 2023-03-28
Address floating point errors for index creation

dethodav:
This PR addresses #1596 and introduces a new test to confirm that this issue does not pop up again with the `Index` class.

### DIFF
--- a/gwpy/types/index.py
+++ b/gwpy/types/index.py
@@ -64,8 +64,8 @@ class Index(Quantity):
         step = Quantity(step, dtype=dtype, copy=False).to(start.unit)
         stop = start + step * num
         return cls(
-                numpy.arange(start.value, stop.value,
-                    step.value, dtype=dtype)[:num],
+            numpy.arange(start.value, stop.value,
+                         step.value, dtype=dtype)[:num],
             unit=start.unit,
             copy=False,
         )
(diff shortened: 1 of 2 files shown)


dethodav:
@duncanmmacleod Is this a reasonable way to fix this bug? This is causing issues right now with DQR tasks.

gwpy/types/index.py
```diff
@@ -64,7 +64,8 @@ def define(cls, start, step, num, dtype=None):
         step = Quantity(step, dtype=dtype, copy=False).to(start.unit)
         stop = start + step * num
         return cls(
-            numpy.arange(start.value, stop.value, step.value, dtype=dtype),
+            numpy.arange(start.value, stop.value,
+                         step.value, dtype=dtype)[:num],
```
duncanmmacleod, line 68:
```suggestion
            numpy.arange(
                start.value,
                stop.value,
                step.value,
                dtype=dtype,
            )[:num],
```

duncanmmacleod (changes requested):
@dethodav, floats are tricksy, so I think this is probably 'good enough' if it avoids the issues you reported.

Thanks for adding a regression test, that will help maintenance in the future.

I have requested one pedantic stylistic change, then we're good to go.

dethodav:
@duncanmmacleod thanks! I just implemented your suggestion

duncanmmacleod (approved):
Thanks @dethodav, looks great.

PR #1606, opened 2023-04-12
Fix tag error

vivienr:
`TimeSeries.fetch_open_data()` reports `TypeError: get_urls() got an unexpected keyword argument 'tag'`, this PR removes `tag` from `timeseries/core.py` and `timeseries/io/losc.py`.

### DIFF
--- a/gwpy/timeseries/core.py
+++ b/gwpy/timeseries/core.py
@@ -380,9 +380,9 @@ class TimeSeriesBase(Series):
 
     @classmethod
     def fetch_open_data(cls, ifo, start, end, sample_rate=4096,
-                        tag=None, version=None,
-                        format='hdf5', host=GWOSC_DEFAULT_HOST,
-                        verbose=False, cache=None, **kwargs):
+                        version=None, format='hdf5', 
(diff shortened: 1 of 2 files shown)


duncanmmacleod (approved):
Thanks @vivienr!

GregoryAshton:
@duncanmmacleod could I ask: is there a plan for when the next release will be made? This will be useful to understand if we should still pin gwosc to earlier versions or if we should wait until a release.

GregoryAshton:
> @duncanmmacleod could I ask: is there a plan for when the next release will be made? This will be useful to understand if we should still pin gwosc to earlier versions or if we should wait until a release.

Sorry, please ignore me I now see you have answered this elsewhere. Thanks!

duncanmmacleod:
> > @duncanmmacleod could I ask: is there a plan for when the next release will be made? This will be useful to understand if we should still pin gwosc to earlier versions or if we should wait until a release.
> 
> Sorry, please ignore me I now see you have answered this elsewhere. Thanks!

For the record, I will tag a new 3.0.4 release as soon as this is reviewed and ready.

PR #1608, opened 2023-04-14
Fix issues writing StateVector.bits to HDF5

duncanmmacleod:
This PR adds a test for, then fixes issues with, writing `StateVector` to HDF5.

There was an issue converting `Bits` to an array that could be embedded inside an HDF5 attribute.


eagoetz (approved):
LGTM

PR #1610, opened 2023-04-24
Fix plotting TimeSeriesDict on separate Axes

### LINKED ISSUE #1609: TimeSeriesDict plotting does not work as expected
The TimeSeriesDict `plot()` function is convenient and has several options like `sharex` and `separate`. But I have found that they don't work as expected. Consider the following example code - 

```
from gwpy.timeseries import TimeSeries, TimeSeriesDict
from gwpy.time import to_gps, from_gps
from gwpy.plot import Plot

start_gps = to_gps('2020-01-04T00:00:00')
end_gps = to_gps('2020-01-4T01:00:00')
example_chs = ['L1:ISI-GND_STS_ETMX_X_BLRMS_30M_100M.rms,s-trend', 'L1:ISI-GND_STS_ETMX_X_BLRMS_100M_300M.rms,s-trend', 'L1:ISI-GND_STS_ETMX_X_BLRMS_1_3.rms,s-trend']
data = TimeSeriesDict.get(example_chs, start=start_gps, end=end_gps, allow_tape=True)

plot = data.plot(label='key', separate=True, sharex=True)
plot.legend()
plot.show()
```

The output plot does not make separate plots for each item of the dictionary (i.e. each channel) - 
98373592-cb84-4a1b-8607-8a627a8a15cf

It's the same result if you use the Plot() function instead as I believe the data.plot() is using Plot() under the hood. 
```
plot = Plot(data, sharex=True, separate=True)
plot.show()
```

If you instead run the following code, you get the expected outcome of `separate` and `sharex` set to True but the plots are empty. 

```
plot = Plot(*data, sharex=True, separate=True)
plot.show()
```

c3e9dbef-c12a-4c44-927f-a92fec8f0fd1

If you run the following code, you get the right plots and `separate` and `sharex` work as expected but there is not way to satisfactorily add an unambiguous legend or title to the subplots to identify them by the channel name. 

```
plot = Plot(*[value for key, value in data.items()], sharex=True, separate=True)
plot.legend(list(data.keys()))
plot.show()
```

8e39415f-7cff-4e57-ba6f-8a5b5a3c9069

The expected behavior of setting `sharex` and `separate` to True for a TimeSeriesDict is a figure with subplots corresponding to the timeseries of each item in the dict where each subplot is identified (via a legend or title) with the channel name. If the legend for the separate subplots is going to be combined like in the last plot above, then the subplots need to be different colors. 

My GwPy version is 3.0.2

duncanmmacleod:
This PR fixes #1609 by adding a special case for `separate=True`.

### DIFF
--- a/gwpy/timeseries/core.py
+++ b/gwpy/timeseries/core.py
@@ -1441,9 +1441,18 @@ class TimeSeriesBaseDict(OrderedDict):
             all other keyword arguments are passed to the plotter as
             appropriate
         """
+        kwargs.update({
+            "method": method,
+            "label": label,
+        })
+
         # make plot
         from ..plot import Plot
-        plot = Plot(self.values(), method=method, label=label, **kwargs)
+
+        if kwargs.get("separate", False):
+            plot = Plot(*self.values(), **kwargs)
+        else:
+            plot = Plot(self.values(), **kwargs)
 
         # update labels
         artmap = {'plot': 'lines', 'scatter': 'collections'}
--- a/gwpy/timeseries/tests/test_core.py
+++ b/gwpy/timeseries/tests/test_core.py
@@ -420,6 +420,22 @@ class TestTimeSeriesBaseDict(object):
             plot.save(BytesIO(), format='png')
             plot.close()
 
+    def test_plot_separate(self, instance):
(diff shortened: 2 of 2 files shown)


eagoetz (approved):
LGTM

Issue #1493, opened 2022-04-19
GWPY resampling issue


kietphm:
There is an issue with the coherence function of gwpy.timeseries. It produces the larger than 1 coherence.
Besides, I tried to calculate coherence between DARM(H1:GDS-CALIB_STRAIN) and a channel (H1:SUS-OM1_M1_DAMP_L_IN1_DQ) with sampling frequency of 256Hz, and GWpy gives the frequencies series of up to ~8000 Hz. 

I dont think it makes sense to have information >128 Hz
I think there is an issue with resampling process in coherence calculation of GWPY.

```
from gwpy.timeseries import TimeSeriesDict
data = TimeSeriesDict.get(
    ['H1:GDS-CALIB_STRAIN', 'H1:SUS-OM1_M1_DAMP_L_IN1_DQ'],
    1245204018,
    1245204018 + 1800,
)
hoft = data['H1:GDS-CALIB_STRAIN']
acc = data['H1:SUS-OM1_M1_DAMP_L_IN1_DQ']
###Calculate coherence here
coh = hoft.coherence(acc, fftlength=2, overlap=0.5)
```
<img width="469" alt="image" src="https://user-images.githubusercontent.com/37116807/164055658-1cdd8e67-b8e5-4325-ac77-19223cc3827a.png">

I tried to go through the gwpy code and I think the issue came from the scipy.signal.resample() in line 955 at

https://github.com/gwpy/gwpy/blob/b5f791105efc5f78ce51f35455bcde8a5d36cef3/gwpy/timeseries/timeseries.py

I tried to use 'scipy.signal.decimate()' for resampling, and scipy.signal.coherence for coherence calculation and it gives a reasonable results. I tested out two filter provided by 'scipy.signal.decimate()', which are FIR AND IIR.

```
fsdown_darm = 16384
down_factor = int(16384//256)

###For FIR filter
downdata = sig.decimate(darm_data, down_factor,
                                ftype = 'fir',
                                zero_phase = True)
ftemp, Cxytemp = sig.coherence(test_data, downdata, fs = 256, nperseg=512)


                                
###For IIR filter
downdata2 = sig.decimate(darm_data, down_factor,
                                ftype = 'iir',
                                zero_phase = True)
ftemp, Cxytemp = sig.coherence(test_data, downdata2, fs = 256, nperseg=512)

```

Here are the results, they looks reasonable: 

[...]

PR #1611, opened 2023-04-24
Resample time series for coherence calculation when sampling rates di…

asouthgate:
Addresses: [#1493 ](https://github.com/gwpy/gwpy/issues/1493)

A new arg is added to _scipy.py::coherence. When sampling frequencies differ, the series with the higher timeseries.sample_rate is downsampled using timeseries.resample. By default, a warning is emitted. If downsample=False is specified, an exception is raised on differing rates. If downsample=True, downsampling is performed but no warning is emitted.

Fortunately @duncanmmacleod, in the examples, sampling frequencies are the same for compared timeseries, so the examples do not have this issue.

Edit: to add: I chose to address resampling in the coherence function itself, instead of in timeseries.coherence method since it is part of public interface which can be used by itself. It may be that elsewhere also needs the same treatment. We talked about cross correlation.


gwpy/signal/spectral/_scipy.py
```diff
@@ -200,6 +206,35 @@ def coherence(timeseries, other, segmentlength, noverlap=None, **kwargs):
...
+            if warn_fs:
+                warnings.warn("Sampling frequencies are unequal. Higher "
+                              "frequency series will be downsampled before "
+                              "coherence is calculated",
+                              category=UserWarning)
+                # downsample the one with the higher rate
+                if timeseries.sample_rate > other.sample_rate:
+                    timeseries = timeseries.resample(other.sample_rate)
+                else:
+                    other = other.resample(timeseries.sample_rate)
```
duncanmmacleod, line 232: This block is currently inside the `warn_fs` block, is tht right? Or should this be

```suggestion
            # downsample the one with the higher rate
            if timeseries.sample_rate > other.sample_rate:
                timeseries = timeseries.resample(other.sample_rate)
            else:
                other = other.resample(timeseries.sample_rate)
```

gwpy/signal/spectral/_scipy.py
```diff
@@ -200,6 +206,35 @@ def coherence(timeseries, other, segmentlength, noverlap=None, **kwargs):
...
+
+    # should never be equal from here on
+    if downsample is False:
+        assert timeseries.sample_rate == other.sample_rate
```
duncanmmacleod, line 236: Is this test necessary? If it should be kept, please add an explanatory comment (something along the lines of 'how did we get here?'

Also should the comment read 'should be equal from here on' (i.e. without the word 'never')?

gwpy/signal/tests/test_coherence.py
```diff
@@ -0,0 +1,155 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) Duncan Macleod (2014-2023)
```
duncanmmacleod, line 2:
```suggestion
# Copyright (C) Cardiff University (2023)
```

gwpy/signal/spectral/_scipy.py
asouthgate: Thanks, yes, fixed and added a test to account for the resulting problem

gwpy/signal/spectral/_scipy.py
asouthgate: Yes, should have been "never" or "not". It's now not conditional on downsample. Put assertion in there for "defensive programming". Can remove if desired.

gwpy/signal/spectral/_scipy.py
```diff
@@ -200,6 +206,34 @@ def coherence(timeseries, other, segmentlength, noverlap=None, **kwargs):
...
+    if downsample is None:
+        warn_fs = True
+        downsample = True
+
+    if timeseries.sample_rate != other.sample_rate:
+        # scipy assumes a single sampling frequency
+        if not downsample:
+            raise ValueError("Cannot calculate coherence when sampling "
+                             "frequencies are unequal")
+        else:
```
duncanmmacleod, line 222: Since the `if not downsample` clause raises an exception, the `else` is not required, so that can be removed and the whole of its contents just dedented by one level.

asouthgate: Removed as requested

duncanmmacleod (approved):
Thanks @asouthgate!

PR #1613, opened 2023-04-26
Update all GWOSC URLs to gwosc.org

duncanmmacleod:
This PR updates all references to `{www.}gw-openscience.org` to point at `gwosc.org`.


examples/miscellaneous/range-spectrogram.py
```diff
@@ -29,7 +29,7 @@
 __author__ = 'Alex Urban <alexander.urban@ligo.org>'
 
 # First, we need to load some data. As before we can `fetch` the
-# `public data <https://www.gw-openscience.org/catalog/>`__
+# `public data <https://gwosc.org/catalog/>`__
```
robertbruntz, line 32: I don't know if it matters, since this is a comment, but neither the original URL ([https://www.gw-openscience.org/catalog/](https://www.gw-openscience.org/catalog/)) nor the modified URL ([https://gwosc.org/catalog/](https://gwosc.org/catalog/)) works.

examples/miscellaneous/range-timeseries.py
```diff
@@ -31,7 +31,7 @@
 __credits__ = 'Alex Urban <alexander.urban@ligo.org>'
 
 # First, we need to load some data. We can `fetch` the
-# `public data <https://www.gw-openscience.org/catalog/>`__
+# `public data <https://gwosc.org/catalog/>`__
```
robertbruntz, line 34: I don't know if it matters, since this is a comment, but neither the original URL (https://www.gw-openscience.org/catalog/) nor the modified URL (https://gwosc.org/catalog/) works.

examples/timeseries/inject.py
```diff
@@ -41,7 +41,7 @@
 # Then we can download a simulation of the GW150914 signal from GWOSC:
 
 from astropy.utils.data import get_readable_fileobj
-url = ("https://www.gw-openscience.org/s/events/GW150914/P150914/"
+url = ("https://gwosc.org/s/events/GW150914/P150914/"
```
robertbruntz, line 44: Should this URL be world-readable?  Both the original and the modified URLs return `Forbidden` to me.

robertbruntz (approved):
The only things I see involve whether the target URL works or not, which is not what this PR is about and can be split off into a new issue or ignored.

LGTM

examples/miscellaneous/range-spectrogram.py
duncanmmacleod: Good catch, I'll fix that later...

examples/timeseries/inject.py
duncanmmacleod: I think it's just that this URL is out-of-date, and needs to be updated, I'll fix that later...

PR #1616, opened 2023-05-04
Add support for parsing multiple FFL files for a given dataset

duncanmmacleod:
This PR addresses https://git.ligo.org/computing/helpdesk/-/issues/3841 by adding support for parsing multiple FFLs for a single dataset at Virgo.

cc @dethodav

### DIFF
--- a/gwpy/io/ffldatafind.py
+++ b/gwpy/io/ffldatafind.py
@@ -38,6 +38,7 @@ from ligo.segments import (
 )
 
 from .cache import (
+    _CacheEntry,
     _iter_cache,
     cache_segments,
     file_segment,
@@ -157,7 +158,7 @@ def _read_ffls(site, tag, basedir=None):
     for ffl in _ffl_paths(site, tag, basedir=basedir):
         with open(ffl, "r") as fobj:
             entries.extend(
-                type(entry)(site, tag, entry.segment, entry.path)
+                _CacheEntry(site, tag, entry.segment, entry.path)
                 for entry in _iter_cache(fobj, gpstype=float)
             )
     return entries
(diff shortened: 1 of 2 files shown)


rhyspoulton (commented):
I have tested these changes and I can confirm that it is now able to deal with tag split across different ffl's with non-overlapping segments.

rhyspoulton:
@duncanmmacleod is it possible to get this merged?

gwpy/io/ffldatafind.py
```diff
@@ -147,15 +149,18 @@ def _ffl_path(site, tag, basedir=None):
...
-    with open(ffl, "r") as fobj:
-        return [
-            type(entry)(site, tag, entry.segment, entry.path)
-            for entry in _iter_cache(fobj, gpstype=float)
-        ]
+    entries = []
+    for ffl in _ffl_paths(site, tag, basedir=basedir):
+        with open(ffl, "r") as fobj:
+            entries.extend(
+                type(entry)(site, tag, entry.segment, entry.path)
```
asouthgate, line 160: why use type here instead of specific name?

asouthgate (approved):
Assuming ` type(entry)(site, tag, entry.segment, entry.path)` cannot be clearer, approved

gwpy/io/ffldatafind.py
duncanmmacleod: No good reason other than using `type` saves from importing the actual type object and using it explicitly. However, that is probably not a very good reason...

duncanmmacleod: @asouthgate, I have updated this code to use the actual type object, which is clearer.

asouthgate (approved):

PR #1619, opened 2023-05-05
Address inconsistency in transfer_function between CSD and PSD averaging methods

### LINKED ISSUE #1618: transfer_function needs a default average argument
The `gwpy.timeseries.transfer_function()` method needs a default method when computing the CSD and PSD used in the transfer function calculation. Currently if one doesn't specify the method, the results are incorrect, perhaps because CSD does not define a default value (so `'mean'` is used), but PSD does define a default value `'median'`.

See PSD: https://gwpy.github.io/docs/latest/api/gwpy.timeseries.TimeSeries/#gwpy.timeseries.TimeSeries.psd
See CSD: https://gwpy.github.io/docs/latest/api/gwpy.timeseries.TimeSeries/#gwpy.timeseries.TimeSeries.csd

And how they are used in the `transfer_function` method (note no `average` value is passed, so it uses the default from each): https://github.com/gwpy/gwpy/blob/eee114f78af4b838b927d5ba0af476e36e494880/gwpy/timeseries/timeseries.py#L1151-L1154

The simplest solution is to chose a default method for the transfer function, and then pass that to both the CSD and PSD methods

eagoetz:
This PR fixes incorrect results when no average method is supplied to the transfer_function method (I note here that attempting to supple the `method=` argument yields an error in the csd() method). Here we explicitly pass an averaging method to both the PSD and CSD methods so that the results are consistent.

Closes #1618


gwpy/timeseries/timeseries.py
```diff
@@ -1107,7 +1107,7 @@ def filter(self, *filt, **kwargs):
         return new
 
     def transfer_function(self, other, fftlength=None, overlap=None,
-                          window='hann', **kwargs):
+                          window='hann', average='mean', **kwargs):
```
duncanmmacleod, line 1110: This should use the `DEFAULT_FFT_METHOD` module variable to select the default:

```suggestion
                          window='hann', average=DEFAULT_FFT_METHOD, **kwargs):
```

This does change the default to `'median'`, but I think that should be ok.

duncanmmacleod (changes requested):
@eagoetz, one small change requested for this one, but am happy to hear an argument that the default for the `transfer_function()` really _should be_ `'mean'` as opposed to `'median'` (the global default).

eagoetz:
@duncanmmacleod I believe that as long as the CSD and PSD use the same averaging method, one should find a reasonably correct result. However, everywhere I've looked in references to transfer function calculations, they explicitly use the mean value for CSD and PSD, and not median. It could be for simplicity, I don't know. What I do know is that if the data is reasonably behaved, then the median or mean (for both CSD and PSD) should give approximately the same result.

duncanmmacleod (approved):
After discussion, this is ready to go, thanks @eagoetz.

PR #1621, opened 2023-05-12
Fix issues with Series.crop precision

### LINKED ISSUE #1601: Unexpected behavior when using crop function
The use of floor division in the `Series.crop()` function can result in unexpected behavior due to float precision: 
https://github.com/gwpy/gwpy/blob/5a8b781d6e0b398e4fe435a3f818d98ea2e04daf/gwpy/types/series.py#L1019

This generally results in one less sample returned than expected. For example, cropping a series with `dx=0.01` with `end=1.0` returns one fewer sample if the series x-values extend past 1.0: 
```
>>> from gwpy.types import Series
>>> Series([0]*100, dx=0.01).crop(end=1.).xindex[-1]
<Quantity 0.99>
>>> Series([0]*101, dx=0.01).crop(end=1.).xindex[-1]
<Quantity 0.98>
```
These errors are due to floating point precision. For example, we don't see the same issue when `dx=1`:
```
>>> Series([0]*100, dx=1).crop(end=100).xindex[-1]
<Quantity 99.>
>>> Series([0]*101, dx=1).crop(end=100).xindex[-1]
<Quantity 99.>
```

duncanmmacleod:
This PR fixes #1601 by tweaking the way that `Series.crop` rounds floats to integers, using `floor` instead of `//`.

I also cleaned up the code and its tests a bunch.

### DIFF
--- a/gwpy/types/series.py
+++ b/gwpy/types/series.py
@@ -1005,7 +1005,7 @@ class Series(Array):
             idx0 = None
         else:
             if not irregular:
-                idx0 = int((xtype(start) - x0) // self.dx.value)
+                idx0 = floor((xtype(start) - x0) / self.dx.value)
             else:
                 idx0 = numpy.searchsorted(
                     self.xindex.value, xtype(start), side="left"
@@ -1016,7 +1016,7 @@ class Series(Array):
             idx1 = None
         else:
             if not irregular:
-                idx1 = int((xtype(end) - x0) // self.dx.value)
+                idx1 = floor((xtype(end) - x0) / self.dx.value)
                 if idx1 >= self.size:
                     idx1 = None
             else:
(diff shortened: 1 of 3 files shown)


gwpy/types/series.py
```diff
@@ -978,18 +979,20 @@ def crop(self, start=None, end=None, copy=False):
...
         # pin late ends to time-series end
         if end == x1:
             end = None
         if end is not None and xtype(end) > x1:
-            warn('%s.crop given end larger than current end, '
-                 'crop will end when the Series actually ends.'
-                 % type(self).__name__)
+            warn(
+                f"{type(self).__name__}.crop given end larger than current "
+                "end, crop will begin when the Series actually ends.",
```
a-renzini, line 994: 'crop will begin when the series actually ends' -> 'crop will end when the series actually ends'?

Issue #1544, opened 2022-09-02
trouble with zpk methods


jrollins:
I've been having trouble getting the `TimeSeries.zpk()` and `FrequencySeries.zpk()` methods to produce sensible results.  Attached is a script that produce a plot (below) that demonstrates the issues.

![gwpy_zpk_confused](https://user-images.githubusercontent.com/9920104/188206594-a198ecd7-6105-48c0-99cd-aa5117ab9239.png)

The plot shows four traces:

- ASD of the original data (`TimeSeries.asd()`)
- ASD "filtered" with the zpk method (`TimeSeries.asd().zpk()`)
- original time series first zpk filtered, then converted to ASD (`TimeSeries.zpk().asd()`)
- original time series ASD multiplied by the frequency response of a ZPK filter generated with scipy

The last red trace is the expected result, the orange and green traces are definitely not the expected results.

I have a suspicion that the FrequencySeries.zpk method is filtering the frequency series as if it's a time series, which would obviously not be correct.  As for the ASD of the ZPK filtered time series it looks sort of like what happens if the data includes a large transient, but I played around with clipping off the ends of the data before producing the spectrum and I couldn't produce any sensible results.

It would be good if these methods could be fixed, or probably removed if they can't produce sensible expected results.

jrollins:
Stupid github won't allow me to just attach the python script :-1:
```
import numpy as np
import scipy.signal as sig
import matplotlib.pyplot as plt

from gwpy.timeseries import TimeSeries


#from joblib import Memory
#@Memory('_cache').cache
def get_ts(*args):
    return TimeSeries.get(*args)


def tf_zpk(freq, z, p, k):
    """frequency response of ZPK filter, in Hertz"""
    # correct the units for the zeros and poles, and "normalize" the
    # gain by multiplying by the product of the poles and dividing by
    # the product of the zeros
    z = -2 * np.pi * np.array(z)
    p = -2 * np.pi * np.array(p)
    k = (
        np.abs(np.prod(p[p != 0]) / np.prod(z[z != 0]))
        * ((2 * np.pi) ** p[p == 0].size)
        / ((2 * np.pi) ** z[z == 0].size)
        * k
    )
    w = 2 * np.pi * freq
    b, a = sig.zpk2tf(z, p, k)
    tf = sig.freqs(b, a, worN=w)[1]
    return tf


channel = 'H1:OAF-CLEAN_DELTAL_EXTERNAL_DQ'
[...]

PR #1626, opened 2023-05-24
Fix a bug with FrequencySeries analog zpk filters

asouthgate:
Should fix one of the problems with https://github.com/gwpy/gwpy/issues/1544, where using an analog filter on a FrequencySeries gives an incorrect answer, since it is converted to digital with bilinear(), then analog scipy methods used, lti instead of dlti etc.

I added another arg to parse_filter, partly for compatibility. If an analog filter is parsed, but we don't want to convert it to digital (it is unnecessary and unexpected in this case) with bilinear_zpk, we can use analog_to_digital=False. The function doesn't do much else in this case. The other option was to check if analog=True in filter(), then call parse_filter(analog=False), but that seemed wrong.

Another option was to convert all filters to digital and use digital methods inside fdfilter. But I thought that might be unexpected (if an analog filter can be used directly and is passed to zpk(), it probably is expected to be used with equivalent of freqs, not silently converted). Also, I think the result can be non-identical in some cases.

I added another test, expanded one of the existing tests (test_zpk didn't actually test the result, and failed if that assertion was added). May need to chat about some of the existing tests which passed but maybe should not have.

May be worth yet more tests, too.


asouthgate:
To add, to discuss:

1. Analog=True for .zpk(self..) methods as default but False for .filter(self,...) methods, which are called by default. 
2. Some tests need to be looked at. Previously used digital=False but compared to analog expected values. Did these tests pass because of the bug?

asouthgate:
Depends on https://github.com/gwpy/gwpy/pull/1632

PR #1632, opened 2023-05-31
Extract analog to digital filter conversion out of parse_filter 

asouthgate:
parse_filter previously converted analog filters to digital. This MR extracts this responsibility out into a separate function.

### DIFF
--- a/gwpy/plot/bode.py
+++ b/gwpy/plot/bode.py
@@ -208,7 +208,7 @@ class BodePlot(Plot):
                 frequencies *= dt
 
         # parse filter (without digital conversions)
-        _, fcomp = parse_filter(filter_, analog=False)
+        _, fcomp = parse_filter(filter_)
         if analog:
             _lti = lti(*fcomp)
         else:
--- a/gwpy/signal/filter_design.py
+++ b/gwpy/signal/filter_design.py
@@ -256,6 +256,56 @@ def fir_from_transfer(transfer, ntaps, window='hann', ncorner=None):
     return out
 
 
+def bilinear_zpk(zeros, poles, gain, fs=1.0, unit='Hz'):
+    """Convert an analogue ZPK filter to digital using a bilinear transform
+
+    Parameters
+    ----------
+    zeros : array-like
+        list of zeros
+
+    poles : array-like
+        list of poles
+
+    gain : `float`
+        filter gain
+
+    fs : `float`, `~astropy.units.Quantity`
+        sampling rate at which to evaluate bilinear transform, default: 1.
+
+    unit : `str`, `~astropy.units.Unit`
+        unit of inputs, one or 'Hz' or 'rad/s', default: ``'Hz'``
+
+    Returns
+    -------
+    zpk : `tuple`
+        digital version of input zpk
+    """
+    zeros = numpy.array(zeros, dtype=float, copy=False)
+    zeros = zeros[numpy.isfinite(zeros)]
+    poles = numpy.array(poles, dtype=float, copy=False)
+    gain = gain
+
+    # convert from Hz to rad/s if needed
+    unit = Unit(unit)
+    if unit == Unit('Hz'):
+        zeros *= -2 * pi
+        poles *= -2 * pi
+    elif unit != Unit('rad/s'):
+        raise ValueError("zpk can only be given with unit='Hz' "
+                         "or 'rad/s'")
+
+    # convert to Z-domain via bilinear transform
+    fs = 2 * Quantity(fs, 'Hz').value
+    dpoles = (1 + poles/fs) / (1 - poles/fs)
+    dzeros = (1 + zeros/fs) / (1 - zeros/fs)
+    dzeros = numpy.concatenate((
+        dzeros, -numpy.ones(len(dpoles) - len(dzeros)),
+    ))
+    dgain = gain * numpy.prod(fs - zeros)/numpy.prod(fs - poles)
+    return dzeros, dpoles, dgain
+
+
 def convert_to_digital(form, filter, sample_rate):
     """Convert an analog filter to digital via bilinear functions.
 
@@ -286,7 +336,7 @@ def convert_to_digital(form, filter, sample_rate):
         return 'ba', signal.bilinear(*filter, fs=sample_rate)
 
     if form == 'zpk':
-        return 'zpk', signal.bilinear_zpk(*filter, fs=sample_rate)
+        return 'zpk', bilinear_zpk(*filter, fs=sample_rate)
 
 
 def parse_filter(args):
--- a/gwpy/signal/filter_design.py
+++ b/gwpy/signal/filter_design.py
@@ -286,9 +286,6 @@ def bilinear_zpk(zeros, poles, gain, fs=1.0, unit='Hz'):
     poles = numpy.array(poles, dtype=float, copy=False)
     gain = gain
 
-    if len(zeros) > len(poles):
-        raise ValueError("More zeros than poles. Pad if required.")
-
     # convert from Hz to rad/s if needed
     unit = Unit(unit)
     if unit == Unit('Hz'):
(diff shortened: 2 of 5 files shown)


gwpy/signal/filter_design.py
```diff
@@ -306,7 +306,40 @@ def bilinear_zpk(zeros, poles, gain, fs=1.0, unit='Hz'):
...
+
+    if form not in ('zpk', 'ba'):
+        raise ValueError(f"Cannot convert form {form} to digital. Only"
+                         f"'zpk' or 'ba' are allowed.")
+
+    if form == 'ba':
+        return 'ba', signal.bilinear(*filter, fs=sample_rate)
+
+    if form == 'zpk':
+        return 'zpk', bilinear_zpk(*filter, fs=sample_rate)
```
duncanmmacleod, line 339: It might be nice, if a little redundant, to have `convert_to_digital` call out to `parse_filter` to get things in a standard format, rather than requiring an extra argument, or introspecting the input object.

gwpy/signal/filter_design.py
```diff
@@ -306,7 +306,40 @@ def bilinear_zpk(zeros, poles, gain, fs=1.0, unit='Hz'):
     return dzeros, dpoles, dgain
 
 
-def parse_filter(args, analog=False, sample_rate=None):
+def convert_to_digital(form, filter, sample_rate):
```
duncanmmacleod, line 309: It is reasonable to call this method `bilinear`, since that's what it actually does, or is that just confusing for the user?

gwpy/signal/tests/test_filter_design.py
```diff
@@ -153,3 +153,27 @@ def test_parse_filter():
     parsed = filter_design.parse_filter(zpk)
     assert parsed[0] == 'zpk'
     utils.assert_zpk_equal(parsed[1], zpk)
+
+
+def test_convert_to_digital():
+
```
duncanmmacleod, line 159:
```suggestion
```

gwpy/signal/tests/test_filter_design.py
```diff
@@ -153,3 +153,27 @@ def test_parse_filter():
...
+    assert dformba == 'ba'
+    b2, a2 = dfiltba
+    b3, a3 = signal.bilinear(b, a, fs=fs)
+    assert all(b3 == b2)
+    assert all(a3 == a2)
+
+    with pytest.raises(ValueError, match='Cannot convert'):
+        filter_design.convert_to_digital(
+            "badstr", ([1], [1], 1), fs
+        )
```
duncanmmacleod, line 179: Can this test be separated into a number of smaller, independent tests?

gwpy/timeseries/timeseries.py
```diff
@@ -1078,11 +1078,13 @@ def filter(self, *filt, **kwargs):
...
-            filt,
-            analog=kwargs.pop('analog', False),
-            sample_rate=self.sample_rate.to('Hz').value,
-        )
+        form, filt = filter_design.parse_filter(filt)
+
+        if kwargs.pop('analog', False):
+            form, filt = filter_design.convert_to_digital(
+                form, filt, self.sample_rate.to('Hz').value
+            )
```
duncanmmacleod, line 1086: Please reformat this to one argument per-line, to minimise future diffs.
```suggestion
            form, filt = filter_design.convert_to_digital(
                form,
                filt,
                self.sample_rate.to('Hz').value,
            )
```

gwpy/plot/bode.py
```diff
@@ -208,7 +208,7 @@ def add_filter(self, filter_, frequencies=None, dB=True,
                 frequencies *= dt
 
         # parse filter (without digital conversions)
-        _, fcomp = parse_filter(filter_, analog=False)
```
asouthgate, line 211: Here we assume the comment is correct. No digital conversion is required.

gwpy/signal/filter_design.py
```diff
@@ -306,7 +309,51 @@ def bilinear_zpk(zeros, poles, gain, fs=1.0, unit='Hz'):
...
+
+    if form == 'ba':
+        b, a = filter
+        if len(a) < len(b) and a[0] == 1:
+            a = numpy.append(a, numpy.zeros(len(b)-len(a)))
+        filter = signal.tf2zpk(b, a)
+
+    filter = bilinear_zpk(*filter, fs=sample_rate, unit=unit)
+
+    return parse_filter(filter)
```
asouthgate, line 353: second parse_filter could be replaced with return "zpk", filter

gwpy/signal/filter_design.py
```diff
@@ -306,7 +309,51 @@ def bilinear_zpk(zeros, poles, gain, fs=1.0, unit='Hz'):
...
+    form, filter = parse_filter(filter)
+
+    if form not in ('ba', 'zpk'):
+        raise ValueError(f"Cannot convert {form}, only 'zpk' or 'ba'")
+
+    if form == 'ba':
+        b, a = filter
+        if len(a) < len(b) and a[0] == 1:
+            a = numpy.append(a, numpy.zeros(len(b)-len(a)))
+        filter = signal.tf2zpk(b, a)
```
asouthgate, line 349: if not desired, alternative is to raise an exception

gwpy/signal/filter_design.py
```diff
@@ -286,6 +286,9 @@ def bilinear_zpk(zeros, poles, gain, fs=1.0, unit='Hz'):
     poles = numpy.array(poles, dtype=float, copy=False)
     gain = gain
 
+    if len(zeros) > len(poles):
```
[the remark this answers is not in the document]
duncanmmacleod: What happens is this logic is omitted and the user provides more zeros than poles?

gwpy/signal/filter_design.py
duncanmmacleod: Given that `filter` is always going to be the output of `bilinear_zpk`, we know the form, so shouldn't have to parse it. Can we just return the output of `bilinear_zpk`?

gwpy/signal/filter_design.py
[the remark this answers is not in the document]
asouthgate: You get an error from numpy, but it's not necessarily informative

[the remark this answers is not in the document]
duncanmmacleod: An error from what? Is it possible to just catch that error when it happens and improve the error message?

gwpy/signal/filter_design.py
```diff
@@ -306,7 +309,50 @@ def bilinear_zpk(zeros, poles, gain, fs=1.0, unit='Hz'):
...
+
+    Returns
+    -------
+    dform : `str`
+        type of filter, 'zpk' format
+
+    dfilter: 'tuple'
+        digital filter values
+    """
+
```
duncanmmacleod, line 334:
```suggestion
```

gwpy/signal/filter_design.py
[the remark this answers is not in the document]
asouthgate: Error comes in negative arg to 
`numpy.ones(len(dpoles) - len(dzeros))`

Scipy has a similar check
    degree = len(p) - len(z)```

    if degree < 0:
        raise ValueError("Improper transfer function. "
                         "Must have at least as many poles as zeros.")
    else:
```
But yes, we can just handle it. Happy to remove this and add a test for the next MR instead. I don't have a strong preference.

File: gwpy/signal/tests/test_filter_design.py [hunk 11]
```diff
@@ -18,15 +18,15 @@
...
-
 import numpy
+import pytest
 
 from scipy import signal
 
 from astropy import units
 
+from unittest import mock
+
```
duncanmmacleod, line 29: Pedantically, since `unittest` is a standard library module, this should be ordered above the others:

```suggestion
from unittest import mock

import numpy

import pytest

from scipy import signal

from astropy import units
```

duncanmmacleod (approved):
Thanks @asouthgate!

PR #1639, opened 2023-06-09
Add new scan feature for TimeSeries.find

### LINKED ISSUE #1638: Add TimeSeries data-access method to read data from a directory
@stuartthebruce has suggested a new data-discovery method that is half-way between `TimeSeries.read` (which requires a list of files) and `TimeSeries.find` (which calls out to gwdatafind) which does 'here's a directory and some parameters, go find my data in there and read them'.

The use case for LIGO is point GWpy at `/dev/shm{/kafka}` and have it read the data from there for a given GPS interval.

duncanmmacleod:
This PR adds a new `scan` keyword argument to `TimeSeries.find` that allows for manually scanning a directory hierarchy instead of calling out to `gwdatafind`, e.g (from LIGO computing centres):

```python
from gwpy.time import to_gps
print(TimeSeries.get(
    "L1:GDS-CALIB_STRAIN",
    to_gps("now") - 100,
    to_gps("now") - 90,
    frametype="L1_llhoft",
    verbose=True,
    scan="/dev/shm/kafka",
))
```

Closes #1638.

### DIFF
--- a/gwpy/io/datafind.py
+++ b/gwpy/io/datafind.py
@@ -45,7 +45,10 @@ from unittest import mock
 
 import gwdatafind
 
-from ligo.segments import segment as LigoSegment
+from ligo.segments import (
+    segment as LigoSegment,
+    segmentlist as LigoSegmentList,
+)
 
 from ..time import to_gps
 from . import ffldatafind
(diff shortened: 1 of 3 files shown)


dethodav:
@duncanmmacleod These feature would be extremely useful for the Data Quality Report! Echoing something @areeda had said on Mattermost, it would be helpful if gwpy defaulted to a different data find method if the scan does not return the requested data, similar to how `TimeSeries.get()` first tries `find()` and then `fetch()`. I think a reasonable order to `TimeSeries.get()` to look for data would be
1. `scan()`
2. `find()`
3. `fetch()`

rutujagurav:
Any updates on this feature? In what version of gwpy is this available?

duncanmmacleod:
Migrated to https://gitlab.com/gwpy/gwpy/-/merge_requests/1639.

PR #1647, opened 2023-07-17
Refactor colorbar generation with inset axes

duncanmmacleod:
This PR refactors the way that colourbars are generated using inset Axes to use the native `Axes.inset_axes` method, rather than calling out to `mpl_toolkits.axes_grid1` manually. This should fix compatibility with matplotlib 3.7.2, which is causing some CI failures elsewhere.


gwpy/plot/colorbar.py
```diff
@@ -155,67 +148,110 @@ def _scale_height(value, ax):
...
+
+
+def make_inset_axes(
+    ax,
+    location='right',
+    width=0.012,
+    length=1.,
+    pad=None,
+    **kwargs,
+):
```
asouthgate, line 232: docstring for non-"private" func?

gwpy/plot/colorbar.py
```diff
@@ -155,67 +148,112 @@ def _scale_height(value, ax):
...
-        inset_kw[key] = kwargs.pop(key)
-
-    return inset_axes(ax, **inset_kw), kwargs
+        raise ValueError(f"inset Axes location '{location}' not recognised")
+
+    # set default ticklocation for the right location/orientation
+    if location == "top" and orientation == "horizontal":
+        kwargs.setdefault("ticklocation", "top")
+    if location == "left" and orientation == "vertical":
+        kwargs.setdefault("ticklocation", "left")
```
asouthgate, line 247: what happens if location == "bottom" or "right"?

duncanmmacleod: Then the 'default' `ticklocation` is used, which is `bottom` or `right` respectively.

gwpy/plot/colorbar.py
```diff
@@ -155,67 +148,112 @@ def _scale_height(value, ax):
...
+
+    # where to start on the long axis (for centre-aligned Axes)
+    l0 = .5 - length / 2.
+
+    if location == "left":
+        return 0 - pad - width, l0, width, length
+    if location == "right":
+        return 1 + pad, l0, width, length
+    if location == "top":
+        return l0, 1 + pad, length, width
```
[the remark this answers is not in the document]
duncanmmacleod: The return is explicitly `x0`, `y0`, `width` (size left-to-right), and `height` (size top-to-bottom). The input arguments are where I tried to use 'sensible' names for the long and short axes of the colorbar, being `length` and `width` respectively. The catch is that for a horizontal colorbar the `width` input corresponds to the top-to-bottom size, which is actually the `height`, while `length` is the left-to-right size -> `width`. 

One option is to use `breadth` or similar for the input argument, so as to not overload the term `width`, but I'm not convinced.

asouthgate (approved):

PR #1649, opened 2023-07-20
Don't round TimeSeries.dt when setting via sample_rate

### LINKED ISSUE #1646: TimeSeries with sample_rate > 0.1 GHz is buggy
It seems there are several issues popping up when trying to use TimeSeries with high sampling rates, 0.1 GHz and above.

Below is a minimal code example to reproduce. Tested with GWpy **3.0.5**.

```python
import numpy as np
np.random.seed(0)
series = np.random.normal(size=10000)
test_ts = TimeSeries(series, sample_rate = 100000000)
# asd fails with "ValueError: cannot convert float NaN to integer"
test_ts.asd()
# plot also fails with "ValueError: arange: cannot compute length"
test_ts.plot()
```

duncanmmacleod:
This PR fixes #1646 by removing a rounding of `self.dt` when that is set via the `sample_rate` property. I can't remember why that was there in the first place, but it's clearly broken for very high sample rates (for which the `dt` is very small).

Closes #1646.

### DIFF
--- a/gwpy/timeseries/tests/test_core.py
+++ b/gwpy/timeseries/tests/test_core.py
@@ -107,26 +107,48 @@ class TestTimeSeriesBase(_TestSeries):
             a.t0, units.Quantity(1126259462, 's'))
 
     def test_sample_rate(self):
-        """Test `gwpy.timeseries.TimeSeriesBase.sample_rate`
+        """Test `gwpy.timeseries.TimeSeriesBase.sample_rate`.
         """
         # check basic conversion from dt -> sample_rate
         a = self.create(dt=0.5)
         assert a.sample_rate == 2 * units.Hz
 
+    def test_sample_rate_del(self, array):
+        """Test that `sample_rate` cannot be deleted.
+        """
         # test that we can't delete sample_rate
-        with pytest.raises(AttributeError):
-            del a.sample_rate
+        with pytest.raises(
+            AttributeError,
+            match="(can't delete attribute|has no deleter)",
+        ):
+            del array.sample_rate
 
+    def test_sample_rate_none(self, array):
+        """Test that `sample_rate = None` is effectively a deletion.
+        """
         # check None gets preserved
-        a.sample_rate = None
-        with pytest.raises(AttributeError):
-            a._t0
+        array.sample_rate = None
+        with pytest.raises(AttributeError, match="_t0"):
+            array._t0
 
-        # check other types
-        a.sample_rate = units.Quantity(128, units.Hz)
-        utils.assert_quantity_equal(a.dt, units.s / 128.)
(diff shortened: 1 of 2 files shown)


gwpy/timeseries/tests/test_core.py
```diff
@@ -107,26 +107,45 @@ def test_epoch(self):
...
         a = self.create(dt=0.5)
         assert a.sample_rate == 2 * units.Hz
 
+    def test_sample_rate_del(self, array):
+        """Test that `sample_rate` cannot be deleted.
+        """
         # test that we can't delete sample_rate
         with pytest.raises(AttributeError):
-            del a.sample_rate
+            del array.ample_rate
```
asouthgate, line 121: ample_rate?

asouthgate (changes requested):
Will approve assuming there is no need to round dt to nearest int, failing test passes, and ample_rate -> sample_rate

gwpy/timeseries/tests/test_core.py
duncanmmacleod: That will definitely raise an `AttributeError`, good catch.

duncanmmacleod:
I found the original commit for the `round()` call: https://github.com/gwpy/gwpy/commit/ad18c91e60ecde37fb2ed3748a54b19ac6236e21. That commit is included in _every_ tag, so has been there for all of time. But, the commit message doesn't give much away, isn't part of a pull request, and didn't come with any tests. Bad past me.

duncanmmacleod:
Some evidence that the rounding doesn't seem to be needed:

```python
>>> from gwpy.timeseries import TimeSeries
>>> rates = [1/16, 1/8, 1/4, 1/2, 1, 2, 4, 8, 16, 32, 100, 256, 512, 1024, 2048, 4096, 16384, 32768]
>>> for rate in rates:
...     a = TimeSeries(range(1000), sample_rate=rate)
...     print(a.dt, a.sample_rate)

16.0 s 0.0625 Hz
8.0 s 0.125 Hz
4.0 s 0.25 Hz
2.0 s 0.5 Hz
1.0 s 1.0 Hz
0.5 s 2.0 Hz
0.25 s 4.0 Hz
0.125 s 8.0 Hz
0.0625 s 16.0 Hz
0.03125 s 32.0 Hz
0.01 s 100.0 Hz
0.00390625 s 256.0 Hz
0.001953125 s 512.0 Hz
0.0009765625 s 1024.0 Hz
0.00048828125 s 2048.0 Hz
0.000244140625 s 4096.0 Hz
6.103515625e-05 s 16384.0 Hz
3.0517578125e-05 s 32768.0 Hz
```

asouthgate (approved):

PR #1651, opened 2023-07-31
Fix a bug with frequency domain filtering

asouthgate:
Fix a bug where frequency domain filtering was not returning an expected result because analog scipy methods were being used despite default analog=False args. 

Now, analog and digital filters use corresponding scipy functions (freqs_zpk and freqz_zpk) that take into account fs.

Also, two tests were added, and some test assertions, to check results of frequency domain filtering for some simple examples.

In a few places, tests have changed.

### DIFF
--- a/gwpy/frequencyseries/_fdcommon.py
+++ b/gwpy/frequencyseries/_fdcommon.py
@@ -54,16 +54,17 @@ def fdfilter(data, *filt, **kwargs):
     form, filt = parse_filter(filt)
     freqs = data.frequencies.value.copy()
 
-    if not analog:
+    if analog:
+        lti = signal.lti(*filt).to_zpk()
+        z, p, k = lti.zeros, lti.poles, lti.gain
+        w, fr = signal.freqs_zpk(z, p, k, worN=freqs)
+    else:
         lti = signal.dlti(*filt).to_zpk()
         z, p, k = lti.zeros, lti.poles, lti.gain
         # dlti.freqresp does not take into account fs
         # better to use the more straightforward functions
         w, fr = signal.freqz_zpk(z, p, k, worN=freqs, fs=fs)
-    else:
-        lti = signal.lti(*filt).to_zpk()
-        z, p, k = lti.zeros, lti.poles, lti.gain
-        w, fr = signal.freqs_zpk(z, p, k, worN=freqs)
+
     fresp = numpy.nan_to_num(abs(fr))
 
     # apply to array
--- a/gwpy/frequencyseries/_fdcommon.py
+++ b/gwpy/frequencyseries/_fdcommon.py
@@ -50,15 +50,21 @@ def fdfilter(data, *filt, **kwargs):
 
     # parse filter
     if fs is None:
-        fs = 2 * (data.shape[-1] * data.df).to('Hz').value
+        fs = 2 * data.frequencies[-1].to('Hz').value
     form, filt = parse_filter(filt)
-    if analog:
-        form, filt = convert_to_digital(filt, fs)
-    lti = signal.lti(*filt)
-
-    # generate frequency response
     freqs = data.frequencies.value.copy()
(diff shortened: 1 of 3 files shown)


gwpy/frequencyseries/_fdcommon.py
```diff
@@ -50,15 +50,21 @@ def fdfilter(data, *filt, **kwargs):
 
     # parse filter
     if fs is None:
-        fs = 2 * (data.shape[-1] * data.df).to('Hz').value
```
asouthgate, line 53: I think technically the old version is actually off by 2 * delta f. Data.shape[-1] gives len L.
e.g. freqs = [0, 1, 2..99], 2 * 100 * 1 = 200. Whereas it should be 2 * 99  = 198?

gwpy/frequencyseries/tests/test_frequencyseries.py
```diff
@@ -128,15 +128,58 @@ def test_ifft(self):
         utils.assert_quantity_sub_equal(ts.fft().ifft(), ts)
         utils.assert_allclose(ts.fft().ifft().value, ts.value)
 
-    def test_filter(self, array):
```
asouthgate, line 131: I renamed this test to test_filter_analog and added a test_filter_digital

gwpy/frequencyseries/tests/test_frequencyseries.py
```diff
@@ -128,15 +128,58 @@ def test_ifft(self):
...
+        fr_a = abs(fr_a)
+
+        assert numpy.allclose(a2.value, fr_a * array.value)
+        assert numpy.allclose(a3.value, fr_d * array.value)
+
+        # check that, in this case, digital and analog filters are close
+        # this check is not essential, since there is no guarantee
+        # filters will have the same frequency response after bilinear
+        # transform, but it is still useful
+        eps = 0.1 * numpy.abs(numpy.max(a3))
```
asouthgate, line 181: At the tail of the filtered values, they are close, but one of them is approximately zero, the other is small but not too near zero. So relative differences don't work. eps here is an absolute tol.

gwpy/frequencyseries/_fdcommon.py
duncanmmacleod: This depends heavily how the frequency series was generated. In general, we cannot guarantee what `fs` is given an arbitrary array input, but using `frequencies[-1]` is probably the easiest to document and use.

gwpy/frequencyseries/_fdcommon.py
```diff
@@ -50,15 +50,21 @@ def fdfilter(data, *filt, **kwargs):
...
+    if not analog:
+        lti = signal.dlti(*filt).to_zpk()
+        z, p, k = lti.zeros, lti.poles, lti.gain
+        # dlti.freqresp does not take into account fs
+        # better to use the more straightforward functions
+        w, fr = signal.freqz_zpk(z, p, k, worN=freqs, fs=fs)
+    else:
+        lti = signal.lti(*filt).to_zpk()
+        z, p, k = lti.zeros, lti.poles, lti.gain
+        w, fr = signal.freqs_zpk(z, p, k, worN=freqs)
```
duncanmmacleod, line 66: Just curious, why is the 'if not analog' clause first here? As opposed to having the 'if analog' clause first and then 'else' for the current 'if not analog'.

asouthgate: No reason, I think. Probably just coded the digital case first. Can change if preferred.

duncanmmacleod:
@asouthgate, can you please rebase this branch, hopefully that should integrate #1654 and resolve the `Lint` CI job failure.

gwpy/frequencyseries/_fdcommon.py
duncanmmacleod: Yes please, I find it a little jarring to read.

duncanmmacleod (approved):
LGTM, thanks @asouthgate.

Issue #1652, opened 2023-08-01
gwpy-plot qtransform fails if more than 1 plot duration is specified


areeda:
This seems to be an issue that produced a warning that has become an error with a current release of astropy.units.
The error message produced is:
```
  File "/Users/areeda/mambaforge/envs/igwn-py310/lib/python3.10/site-packages/gwpy/cli/qtransform.py", line 218, in get_spectrogram
    if self.qxfrm_args.get("whiten"):
  File "/Users/areeda/mambaforge/envs/igwn-py310/lib/python3.10/site-packages/astropy/units/quantity.py", line 1331, in __bool__
    raise ValueError(
ValueError: FrequencySeries truthiness is ambiguous, especially for logarithmic units and temperatures. Use explicit comparisons.
```
This is a result of the qtransform code:
```
            # use the precomputed ASD as the whitener if needed
            if self.qxfrm_args.get("whiten"):
                self.qxfrm_args["whiten"] = asd
```
That works for the first image where `self.qxfrm_args["whiten"]` is True but fails on the secod where it contains`asd` a FrequencySeries object.

duncanmmacleod:
@areeda, can you provide a command-line call that reproduces the failure?

areeda:
Sorry for the delay I missed the request
```
gwpy-plot qtransform --chan L1:GDS-CALIB_STRAIN --gps 1375928212.0000 --outdir . --plot 1.00 4.00 16.00
```
Will create 1 image not 3.

duncanmmacleod:
Migrated to https://gitlab.com/gwpy/gwpy/-/issues/1652, please join us on gitlab to continue your contribution!

PR #1653, opened 2023-08-01
fix test of whitening

areeda:
Trivial one line to address issue #1652 "truthiness" Value error.


areeda:
This is not te way to do it.

PR #1654, opened 2023-08-08
Fix type comparison lint

duncanmmacleod:
This PR fixes some lint complaints regarding exact type comparisons. It seems flake8 6.1.0 is more discerning than 6.0.0, which is probably for the best.

### DIFF
--- a/gwpy/types/series.py
+++ b/gwpy/types/series.py
@@ -839,8 +839,7 @@ class Series(Array):
             N = min(self.shape[0], other.shape[0])
 
         # if units are the same, can shortcut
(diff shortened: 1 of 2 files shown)


asouthgate:
Looks fine. The second round of assertions in the round to power test could have a comment to say test that float in gives float out. But that is outside of scope of change.

asouthgate (approved):

duncanmmacleod:
> Looks fine. The second round of assertions in the round to power test could have a comment to say test that float in gives float out. But that is outside of scope of change.

@asouthgate are you happy to formally approve the change so that it can be merged?

Issue #1656, opened 2023-08-09
Unexpected behavoir when using crop function


dethodav:
This issue was first mentioned in #1601 and fix was merged in #1621. Using gwpy==3.0.5, I still see unexpected behavior related to floating point precision with the `crop()` function. 

The following example shows the problem: 
```
len = 100
right = 0
wrong = 0
for x_max in list(numpy.random.random(100)) :
  x_series = numpy.linspace(0,x_max, len)
  series = Series([0]*len, xindex=x_series)
  expected = series.xindex[-2]
  if series.crop(end=x_max).xindex[-1] == expected:
    right += 1
  else:
    wrong += 1

print('RIGHT', right)
print('WRONG', wrong)
```
In this example, I generate a series with 100 samples and a final index between 0 and 1. I then crop the Series at the chosen final index. Whether the new series will crop 1 or 2 points from the original Series is random. Cropping 1 point happens ~95% of the time and cropping 2 points happens ~5% of the time.

asouthgate:
It's occasionally cutting off one too many values for regular spaced arrays. I suspect it is a precision issue with float division in Series.py:1030. 

`idx1 = floor((xtype(end) - x0) / self.dx.value)`

For example if L=500, end=0.508463154883984, dx=0.001018964238244457, this gives 498.99999999999994, which is then truncated to 498, but we want i * dx for element i+1, so 499 * dx = end for the 500th element.

The example code above passes by replacing floor with round. But some more tests would be good.

duncanmmacleod:
@dethodav, the example you post is, in my opinion, not an ideal example of the more general problem, since the extreme precision of the random float is critical to the operation. Are you able to post a more 'real-world' example where this issue arises?

dethodav:
@duncanmmacleod Here is a simpler example that shows the behavior: 

```
import numpy
from gwpy.types import Series

def check_val(x_max):
      x_series = numpy.linspace(0,x_max, 100)
      series = Series([0]*100, xindex=x_series)
      expected = series.xindex[-2]
      print(expected, series.crop(end=x_max).xindex[-1], expected == series.crop(end=x_max).xindex[-1])

check_val(10)
check_val(1)
```

This returns 
```
>>> check_val(10)
9.8989898989899 9.8989898989899 True
>>> check_val(1)
0.98989898989899 0.9797979797979799 False
```

Hence, we are still susceptible to floating point issues with "normal" numbers.

dethodav:
[...]

PR #1658, opened 2023-08-21
Change floor to round in crop() to fix a bug where an extra value was…

asouthgate:
… cut off. Add another test to go with it. See https://github.com/gwpy/gwpy/issues/1656


gwpy/types/series.py
```diff
@@ -1024,7 +1024,7 @@ def crop(self, start=None, end=None, copy=False):
                 side="left",
             )
         else:
-            idx1 = floor((xtype(end) - x0) / self.dx.value)
+            idx1 = round((xtype(end) - x0) / self.dx.value)
```
asouthgate, line 1027: See below, but slicing is done up to idx1, not inclusive.

So if the division results in 10.99999999, where it is really 11, it will be rounded down to 10. Since the grid is regular, there should be no points where we have e.g. 10.5. They should be floating representation of ints. See the unit test for an example.

duncanmmacleod:
@asouthgate, this PR doesn't seem to do what it is titled to do.

The change removes the use of `floor` and relies on the `searchsorted()` call, and doesn't introduce a new `round()` call.

I would like to keep, if at all possible, the `if irregular` logic so that for `regular` arrays the `xindex` array is not evaluated unnecessarily (which effectively doubles the memory size of the `Series` object).

asouthgate:
> @asouthgate, this PR doesn't seem to do what it is titled to do.
> 
> The change removes the use of `floor` and relies on the `searchsorted()` call, and doesn't introduce a new `round()` call.
> 
> I would like to keep, if at all possible, the `if irregular` logic so that for `regular` arrays the `xindex` array is not evaluated unnecessarily (which effectively doubles the memory size of the `Series` object).

Ah yes, the title is wrong. Apologies. Round doesnt work because it is wrong in another case (which i added a new test for). 

Okay, if end/dx creates something like 5.99999, it will always be rounded down unless we can fix the representation which I am not sure is reliable. We don't want to use round because then for array [1,2,3] specifying end=2.5 will give [1,2,3]. 

Perhaps I can iterate in place over the x values without creating xindex?

duncanmmacleod:
> Perhaps I can iterate in place over the x values without creating xindex?

That might work and should be memory efficient (if a bit slow?); I suppose there's only one way to find out.

gwpy/types/tests/test_series.py
```diff
@@ -238,6 +238,41 @@ def test_crop_float_precision(self):
...
+        series = Series([0] * arrlen, xindex=x_series)
+        expected = series.xindex[-2]
+        mid = (0.6 * series.xindex[-2] + 0.4 * series.xindex[-1])
+        assert series.crop(end=mid).xindex[-1] == expected
+
+        xmin = 0.508463154883984
+        x_series = numpy.linspace(xmin, 1.0, arrlen)
+        series = Series([0] * arrlen, xindex=x_series)
+        expected = series.xindex[0]
+        assert series.crop(start=xmin).xindex[0] == expected
```
duncanmmacleod, line 274: Can you please separate these new tests into at least one new test method, so that they can be evaluated separately to the existing `crop` test(s)?

gwpy/types/tests/test_series.py
```diff
@@ -238,6 +238,59 @@ def test_crop_float_precision(self):
...
+        assert series.crop(end=xmax).xindex[-1] == expected
+
+    def test_crop_between_grid_points_is_floored(self):
+        """Test that when we crop between xindex values, the result is floored.
+
+        This tests regression against https://github.com/gwpy/gwpy/issues/1656.
+        """
+        # e.g. x = [1, 2, 3], end = 2.5, result = [1, 2]
+        series = Series([0] * 3, xindex=[1, 2, 3])
+        print(series.crop(end=2.5))
```
duncanmmacleod, line 260: Please remove this line.

```suggestion
```

Issue #1659, opened 2023-08-21
Provide warning or throw if constructor Series.xindex is not same length as Series.data


asouthgate:
Currently this is allowed and provides no warning. Is this intended behaviour? E.g.

```
>>> series.Series([1,2,3], xindex=[0])

<Series([1., 2., 3.]
        unit=Unit(dimensionless),
        name=None,
        epoch=None,
        channel=None,
        x0=<Quantity 0.>,
        dx=None,
        xindex=<Index [0.]>)>

```

If not, should we add a warning or exception?

duncanmmacleod:
This should definitely be an error.

PR #1660, opened 2023-08-21
raise an exception if xindex has different length to data

asouthgate:
see https://github.com/gwpy/gwpy/issues/1659


duncanmmacleod (approved):
LGTM

PR #1661, opened 2023-08-22
Fix compatibility with matplotlib 3.8.0

duncanmmacleod:
This PR updates the handling of the `PlotArgsProcessor` class that overrides the builtin `_process_plot_var_args` private object that does all of the low-level argument handling and line generation for `Axes.plot()`.

In matplotlib <3.8.0 the parent `Axes` object is passed to the constructor, while in >=3.8.0 that is passed to the caller.

### DIFF
--- a/gwpy/plot/plot.py
+++ b/gwpy/plot/plot.py
@@ -521,9 +521,11 @@ class Plot(figure.Figure):
             pass
 
         # add new axes
-        if ax.get_axes_locator():
+        try:
             divider = ax.get_axes_locator()._axes_divider
-        else:
+        except AttributeError:
+            # get_axes_locator() is None _or_ the _axes_divider property
+            # has been removed
             from mpl_toolkits.axes_grid1 import make_axes_locatable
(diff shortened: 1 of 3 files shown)


duncanmmacleod:
See matplotlib/matplotlib@e418f5ce8cd509585911aa9e45e0c54bb15f2a02 for the impactful upstream change.

gwpy/plot/plot.py
```diff
@@ -521,9 +521,9 @@ def add_segments_bar(self, segments, ax=None, height=0.14, pad=0.1,
             pass
 
         # add new axes
-        if ax.get_axes_locator():
+        try:
             divider = ax.get_axes_locator()._axes_divider
-        else:
+        except (AttributeError, TypeError):
```
asouthgate, line 526: what's the reason for handling both of these errors? diff versions of matplotlib use different errors?

duncanmmacleod: This has been simplified.

Issue #1612, opened 2023-04-26
TimeSeries.fetch_open_data() fails when end time is at or just before a multiple of 4096


GNiklasch:
`TimeSeries.fetch_open_data()` raises an unexpected `ValueError` when the requested end time is an exact multiple of 4096 seconds, thus at a GWOSC file boundary, or a fraction of a second before such a boundary.

It works as expected when the end time is a full second before a multiple of 4096, or when it is just past a multiple of 4096.
(For my present purposes, both alternatives are viable workarounds - just wanted to report it for the record.)

Self-contained reproducible test case:
```python
from gwpy.timeseries import TimeSeries

def load_strain(ifo, start, end) -> None:
    try:
        strain = TimeSeries.fetch_open_data(ifo, start, end)
    except Exception as ex:
        print(type(ex))
        print(ex.args)
        #raise # uncomment to get a backtrace on the first failure
        return
    outcome = """Loaded strain, {0} - {1}, got {2}.
    """.format(start, end, strain.duration)
    print(outcome)
    return

# test case:
ifo, t_start, t_end = 'H1', 1268121592.0, 1268121600.0

# alternate test case:
# ifo, t_start, t_end = 'V1', 1257189368.0, 1257189376.0

load_strain(ifo, t_start, t_end) # fails
load_strain(ifo, t_start, t_end - 1/64.) # fails
load_strain(ifo, t_start, t_end - 1.) # works
load_strain(ifo, t_start, t_end + 1/64.) # works
```

Output:
```
<class 'ValueError'>
[1268121592 ... 1268121600)
<class 'ValueError'>
[1268121592 ... 1268121599.984375)
Loaded strain, 1268121592.0 - 1268121599.0, got 7.0 s.
    
Loaded strain, 1268121592.0 - 1268121600.015625, got 8.015625 s.
    
```

This is reproducible with Python 3.9.6 + GWpy 3.0.4 (latest) + gwosc 0.7.1 (latest)
(and `numpy` 1.24.3, `astropy` 5.2.2).

Uncommenting the re-`raise` in line 9 reveals that the `ValueError` comes from line 179 in `gwpy/timeseries/io/losc.py`:
```python
    keep = file_segment(url) & span
```

PR #1664, opened 2023-09-01
Catch ValueError in `fetch_gwosc_data`

EthanMarx:
Addresses #1612. 

This edge case occurs when the last `url` starts at the same timestamp that the `span` ends. 
This leads to [this](https://git.ligo.org/lscsoft/ligo-segments/-/blob/master/ligo/segments/__init__.py#L403) `ligo.segments`
`ValueError` getting raised for calling `__and__` with non overlapping Segments (My expectation would be to return an empty Segment, but I imagine theres good reason for choosing this behavior). 

This addresses the problem by catching the `ValueError` and continuing the iteration past this `url`

### DIFF
--- a/gwpy/timeseries/io/losc.py
+++ b/gwpy/timeseries/io/losc.py
@@ -176,7 +176,12 @@ def fetch_gwosc_data(detector, start, end, cls=TimeSeries, **kwargs):
     out = None
     kwargs['cls'] = cls
     for url in cache:
-        keep = file_segment(url) & span
+        try:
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@EthanMarx, thanks for putting this together, however this prompted me to actually look at the issue and I've proposed an alternative solution in #1665.

EthanMarx:
Awesome thanks Duncan. Your solution is certainly more robust. I'll close this.

PR #1665, opened 2023-09-01
Sieve URLs from gwosc client before fetch

### LINKED ISSUE #1612: TimeSeries.fetch_open_data() fails when end time is at or just before a multiple of 4096

duncanmmacleod:
This PR fixes #1612 (closes #1664) by `sieve()`ing the list of URLs received from the GWOSC client in case they overrun the actual requested interval.

### DIFF
--- a/gwpy/timeseries/io/losc.py
+++ b/gwpy/timeseries/io/losc.py
@@ -35,7 +35,10 @@ from gwosc.locate import get_urls
 
 from .. import (StateVector, TimeSeries)
 from ...io import (gwf as io_gwf, hdf5 as io_hdf5)
-from ...io.cache import file_segment
+from ...io.cache import (
+    file_segment,
+    sieve as sieve_cache,
+)
 from ...io.utils import file_path
 from ...detector.units import parse_unit
 from ...segments import Segment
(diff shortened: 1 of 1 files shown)


asouthgate:
@duncanmmacleod 

Looks good to me.

I also would have expected the operation in question to return an empty intersection instead of throwing a ValueError.

If you have time, an additional test would be good to protect against this bug.

asouthgate (approved):

Issue #1680, opened 2023-09-24
indexing a `FrequencySeries` returns a new `FrequencySeries` without indexed frequencies


jrollins:
```python
In [3]: from gwpy.timeseries import TimeSeries

In [4]: ts = TimeSeries.fetch('H1:CAL-DARM_ERR_DBL_DQ', 1378540810, 1378540818)

In [6]: psd = ts.psd()

In [7]: psd.size
Out[7]: 16385

In [19]: inds = np.where(psd.frequencies.value > 8000)[0]

In [20]: a = psd[inds]

In [21]: a.size
Out[21]: 384

In [22]: len(a.value)
Out[22]: 384

In [23]: len(a.frequencies)
Out[23]: 16385
```

Note the length of the `frequencies` attribute is the size of the original series, not the newly indexed series.

jrollins:
This problem applies to the `take()` method as well, which should presumably be doing the same thing.

jrollins:
FYI This issue appears to be fixed in 3.0.8.  Thank you all for addressing it.

PR #1685, opened 2023-10-02
Fix array indexing of Series.xindex

asouthgate:
See https://github.com/gwpy/gwpy/issues/1680.

When `__getitem__` is called with an int array, the xindex is not subsetted.

### DIFF
--- a/gwpy/types/sliceutils.py
+++ b/gwpy/types/sliceutils.py
@@ -106,6 +106,7 @@ def null_slice(slice_):
         slice_ = as_slice(slice_)
     except TypeError:
         return False
+    
     if isinstance(slice_, numpy.ndarray) and numpy.all(slice_ == 1):
         return True
     if isinstance(slice_, slice) and slice_ in (
--- a/gwpy/types/sliceutils.py
+++ b/gwpy/types/sliceutils.py
@@ -106,8 +106,7 @@ def null_slice(slice_):
         slice_ = as_slice(slice_)
     except TypeError:
         return False
-
-    if isinstance(slice_, numpy.ndarray) and numpy.all(slice_):
+    if isinstance(slice_, numpy.ndarray) and numpy.all(slice_ == 1):
         return True
     if isinstance(slice_, slice) and slice_ in (
             slice(None, None, None), slice(0, None, 1)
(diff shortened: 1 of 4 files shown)


gwpy/types/tests/test_series.py
```diff
@@ -167,6 +167,32 @@ def test_getitem(self, array):
...
             )
 
+    def test_getitem_index(self, array):
+        """Test that __getitem__ also applies to an xindex.
+
+        When subsetting a Series with an iterable of integer indices,
+        make sure that the xindex, if it exists, is also subsetted. Tests
+        regression against https://github.com/gwpy/gwpy/issues/1680.
+        """
+        xindex = array.xindex  # create xindex
```
duncanmmacleod, line 177: This value doesn't need to be assigned:

```suggestion
        array.xindex  # create xindex
```

gwpy/types/tests/test_series.py
```diff
@@ -167,6 +167,32 @@ def test_getitem(self, array):
...
+    def test_single_getitem_not_created(self, array):
+        """Test that array[i] does not return an object with a new _xindex."""
+
+        # check that there is no xindex when a single value is accessed
+        with pytest.raises(AttributeError):
+            xindex = array[0].xindex
+
+        # we don't need this, we don't want it accidentally injected
+        with pytest.raises(AttributeError):
+            xindex = array[0]._xindex
```
duncanmmacleod, line 194: These values don't need to be assigned either:

```suggestion
        # check that there is no xindex when a single value is accessed
        with pytest.raises(AttributeError):
            array[0].xindex

        # we don't need this, we don't want it accidentally injected
        with pytest.raises(AttributeError):
            array[0]._xindex
```

gwpy/types/sliceutils.py
```diff
@@ -107,7 +107,8 @@ def null_slice(slice_):
     except TypeError:
         return False
 
-    if isinstance(slice_, numpy.ndarray) and numpy.all(slice_ == 1):
+    if isinstance(slice_, numpy.ndarray) and \
+            slice_.dtype == bool and slice_.all():
```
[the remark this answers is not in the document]
asouthgate: Sure, no problem.

duncanmmacleod (approved):
LGTM, thanks @asouthgate.

PR #1695, opened 2023-10-16
Replace maya with dateparser in gwpy.time

duncanmmacleod:
This  PR replaces all use of the `maya` module with `dateparser`.

`maya` is not actively maintained any more, while `dateparser` is, and is what `maya` calls out to anyway.


asouthgate (approved):
LGTM

Issue #1673, opened 2023-09-15
Fix inconsistency with TimeSeries.resample


vbaws:
gwpy.timeseries.TimeSeries.resample should at least replicate LIGO's up-sampling functionality of copying samples when resampling to a higher sample rate.

In my testing asking for a higher sample rate does nothing presently.

asouthgate:
Hi @vbaws, sounds like a bug. It shouldn't do nothing. Could you provide a snippet for a circumstance that fails?

`
from gwpy.timeseries import TimeSeries
import matplotlib.pyplot as plt
from gwpy.plot import Plot
from numpy import random

series = TimeSeries(random.random(1000), sample_rate=100, unit='m')
resamp = series.resample(series.sample_rate * 1.1)

assert series.sample_rate < resamp.sample_rate

Plot(series, resamp)

plt.show()
`

Seems to upsample okay.

asouthgate:
Sure, no problem. I am on leave next week but will take a look afterward.

vbaws:
Regarding "Can I ask more specifically on the reason this particular upsampling method is needed or a reference for my information? We could add this as a feature request." : 

This is how upsampling operates in the LIGO front ends. I needed to utilise it in an attempt to debug the signal flow from some my channels after they are upsampled. Creating imaging of the signal of through these means is genuinely something we make use of when trying to actuate on something at a frequency far above our sampling rate. Filtering above f_s/2 isn't "real" in the digital system and is done through anti-imaging analog chassis. These have their own filter responses that can be modeled independently, or in my case, there might not be any anti-imaging deliberately.

asouthgate:
Okay sure, I am happy to add this as a feature early next week.

duncanmmacleod:
This use case seems extremely specific to the way the LIGO analog<->digital controls system is configured.

GWpy basically relies on Scipy for all of its actual signal processing routines (that's not _entirely_ true, but I would like it to be), so am hesitant to implement something that may require significant testing/maintenance.

However, if the code is simple enough to understand, to test, and to maintain, I may be convinced.

duncanmmacleod:
Migrated to https://gitlab.com/gwpy/gwpy/-/issues/1673, please join us on gitlab to continue your contribution!

PR #1697, opened 2023-10-17
Add an upsample_repeat function that uses numpy.repeat to upsample

asouthgate:
This should address https://github.com/gwpy/gwpy/issues/1673

In the absence of a reference for the correct name for this method of upsampling, I have named it after the repeat function in numpy.

I also have given it a separate interface instead of modifying TimeSeries.resample.

### DIFF
--- a/gwpy/timeseries/tests/test_timeseries.py
+++ b/gwpy/timeseries/tests/test_timeseries.py
@@ -1172,19 +1172,20 @@ class TestTimeSeries(_TestTimeSeriesBase):
         assert factor * upsamped.dx == gw150914.dx
         assert upsamped.x0 == upsamped.xindex[0]
 
-        # since numpy.arange, which is used when generating an index anew,
-        # lacks precision, we must use approximate assertion
+        # Unlike the values, the xindex is not repeat expanded.
+        # Since numpy.arange, which is used when generating an index anew,
+        # lacks precision, we must use approximate assertion.
         numpy.testing.assert_allclose(
             gw150914.xindex.value,
             upsamped.xindex[::factor].value
         )
 
         for j in range(len(gw150914)):
-            val = gw150914[j].value
+            val = gw150914.value[j]
             for fj in range(factor):
                 # Each index of input valj maps to an index factor * valj
                 upj = factor * j + fj  # index of each repeated sample
-                numpy.testing.assert_equal(val, upsamped[upj])
+                numpy.testing.assert_equal(val, upsamped.value[upj])
 
     def test_upsample_repeat_float_exception(self, gw150914):
         """Assert that upsampling by a non-integer factor raises a ValueError.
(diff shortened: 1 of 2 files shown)


gwpy/timeseries/timeseries.py
```diff
@@ -951,6 +952,32 @@ def resample(self, rate, window='hamming', ftype='fir', n=None):
...
+    def upsample_repeat(self, factor):
+        """Upsample this TimeSeries by repeating elements by an integer factor.
+
+        This will return a new TimeSeries with a new sample_rate equal to the
+        previous one multiplied by factor.
+
+        Parameters
+        ----------
+        factor: `int`
+                The number of times to repeat each element
```
duncanmmacleod, line 964:
```suggestion
            The number of times to repeat each element
```

gwpy/timeseries/timeseries.py
```diff
@@ -951,6 +952,32 @@ def resample(self, rate, window='hamming', ftype='fir', n=None):
...
+        TimeSeries
+            a new Series with the resampling applied, and the same
+            metadata
+        """
+        if not isinstance(factor, numbers.Integral):
+            raise ValueError("Cannot upsample_repeat with non-integer factor.")
+        repeat_inds = numpy.repeat(numpy.arange(len(self)), factor)
+        new = self[repeat_inds]  # xindex should also be handled here
+        new.__metadata_finalize__(self)
+        new._unit = self.unit
```
duncanmmacleod, line 977: I hope this shouldn't be required, can you test without it. If it _is_ required, maybe that indicates a hole in the `__getitem__` overrides for `Series`.

duncanmmacleod (changes requested):
Minor pedantry, and a query, but this looks great.

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -1160,6 +1160,34 @@ def test_resample_noop(self):
             new = data.resample(data.sample_rate)
             assert data is new
 
+    def test_upsample_repeat_factor_3(self, gw150914):
+        """Test that repeat upsampling by a factor of 3 works as intended.
+        """
+        factor = 3
+        upsamped = gw150914.upsample_repeat(factor)
+        upsamped.xindex  # generate the x index
```
duncanmmacleod, line 1168: Is the `xindex`  tested here? If not, why generate it?

gwpy/timeseries/tests/test_timeseries.py
```diff
@@ -1160,6 +1160,34 @@ def test_resample_noop(self):
...
+        assert len(upsamped) == len(upsamped.xindex)
+
+    def test_upsample_repeat_float_exception(self, gw150914):
+        """Assert that upsampling by a non-integer factor raises a ValueError.
+        """
+        factor = 3.3
+        with pytest.raises(ValueError):
+            gw150914.upsample_repeat(factor)
```
duncanmmacleod, line 1189: Can this be simplified to not have a variable assignment?

```suggestion
        with pytest.raises(ValueError):
            gw150914.upsample_repeat(3.3)
```

PR #1699, opened 2023-10-18
Fix the Zenodo RST utility module

duncanmmacleod:
This PR fixes the `gwpy.utils.sphinx.zenodo` API interface module uses to populate the `citing` page on the online docs (e.g. [here](https://gwpy.github.io/docs/latest/citing/)), which was broken by a recent (apparently undocumented) Zenodo API change (see https://github.com/zenodo/zenodo/issues/2483).

I also took the opportunity to refactor the module to improve the layout, and added a test module.

### DIFF
--- a/gwpy/utils/tests/test_sphinx_zenodo.py
+++ b/gwpy/utils/tests/test_sphinx_zenodo.py
@@ -1,0 +1,157 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) Cardiff University (2023)
+#
+# This file is part of GWpy.
+#
+# GWpy is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# GWpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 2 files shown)


gwpy/utils/tests/test_sphinx_zenodo.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    rst = gwpy_zenodo.format_citations(
+        597016,
+    )
+    latestversion = latest["tag_name"].lstrip("v")
+    latesthead = "-" * len(latestversion)
+    assert f"""
+{latesthead}
+{latestversion}
+{latesthead}""".strip() in rst
```
asouthgate, line 112: Is this formatting intentional?

duncanmmacleod: I'm not sure what you are asking. This `assert` checks that an RST-format header for the latest-discovered version is present in the rendered text, so 'yes' this is intentional.

duncanmmacleod: I support this could have been achieved via `'\n'.join` or something a little cleaner than a block quote, but heyho.

asouthgate (approved):
Approved, optionally change formatting referred to in comment.

gwpy/utils/tests/test_sphinx_zenodo.py
asouthgate: Sorry, I meant the indentation.

asouthgate: (but it's not a problem)

PR #1701, opened 2023-10-23
Implement concurrent downloads

asouthgate:
Needs benchmarking. Perhaps no benefit at this time (but no downside to have the functionality).

### DIFF
--- a/gwpy/timeseries/io/losc.py
+++ b/gwpy/timeseries/io/losc.py
@@ -24,6 +24,7 @@ For more details, see :ref:`gwpy-table-io`.
 
 import os.path
 import re
+from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor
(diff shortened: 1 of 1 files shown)


gwpy/timeseries/io/losc.py
```diff
@@ -131,7 +132,14 @@ def _overlapping(files):
 
 # -- remote data access (the main event) --------------------------------------
 
-def fetch_gwosc_data(detector, start, end, cls=TimeSeries, **kwargs):
+def fetch_gwosc_data(
+        detector,
+        start,
+        end,
+        nthreads=10,
```
duncanmmacleod, line 139: Can you please add this option to the 'upstream' `TimeSeriesBase.fetch_open_data` method, or at least document the keyword argument there.

And also, probably somewhere in the docs.

PR #1710, opened 2023-11-13
Run apt-get update before apt-get install in CI jobs

duncanmmacleod:
This PR attempts to fix the current failures in the 'experimental' dependencies CI job that is failing to find debian packages to install.


asouthgate (approved):
Looks reasonable to me

PR #1712, opened 2023-11-19
Drop `HAS_TEX` for faster startup time

francois-rozet:
Hello :wave: I have noticed that a significant part of the startup time of `gwpy` comes from the execution of `has_tex` to store the result in `HAS_TEX`, which is never used later on.

In this PR, I deleted `HAX_TEX` and cached the result of `has_tex` in case it would be necessary to call it several times. It reduces the startup time by about 2 seconds which could be huge for some applications.

#### Before

```console
user@host:~ $ time python -c 'from gwpy.timeseries import TimeSeries'
real	0m3,381s
user	0m2,600s
sys	0m2,091s
```

#### After

```console
user@host:~ $ time python -c 'from gwpy.timeseries import TimeSeries'
real	0m1,360s
user	0m1,430s
sys	0m1,419s
```

### DIFF
--- a/gwpy/plot/tests/test_tex.py
+++ b/gwpy/plot/tests/test_tex.py
@@ -39,6 +39,7 @@ def test_has_tex_missing_exe():
     """Test that `gwpy.plot.tex.has_tex` returns `False` when
     any one of the necessary executables is missing.
     """
+    plot_tex.has_tex.cache_clear()
     assert not plot_tex.has_tex()
 
 
@@ -47,6 +48,7 @@ def test_has_tex_bad_latex(_):
     """Test that `gwpy.plot.tex.has_tex` returns `False` when
     the LaTeX figure fails to render.
     """
+    plot_tex.has_tex.cache_clear()
     assert not plot_tex.has_tex()
 
 
@@ -57,6 +59,7 @@ def test_has_tex_true(_which_, _test_usetex):
     all of the necessary executables are found, and the LaTeX figure
     doesn't raise an exception.
     """
+    plot_tex.has_tex.cache_clear()
     assert plot_tex.has_tex()
 
 
--- a/gwpy/plot/tex.py
+++ b/gwpy/plot/tex.py
@@ -21,6 +21,7 @@
 """
 
 import re
+from functools import lru_cache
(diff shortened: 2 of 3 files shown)


duncanmmacleod:
@francois-rozet, thanks for this, this is a great idea for a simple performance bump.

The CI jobs are failing because the test suite is now attempting to test a cached function. The following patch should resolve this, if you are able to apply it with `git am`:

```patch
From 8a447c8f1c7d2692050eb3b1d829a7cab60f4dc6 Mon Sep 17 00:00:00 2001
From: "duncan.macleod" <duncan.macleod@ligo.org>
Date: Thu, 23 Nov 2023 10:04:21 +0000
Subject: [PATCH] gwpy.plot: clear has_tex cache when testing

---
 gwpy/plot/tests/test_tex.py | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/gwpy/plot/tests/test_tex.py b/gwpy/plot/tests/test_tex.py
index c9603990..dd4d23d3 100644
--- a/gwpy/plot/tests/test_tex.py
+++ b/gwpy/plot/tests/test_tex.py
@@ -39,6 +39,7 @@ def test_has_tex_missing_exe():
     """Test that `gwpy.plot.tex.has_tex` returns `False` when
     any one of the necessary executables is missing.
     """
+    plot_tex.has_tex.cache_clear()
     assert not plot_tex.has_tex()


@@ -47,6 +48,7 @@ def test_has_tex_bad_latex(_):
     """Test that `gwpy.plot.tex.has_tex` returns `False` when
     the LaTeX figure fails to render.
     """
+    plot_tex.has_tex.cache_clear()
     assert not plot_tex.has_tex()


@@ -57,6 +59,7 @@ def test_has_tex_true(_which_, _test_usetex):
     all of the necessary executables are found, and the LaTeX figure
     doesn't raise an exception.
     """
+    plot_tex.has_tex.cache_clear()
     assert plot_tex.has_tex()


--
2.42.0

```

francois-rozet:
@duncanmmacleod I applied your patch!

duncanmmacleod (approved):
LGTM, thanks @francois-rozet!

PR #1722, opened 2023-11-28
Read the unitX from a GWF with frameCPP

duncanmmacleod:
This PR patches the frameCPP GWF timeseries reader to actually use the `unitX` attribute stored in the FrVect structure for a given channel in a GWF.

Depends on #1721.

### DIFF
--- a/gwpy/timeseries/io/gwf/framecpp.py
+++ b/gwpy/timeseries/io/gwf/framecpp.py
@@ -401,6 +401,7 @@ def read_frvect(vect, epoch, start, end, name=None, series_class=TimeSeries):
     dim = vect.GetDim(0)
     dx = dim.dx
     x0 = dim.startX
+    xunit = dim.GetUnitX() or None
 
     # start and end GPS times of this FrVect
     dimstart = epoch + x0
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@eagoetz, can you please take a look at this one? In practice this should have no impact, but in principle means that if someone creates a TimeSeries in GWF with time-units of 'hour', that will be respected by GWpy.

eagoetz:
> @eagoetz, can you please take a look at this one? In practice this should have no impact, but in principle means that if someone creates a TimeSeries in GWF with time-units of 'hour', that will be respected by GWpy.

Yes, I'll take a look now

gwpy/timeseries/io/gwf/framecpp.py
```diff
@@ -401,6 +401,7 @@ def read_frvect(vect, epoch, start, end, name=None, series_class=TimeSeries):
     dim = vect.GetDim(0)
     dx = dim.dx
     x0 = dim.startX
+    xunit = dim.GetUnitX()
```
eagoetz, line 404: Is there any risk that the units for "x" do not exist? I ask because of the line `unit = vect.GetUnitY() or None`

duncanmmacleod: According to the GWF spec there is no risk that the unit is not specified at all (the variable doesn't exist), but there is a possibility that it is set to the empty string, which is probably accepted by the constructor, but is actually an invalid value. I have updated this patch to include the `or None` catch just in case. Thanks.

eagoetz (approved):
Thanks for the update! LGTM

PR #1723, opened 2023-12-05
Allow localhost sockets during dist testing

duncanmmacleod:
~This PR adds the `--allow-unix-socket` to the `dist` CI jobs so that `asyncio` works with 'self-sockets' (required by the latest release of `freezegun`).~

EDIT: for obvious reasons this solution doesn't work on Windows, so now...

This PR adds `--allow-hosts=127.0.0.1` to the `dist` CI jobs so that `asyncio` works with 'self-sockets' (required by the latest release of `freezegun`).

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -118,6 +118,7 @@ jobs:
             --cov gwpy \
             --cov-report=xml \
             --disable-socket \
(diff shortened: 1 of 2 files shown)


eagoetz (approved):
LGTM

PR #1730, opened 2023-12-08
Propagate source keyword in DataQualityDict.populate

duncanmmacleod:
This PR fixes #1729 by correctly propagating the `source` keyword as the URL for DQSegDB queries.

PR #1731, opened 2023-12-08
Rename DataQualityDict.query_dqsegdb 'url' keyword to 'host'

duncanmmacleod:
This PR renames the `url` keyword for `DataQuality{Flag,Dict}.query_dqsegdb` from `url` to `host`, to better match the upstream `dqsegdb2` parameters.

This also updates the `DEFAULT_SEGMENT_SERVER` to be imported from `dqsegdb2` rather than independently set, and introduces some reformatting for readability.

Depends on #1730.

### DIFF
--- a/gwpy/segments/flag.py
+++ b/gwpy/segments/flag.py
@@ -30,7 +30,6 @@ for handling multiple flags over the same global time interval.
 import datetime
 import json
 import operator
-import os
 import re
 import warnings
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
Migrated to https://gitlab.com/gwpy/gwpy/-/merge_requests/1731.

PR #1737, opened 2023-12-15
Refactor colorbar() function outside of Plot or Axes

### LINKED ISSUE #1736: Cannot generate colour bar for Axes attached to mpl.figure.Figure
Attempting to create a GWpy colourbar for an `Axes` attached directly to a `matplotlib.figure.Figure` (rather than a `gwpy.plot.Plot`) fails in a fairly incomprehensible way:

```python
>>> import gwpy
>>> import numpy
>>> from matplotlib import pyplot
>>> data = numpy.random.rand(100, 100)
>>> fig = pyplot.figure()
>>> ax = fig.gca()
>>> ax.imshow(data)
>>> ax.colorbar()
  File "/home/duncan/opt/mambaforge/envs/py311/lib/python3.11/site-packages/matplotlib/figure.py", line 163, in update
    raise ValueError('left cannot be >= right')
ValueError: left cannot be >= right
```

This was introduced by 73b5ef065c8fafdd625601671e5057750a65bbee, which hasn't been released yet.

duncanmmacleod:
This PR refactors the `colorbar()` creation function to be outside of both the `Plot` and `Axes` classes, but called from the `.colorbar()` methods of both.

This is mainly to support calling `ax.colorbar()` from `Axes` that are attached to a `Figure` (the superclass of `Plot`), rather than a `Plot`. Fixes #1736.


duncanmmacleod:
Migrated to https://gitlab.com/gwpy/gwpy/-/merge_requests/1737.

PR #1746, opened 2024-04-08
Change coherence unit to dimensionless

artem-basalaev:
Coherence between time series should be a dimensionless number from 0 to 1 (in my understanding), however in GWpy a fictional unit of "coherence" is assigned. This results in `UnitConversionError` in any arithmetic involving coherence and unitless quantities. Such as this example code:

```python
from numpy import random
ts1 = TimeSeries(random.random(1000), sample_rate=100, unit='m')
ts2 = TimeSeries(random.random(1000), sample_rate=100, unit='m')
coh = ts1.coherence(ts2)
# line below will cause UnitConversionError
print(1.0-coh)
```

This pull request instead forces the unit to be `dimensionless_unscaled`, which does not cause this error.

### DIFF
--- a/gwpy/signal/spectral/_scipy.py
+++ b/gwpy/signal/spectral/_scipy.py
@@ -23,6 +23,7 @@ import numpy
 import warnings
 
 import scipy.signal
+from astropy import units
 
 from ...frequencyseries import FrequencySeries
 from ._utils import scale_timeseries_unit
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
Migrated to https://gitlab.com/gwpy/gwpy/-/merge_requests/1746.

PR #1747, opened 2024-04-14
Quick fix reuse of whiten argument issue #1652

areeda:
Another try to deal with this change necessary because tests for thuthiness are more strict.

### DIFF
--- a/gwpy/cli/qtransform.py
+++ b/gwpy/cli/qtransform.py
@@ -215,7 +215,8 @@ class Qtransform(Spectrogram):
             outseg = Segment(gps, gps).protract(args.plot[self.plot_num])
 
             # use the precomputed ASD as the whitener if needed
-            if self.qxfrm_args.get("whiten"):
+            if type(self.qxfrm_args.get("whiten")) == bool and \
+                    self.qxfrm_args.get("whiten"):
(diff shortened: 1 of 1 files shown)


gwpy/cli/qtransform.py
```diff
@@ -215,7 +215,8 @@ def get_spectrogram(self):
             outseg = Segment(gps, gps).protract(args.plot[self.plot_num])
 
             # use the precomputed ASD as the whitener if needed
-            if self.qxfrm_args.get("whiten"):
+            if type(self.qxfrm_args.get("whiten")) == bool and \
+                    self.qxfrm_args.get("whiten"):
```
duncanmmacleod, line 219: I'm not a fan of direct type comparisons (neither is PEP-0008), so this might be better constructed as:

```suggestion
            if self.qxfrm_args.get("whiten") is True:
```

duncanmmacleod (changes requested):
Thanks for catching this @areeda, I understand the need for this change but have suggested a different solution.

areeda:
I'm not sure why but having accepted and committed your suggestion the ststus remains "changes requeste.
@duncanmmacleod anything else I need to do?

duncanmmacleod (approved):
LGTM

PR #1753, opened 2024-05-13
Fix usage of hann window for scipy 1.13.0

duncanmmacleod:
This PR fixes compatibility with scipy 1.13.0 by using `scipy.signal.get_window` to create a `hann` window in the tests (accessing window functions as attributes of the `scipy.signal` module is deprecated).


duncanmmacleod:

PR #1755, opened 2024-05-13
Skip tests that fail on Zenodo rate limit

duncanmmacleod:
This PR adds a handler for tests of `gwpy.utils` that fail because they hit a Zenodo rate limit, since those are not indicative of an actionable failure in the code itself.


duncanmmacleod:

areeda (approved):

PR #1764, opened 2024-06-20
Improve handling of invalid inputs when reading data from GWF with LALFrame

duncanmmacleod:
This PR improves the handling of invalid inputs (start and end GPS times) when attempting to read data from GWF with LALFrame to provide better error messages and avoid [inconsistent behaviour of LALFrame on non-x86_64 systems](https://git.ligo.org/lscsoft/lalsuite/-/issues/745).

The result should be no change in the GWpy UX, other than perhaps a different error class when passing invalid inputs (`ValueError` instead of `RuntimeError`).

### DIFF
--- a/gwpy/testing/utils.py
+++ b/gwpy/testing/utils.py
@@ -34,12 +34,14 @@ from numpy.testing import (assert_array_equal, assert_allclose)
 
 from astropy.time import Time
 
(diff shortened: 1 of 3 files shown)


duncanmmacleod:

areeda (approved):
LGTM

PR #1773, opened 2024-07-16
Update all references to nds.gwosc.org

duncanmmacleod:
This PR updates any references to `losc-nds.ligo.org` to `nds.gwosc.org`.


duncanmmacleod:
This needs to be migrated to gitlab.com.

PR #1776, opened 2026-09-15
Updates default segment server to igwn

robertbruntz:
Updates default segment server from segments.ligo.org to segments.igwn.org in `gwpy/segments/flag.py`.


robertbruntz:
This PR was based on an old branch of GWpy (master, vs. main).

